PR #1, opened 2022-08-01
setup initial documentation folders

loganmeetsworld:
# Initial documentation setup #

## 🗣 Description ##

This sets up several folders: 
- `.github` with workflows and issue templates we can add to
- `docs` with ADRs, diagrams, and research folders. I am particularly interested in adding secondary research here.
- A basic README.md
- A CODEOWNERS file with just one assignment for now

### DIFF
--- a/docs/research/README.md
+++ b/docs/research/README.md
@@ -1,0 +1,3 @@
+# Dotgov Research
+
+This contains scripts for gathering data about dotgov usage and answering various secondary research questions.
--- a/docs/research/README.md
+++ b/docs/research/README.md
@@ -1,3 +1,3 @@
-# Dotgov Research
+# .gov Research
 
-This contains scripts for gathering data about dotgov usage and answering various secondary research questions.
+This contains scripts for gathering data about .gov usage and answering various secondary research questions.
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,3 @@
+# For more information on CODEOWNERS, see the documentation: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
+
+docs/* @dotgov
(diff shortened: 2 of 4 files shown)


docs/research/README.md
```diff
@@ -0,0 +1,3 @@
+# Dotgov Research
+
+This contains scripts for gathering data about dotgov usage and answering various secondary research questions.
```
h-m-f-t, line 3: I have found "dotgov" means many different (sometimes diverging) things to different parties. I generally use the following for different flavors of .gov:
* DOTGOV - an act of Congress
* .gov - a TLD on the internet

GSA used to use DotGov/dotgov as shorthand for the program that exists to manage the .gov registry. Outside of hostnames ([home.dotgov.gov)](https://home.dotgov.gov), I intentionally don't use DotGov anymore because I think using it as a modifier of some _thing_ (".gov {program, registry, namespace}") is easier to intuit when reading and more technically accurate. 

Recognizing the repo is called "dotgov", when this sentence says "data about dotgov usage", is it talking about the new .gov registrar, the set (or some part) of domains and hosts that comprise the .gov namespace, or something else?

h-m-f-t (approved):
I left one non-critical comment, but no changes are needed. Thanks for preparing this!

colinmurphy01:
I've also used the Github wiki feature for documentation purposes. Is there a benefit to having the docs hosted as files or in the github wiki?

SSPJ:
> I've also used the Github wiki feature for documentation purposes. Is there a benefit to having the docs hosted as files or in the github wiki?

I believe there is a benefit to having the ADRs committed directly to the repository, at least.

Lost documentation is a major problem for code maintainability. GitHub wiki is a GH product, so it does not have the same expectation of longevity as `git` itself. Checking the ADRs into the repo is also our best shot at ensuring the documentation travels with the code, including onto the very servers running the code.

loganmeetsworld:
I agree with @SSPJ. There might be other uses for the wiki for other types of documentation in the longrun, like onboarding guides and runbooks, but for now .git is best for these types of documentation we have here.

PR #2, opened 2022-08-03
add scripts for generating icann lookup and response code csvs

loganmeetsworld:
# Add lookup/response scripts #

## 🗣 Description ##

I have generated some data for the team and these are scripts for performing ICANN lookups and retrieving response statuses for domains. These are the basic versions without any sort of asynchronous processing of the requests. I run these with tmux for a long period locally. 

We have certain research questions about the rate of responses for the current list of domains. We also have questions about current registrar usage. This helps us answer some of those questions with data. 

## Questions for reviewers ##

Should I add the current set of generated data to this repository? This would allow us to show notebooks using the data with research questions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+docs/research/data/**
--- a/docs/research/scripts/icann_lookup.py
+++ b/docs/research/scripts/icann_lookup.py
@@ -1,3 +1,10 @@
+"""
+This script takes each domain in a dataset of non-.gov government domains and looks for 
+which registrar they are currently registered with. 
+
+This script can be run locally to generate data and currently takes some time to run.
+"""
+
 import csv
 import requests
 import whois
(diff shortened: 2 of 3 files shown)


SSPJ (approved):
I have not tested these scripts, but I've read the code and can't see any issues with it.

👍

SSPJ:
> Should I add the current set of generated data to this repository? This would allow us to show notebooks using the data with research questions.

I have some concerns about committing data which will become outdated. Having the data in our drive might be a more accessible format for the UX folks as well.

h-m-f-t (approved):
* I was able to run response_codes.py successfully and the results are as expected.
* I kept hitting the exception in L24 in icann_lookup.py ("Something went wrong").

It'd be good to have another person do a successful run before merging, but that's not _really_ necessary for little scripts like this, intended for local use.

loganmeetsworld:
> I kept hitting the exception in L24 in icann_lookup.py ("Something went wrong").

@h-m-f-t that's fine as long as the script didn't exit. It should hit that and report something went wrong and keep running.

PR #35, opened 2022-08-10
Tiny bit of Docker

SSPJ:
## 🗣 Description ##

Adding docker and pipenv to the repository, plus a number of related ADRs.

Also a license file.

To keep PRs small and reviewable, this represents a small chunk of the work needed to finish out the proof of concept epic.

### DIFF
--- a/docs/architecture/decisions/0002-docker-for-development.md
+++ b/docs/architecture/decisions/0002-docker-for-development.md
@@ -1,0 +1,25 @@
+# 2. Docker for development
+
+Date: 2022-08-09
+
+## Status
+
+Accepted
+
+## Context
+
+The [Federal Audit Clearinghouse](https://github.com/GSA-TTS/FAC) project is a similar project to ours, from which we are taking many examples for reuse. FAC uses Docker. Docker is widely accepted at TTS currently and in the development community at large.
+
+Other options include using virtual machines or installing dependencies directly onto the host system. The first is not widely used at TTS; the second tends to cause problems with dependency management.
+
+## Decision
+
+To include a Dockerfile and docker-compose file.
+
+## Consequences
+
+We intend to deploy using containerized images; therefore, use of Docker on local will give the closest approximation to higher environments.
+
+The project may be perceived as less friendly to outsiders.
+
+A greater technical burden may be placed on new employees during onboarding.
--- a/docs/architecture/decisions/0002-docker-for-development.md
+++ b/docs/architecture/decisions/0002-docker-for-development.md
@@ -8,7 +8,9 @@ Accepted
 
 ## Context
 
-The [Federal Audit Clearinghouse](https://github.com/GSA-TTS/FAC) project is a similar project to ours, from which we are taking many examples for reuse. FAC uses Docker. Docker is widely accepted at TTS currently and in the development community at large.
+Docker is widely accepted at TTS currently and in the development community at large. It works well with cloud.gov and is very portable to other cloud solutions.
+
+The [Federal Audit Clearinghouse](https://github.com/GSA-TTS/FAC) project is a similar project to ours, from which we are taking many examples for reuse. FAC uses Docker.
 
 Other options include using virtual machines or installing dependencies directly onto the host system. The first is not widely used at TTS; the second tends to cause problems with dependency management.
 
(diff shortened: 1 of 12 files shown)


docs/architecture/decisions/0002-docker-for-development.md
```diff
@@ -0,0 +1,25 @@
...
+
+Date: 2022-08-09
+
+## Status
+
+Accepted
+
+## Context
+
+The [Federal Audit Clearinghouse](https://github.com/GSA-TTS/FAC) project is a similar project to ours, from which we are taking many examples for reuse. FAC uses Docker. Docker is widely accepted at TTS currently and in the development community at large.
```
loganmeetsworld, line 11: Might add that Docker works well with cloud.gov and is very portable to other cloud solutions.

docs/architecture/decisions/0002-docker-for-development.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Decision
+
+To include a Dockerfile and docker-compose file.
+
+## Consequences
+
+We intend to deploy using containerized images; therefore, use of Docker on local will give the closest approximation to higher environments.
+
+The project may be perceived as less friendly to outsiders.
```
loganmeetsworld, line 23: What does this mean? Why so?

docs/architecture/decisions/0002-docker-for-development.md
```diff
@@ -0,0 +1,25 @@
...
+
+A greater technical burden may be placed on new employees during onboarding.
```
loganmeetsworld, line 25: How so?

docs/architecture/decisions/0003-django.md
```diff
@@ -0,0 +1,35 @@
...
+
+Date: 2022-08-09
+
+## Status
+
+Accepted
+
+## Context
+
+The [Federal Audit Clearinghouse](https://github.com/GSA-TTS/FAC) project is a similar project to ours, from which we are taking many examples for reuse. FAC uses Django.
```
loganmeetsworld, line 11: I would maybe put this at the bottom. We were thinking of using Django independently but it is a happy coincidence that FAC uses it. Putting at the top might imply we are somehow tied to their technology choices. I think it is a plus but not the strongest reason.

docs/architecture/decisions/0003-django.md
```diff
@@ -0,0 +1,35 @@
...
+
+Django is well documented and approachable, even for people with limited programming experience.
+
+We will have a dependency on the health of the Django project and the Python ecosystem at large.
+
+Python is an interpreted language and will not have the same performance as compiled languages. Django is slower than its less featureful Python competitors (while being faster than Ruby on Rails, its nearest other-language competitor).
+
+The job market for Django developers is currently (8/2022) strong, with 7,000 open positions on a popular career site.
+
+Moving away from Django at a future point would require a complete rewrite of the codebase.
```
loganmeetsworld, line 35: Would it require a rewrite of the frontend?

loganmeetsworld (approved):
Looks great! I am so appreciative of the ADRs 🙏🏻 

I had a couple questions on ADRs but otherwise everything looks good. 

I successfully built the image and started it 🚀

docs/architecture/decisions/0002-docker-for-development.md
SSPJ: Added, ty.

docs/architecture/decisions/0003-django.md
SSPJ: Good point. I've moved it.

docs/architecture/decisions/0002-docker-for-development.md
SSPJ: After giving some thought, I've removed this line.

It has to do with history and culture in ways that are difficult to summarize in a brief paragraph. There's a quote from Brendan Eich in that [JavaScript article](https://devclass.com/2022/08/04/retire_javascript_says-json-creator-douglas-crockford/) that was recently shared in Slack which came close: "I went in to do … a programming language for HTML, for web designers and programmers to use, embedded directly in the web page … not like Java which was a professional’s language where you would run real code with type declarations, and you’d have to write in a way that compiled."

docs/architecture/decisions/0003-django.md
SSPJ: Hard to say, I don't know what our frontend is yet. 😆 

More broadly, if we use a separate server communicating over a REST API (the "backend" and "frontend" paradigm), then the frontend code would only need updated if REST API changed.

If we use server side rendering in an MVC pattern, then Django has the flexibility to use numerous templating engines. The default is Django's own engine, but Jinja2 is a popular replacement (both have very similar syntax).

Having to rewrite the templates would mainly be a matter of whether the new framework/paradigm could support whatever templating engine was picked or not.

PR #39, opened 2022-08-11
add ability to one-off domain lookups, add more error handling

loganmeetsworld:
# Update ICANN lookups #

## 🗣 Description ##

@h-m-f-t was having some issues with this script. I've updated it to allow for one off domains like 

```
python3 icann_lookup.py --domain <DOMAIN>
```

to do some testing. I could not replicate issues with the domain that @h-m-f-t was having issues with, and my script completed running without issue. I think that it may be something in a network that prevents hitting certain domains when run locally? Not sure. But I've also added some more info about which domain had an issue before exiting to help debug.

### DIFF
--- a/docs/research/scripts/icann_lookup.py
+++ b/docs/research/scripts/icann_lookup.py
@@ -42,6 +42,8 @@ def main(domain):
                 registrar = check_registration(domain_name)
                 full_data.append(domain + [registrar])
 
+    Path("../data").mkdir(exist_ok=True)
+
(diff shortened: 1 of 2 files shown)


SSPJ (approved):
Nice addition! 👍 

nonblocking: This script is getting long enough that I think some docstrings on the functions would be appropriate, time permitting.

nonblocking: Would suggest adding shebang line at top so that this script can be made executable and run directly.

```bash
#!/usr/bin/env python3
```

PR #42, opened 2022-08-12
Add Django

### LINKED ISSUE #12: Determine what language and framework to use for registrar backend
- [x] Draft ADR and present to CISA for sign-off
    - [x] document trade-offs
- [x] Create ~and deploy~ hello world

Deployment happens in #11 .

### LINKED ISSUE #13: Determine what language and framework to use for registrar frontend
- [x] Hold conversation between Design and Engineering
    The focus is to determine whether any components of the interface justify use of a JavaScript framework or a separate frontend service. The result of this conversation may be clear consensus or may it may lead to areas for research or items to test out in the POC.
- [x] Draft ADR to document choice and trade-offs
    - [x] Create a system diagram to illustrate components in chosen solution.

SSPJ:
## 🗣 Description ##

This PR adds a GitHub Actions pipeline to automate deploying the main branch to the sandbox.

It also adds Django with the minimal code required to answer a health check: https://getgov-dev.app.cloud.gov/health/

Example GitHub Action runs:
* ✅ [Build and deploy](https://github.com/cisagov/dotgov/actions/runs/2876764894)
* ✅ [Run Migrations](https://github.com/cisagov/dotgov/actions/runs/2876938716)

See ADRs and git commit messages.

Closes #12 
Closes #13


.github/workflows/migrate.yaml
```diff
@@ -0,0 +1,34 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run Django migrations for sandbox
+        uses: 18f/cg-deploy-action@main
+        with:
+          cf_username: ${{ secrets.CF_USERNAME }}
+          cf_password: ${{ secrets.CF_PASSWORD }}
+          cf_org: sandbox-gsa
+          cf_space: dotgov-poc
+          full_command: "cf run-task getgov-dev --wait --command 'python manage.py migrate' --name migrate"
```
loganmeetsworld, line 31: I am not super familiar with this pattern - are we talking about database migrations here or db migrations + other tasks that need to be run after startup?

docs/architecture/decisions/0007-python-buildpack.md
```diff
@@ -0,0 +1,28 @@
...
+## Status
+
+Accepted
+
+## Context
+
+We had previously drafted ADRs to use Docker to build images for containerized deployment. The motivation was to reduce dependency on Cloud.gov. These ADRs were rejected, mainly owing to the necessity of having an image repository to hold the built images. The team had concerns about an overreliance on GitHub with their GitHub Packages service and about the overall maintenance burden of any image repository.
+
+Cloud.gov uses Cloud Foundry which provides several “buildpacks”. These are automated environments which will take a code repository of a certain language and do the usual setup steps to prepare a deployment of that code. In the case of Python, this means automated detection of Pipfile and installation of packages.
+
```
loganmeetsworld, line 14: Want to add that we didn't think we'd need a "custom" buildpack since the current use case is so simple, as a reason too?

ops/README.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+After authenticating, make sure you are targeting the correct org and space!
+
+```bash
+cf spaces
+cf target -o <ORG> -s <SPACE>
+```
+
+## Rotating Secrets
```
loganmeetsworld, line 23: Can we specify "environment secrets" here?

ops/README.md
```diff
@@ -0,0 +1,61 @@
...
+cf restage getgov-dev --strategy rolling
+```
+
+Non-secret environment variables can be declared in `manifest-<ENVIRONMENT>.json` directly.
+
+## Database
+
+In sandbox, created with `cf create-service aws-rds micro-psql getgov-database`.
+
+Binding the database in `manifest-<ENVIRONMENT>.json` automatically inserts the connection string into the environment as `DATABASE_URL`.
```
loganmeetsworld, line 59: This is great this works like this.

src/docker_entrypoint.py
```diff
@@ -11,7 +11,7 @@
 import sys
 import os
 import pwd
-import subprocess  # nosec
+import subprocess
 
 HOST_UID = os.stat("/app").st_uid
 HOST_USER = "james"
```
loganmeetsworld, line 17: ?

src/registrar/config/urls.py
```diff
@@ -0,0 +1,23 @@
...
+"""
+
+from django.conf import settings
+from django.contrib import admin
+from django.urls import include, path
+
+from registrar.views import health
+
+urlpatterns = [
+    path("admin/", admin.site.urls),
```
loganmeetsworld, line 14: Is this where django puts its admin interface?

ops/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+[Cloud.gov RDS documentation](https://cloud.gov/docs/services/relational-database/).
```
loganmeetsworld, line 61: One other thing we might add is the commands for tailing logs/debugging issues with the django app running in cloud.gov?

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,37 @@
+name: Build and deploy
+
+on:
+  push:
+    branches:
+      - main
+
+    tags:
+      - v*
```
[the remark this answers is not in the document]
SSPJ: It is saying to run this action when either a new commit(s) are pushed to the main branch or new commit(s) are pushed with tags attached which start with the letter "v".

The file has multiple jobs, as you observed; they are intended for deploying to sandbox and staging (although staging doesn't exist yet).

The workflow I envisioned was that merged pull requests are automatically deployed to sandbox/unstable, while deploys to staging require tagging a commit with a version number (e.g. `v0.0.2`). That can be done via the CLI but more typically would be done via [GitHub UI](https://github.com/cisagov/dotgov/tags).

I've added comments to explain this better in the file.

.github/workflows/migrate.yaml
SSPJ: Just database migrations. It would be nice to be able to run them from a click of button in GitHub, but for some reason the manual workflow button isn't showing up for me. Maybe CISA's org settings?

A common pattern is to wrap the migrations into the startup command, but if we are running replicated instances, that's probably not the greatest idea. (Besides which, zero downtime deploys is not likely to ever be a top priority for the registrar.)

Much easier to put the app into maintenance mode, migrate the database, then deploy the new code. (That process could even be automated, I bet.)

ops/README.md
SSPJ: Done.

src/registrar/config/urls.py
SSPJ: Yep.

src/docker_entrypoint.py
SSPJ: It is important that the UUID maps correctly (that's what makes this script work) but the username doesn't matter. The original script that I lifted from [django-template](https://github.com/18F/django-template/blob/main/django_template/templates/docker_entrypoint.py.jinja) used `HOST_USER = env.get_credential("HOST_USER", "{{ app_name }}_user")`

ops/README.md
SSPJ: Good point. I've added a todo. Tailing the logs won't mean much until we get the logging config cleaned up, which is another todo. 😁

Issue #11, opened 2022-08-04
Setup cloud.gov instance


SSPJ:
Use GSA sandbox (might have some limitations).

- [x] Use existing 18F GitHub Action templates
    - [x] Draft ADR for CI/CD choice
- [x] Address and document considerations for deploying to test/stage/prod environments (partially done)
- [x] Draft ADR for docker container
    Address trade-offs with buildpack
- [ ] Draft ADR for cloud.gov itself
- [x] Ensure that who wants to deploy can from CLI (seamus, logan, cameron, anyone)
- [x] Assess if limitations of sandbox

Completing this task has a dependency on having _something_ to deploy. Consider finishing #12 first.

loganmeetsworld:
After some investigating, it makes sense to setup a CISA prototyping org for cloud.gov now. I am pursuing that.

loganmeetsworld:
I've setup a space in our sandbox called `dotgov-poc` and added @SSPJ as a developer. I am pursuing setting up an actual space for us to deploy to as well.

loganmeetsworld:
One limitation of the sandbox is that @h-m-f-t cannot join without a GSA email. I am in the process of setting up a prototyping org that can live with CISA and have us work together but in the meantime we will continue to work out of the `sandbox-gsa/dotgov-poc` space.

PR #43, opened 2022-08-12
add script for rotating cloud.gov secrets and runbook for description of script

loganmeetsworld:
# Setup Github Cloud.gov CD Secrets Rotate  #

## 🗣 Description ##

This adds a script and documentation for the process of rotating our cloud.gov secrets, which is something we must routinely do but is also important to have a runbook for rotating if the secrets are compromised.

You can see the repository secrets successfully get set here: 

Screen Shot 2022-08-12 at 9 22 49 AM

This will not be used until we setup a github action. We can decide at that point which environments we'd like to CD to and what criteria we have for deployment. 

Only people with admin access (SpaceDevelopers) will be able to run this script.

### DIFF
--- a/scripts/rotate_cloud_secrets.sh
+++ b/scripts/rotate_cloud_secrets.sh
@@ -1,0 +1,46 @@
+if brew ls --versions jq gh > /dev/null; then
+  continue
+else
+  echo "jq and gh packages must be installed. Please run 'brew install jq gh'"
+  exit 1
+fi
+
+if cf --version > /dev/null; then
+  continue
+else
+  echo "The cf CLI must be installed. Please see more information: https://docs.cloudfoundry.org/cf-cli/install-go-cli.html"
+  exit 1
+fi
+
+cf spaces
+read -p "Are you logged in to the dotgov-poc CF space above? (y/n) " -n 1 -r
+echo
+if [[ ! $REPLY =~ ^[Yy]$ ]]
+then
+    cf login -a https://api.fr.cloud.gov --sso
+fi
+
+gh auth status
+read -p "Are you logged into a Github account with access to cisagov/dotgov? (y/n) " -n 1 -r
+echo
+if [[ ! $REPLY =~ ^[Yy]$ ]]
+then
+    cf login -a https://api.fr.cloud.gov --sso
+fi
+
+echo "Great, removing and replacing Github CD account..."
+cf delete-service-key github-cd-account github-cd-key
+cf create-service-key github-cd-account github-cd-key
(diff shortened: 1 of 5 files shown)


loganmeetsworld:
Part of #11

scripts/rotate_cloud_secrets.sh
```diff
@@ -0,0 +1,46 @@
+if brew ls --versions jq gh > /dev/null; then
+  continue
+else
```
SSPJ, line 3: Would suggest making this line more portable, since future maintainers may not have brew on their OS.

```suggestion
if [ ! $(command -v gh) ] || [ ! $(command -v jq) ]; then
```

scripts/rotate_cloud_secrets.sh
```diff
@@ -0,0 +1,46 @@
...
+cf create-service-key github-cd-account github-cd-key
+cf service-key github-cd-account github-cd-key
+read -p "Please confirm we should set the above username and key to Github secrets. (y/n) " -n 1 -r
+echo
+if [[ ! $REPLY =~ ^[Yy]$ ]]
+then
+    exit 1
+fi
+
+cf service-key github-cd-account github-cd-key | sed 1,2d  | jq -r '[.username, .password]|@tsv' |
```
SSPJ, line 42: I had to change this `jq` command to get it to work. I figured it was a typo in the PR but could you double check on your end?

This is what the output of `cf` looks like for me:

```json

{
  "credentials": {
    "password": "hhhhhhhhhh",
    "username": "hhhhhhhhhh"
  }
}

```
```suggestion
cf service-key github-cd-account github-cd-key | sed 1,2d  | jq -r '.credentials|[.username, .password]|@tsv' | 
```

SSPJ (approved):
Nice work! 👍 

I had one issue that prevented the script running and an additional suggestion for interoperability. Good to merge once feedback is addressed.

scripts/rotate_cloud_secrets.sh
loganmeetsworld: Ahhh this might be a difference in versions of `cf`!

loganmeetsworld: What version of cf are you running?

SSPJ: That was it! The version at 4d44c86 works fine when I use `cf version 7.5.0+0ad1d63.2022-06-04`.

Since this script will break on future version of `cf`, I'd recommend indicating the fix in a code comment to speed up future debugging.

PR #54, opened 2022-08-18
add notes about commit signing and required installations to onboarding template

loganmeetsworld:
# Update onboarding template #

## 🗣 Description ##

This adds some notes about commit signing to the onboarding template along with required installations.


SSPJ (approved):
LGTM 👍 

Let's see what @igorkorenfeld makes of it.

PR #68, opened 2022-08-22
Add proposed edits to the developer onboarding template

igorkorenfeld:
## 🗣 Description ##

This PR proposes some changes to the developer onboarding instructions based on my experience going through them. In particular, it updates the Access section.

There were a couple of instructions that seemed like they were steps for me to take as an onboardee, but were actually steps for the onboarder or admin. I've added subsections to clarify the actor for each step. 

The other thing that confused and tripped me up was the Login sandbox account. This was my first time using the sandbox, and I initially did not understand that this was different from my regular Login account. I think it may be helpful to add a note clarifying that.

I did run into other issues that I will note below but I'm not sure if we want to incorporate  any learnings from them into the issue template. They may be just individual troubleshooting issues. 

### Other notes / Troubleshooting
- When trying cf CLI, I got an error that my xcode dev tools were outdated and it took a little while to get the newest ones installed (had to clear up storage space and wait for download install)
- I did not have gnupg installed, and it wasn't clear right away which tool I needed to install to run the `gpg --version` command.
- When I did install gnupg, I was able to generate my keys, but then I could not actually sign a commit until I added these two lines to my `.zshrc`:
```zsh
GPG_TTY=$(tty)
export GPG_TTY
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/developer-onboarding.md
+++ b/.github/ISSUE_TEMPLATE/developer-onboarding.md
@@ -34,6 +34,16 @@ cf login -a api.fr.cloud.gov  --sso
  **Note:** As mentioned in the [Login documentation](https://developers.login.gov/testing/), the sandbox Login account is different account from your regular, production Login account. If you have not created a Login account for the sandbox before, you will need to create a new account first.
 - [ ] Setup [commit signing in Github](#setting-up-commit-signing) and with git locally.
 
+ **Note:** if you are on a mac and not able to successfully create a signed commit, getting the following error:
+```zsh
+error: gpg failed to sign the data
+fatal: failed to write commit object
+```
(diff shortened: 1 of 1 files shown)


loganmeetsworld (approved):
This is great! Do you think we should add the `tty` export to the documentation here as well?

igorkorenfeld:
I think it could be helpful. I don't know if it's mac/unix specific, but what do you think about this note after the github commit signing step?:


> Note: if you are not able to successfully create a signed commit and get the following error
>```zsh
>error: gpg failed to sign the data
>fatal: failed to write commit object
>```
> You may need to add the following two lines to your shell's rc file (e.g. `.bashrc` or `.zshrc`)
>```zsh
>GPG_TTY=$(tty)
>export GPG_TTY
>```

loganmeetsworld:
Yep! That looks great. It is probably mac specific, we can note that.

Issue #82, opened 2022-08-23
Address local runs of tests in `dev`


loganmeetsworld:
We need a way for people to run the things that run in CI locally so they don't have to go through CI over and over. Options: 

- A template pre-commit hook that runs the same things as in CI (would have to be added individually) 
- Use something like [`act`](https://github.com/nektos/act) to run the github actions locally
- A script that is used locally and in CI
- A requirements-dev.txt file and freeze may be necessary in all cases to pin requirements in dev and CI.

Issue #81, opened 2022-08-23
Add bandit Python security scanning to CI


loganmeetsworld:
Find and address security issues in Python: https://bandit.readthedocs.io/en/latest/

Address anything and set a standard for upgrading.

loganmeetsworld:
Covered in #85

Issue #75, opened 2022-08-22
Setup CI pipeline to gate pull requests


loganmeetsworld:
CI should run on every branch and should be able to include 
- Linting
- Pytest tests

We will run this with Github actions.

loganmeetsworld:
Addressed in #85

SSPJ:
Edited to remove Accessibility testing, as that will be tackled in #72

loganmeetsworld:
This has been setup in #85, will be extended in other issues!

PR #85, opened 2022-08-24
Setup initial CI gating on tests and add linting tests

### LINKED ISSUE #76: Add linting to CI pipeline
Choose a [linting tool ](https://books.agiliq.com/projects/essential-python-tools/en/latest/linters.html) and make it part of CI.

loganmeetsworld:
## 🗣 Description ##

This adds a couple tools for us to gate our pull requests to `main`: 

* `flake` for basic python linting
* `black` for formatting consistency
* `mypy` for static type checking

This will close #76 and #75. 

## Next

* Add Bandit in #81 and address some security concerns presented by the docker entrypoint. 
* A solution for running these locally will be added in response to #82.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -25,7 +25,7 @@ jobs:
           pip install bandit black flake8 mypy types-requests
       - name: Lint with flake8
         working-directory: ./src
-        run: flake8 . --count --max-complexity=10  --show-source --statistics --max-line-length=120
+        run: flake8 . --count --show-source --statistics
       - name: Check formatting with black
         working-directory: ./src
         run: black --check .
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -29,9 +29,6 @@ jobs:
       - name: Check formatting with black
         working-directory: ./src
         run: black --check .
-      - name: Run bandit
-        working-directory: ./src
-        run: bandit -r .
       - name: Run type checking
         working-directory: ./src
         run: mypy .
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,37 @@
+# test.yml
+name: Testing
+
+on:
+  push:
+    branches:
+    - main
+  pull_request:
+    branches:
+    - main
+
+jobs:
+  python-linting:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.10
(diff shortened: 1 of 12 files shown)


.github/workflows/test.yaml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+  python-linting:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v3
+        with:
+          python-version: "3.10"
```
loganmeetsworld, line 21: I don't know if there is any better way to pin a version to what we'll use in docker compose?

src/.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
+max-line-length = 120
+max-complexity = 10
```
loganmeetsworld, line 3: Flake doesn't support pyproject.toml so an extra configuration file it is!

src/registrar/config/settings.py
```diff
@@ -43,11 +43,8 @@
 DEBUG = env.bool("DJANGO_DEBUG", default=False)
 
 # TODO: configure and document security settings
-ALLOWED_HOSTS = [
-    'getgov-unstable.app.cloud.gov',
-    'get.gov'
-]
-ALLOWED_CIDR_NETS = ['10.0.0.0/8']  # nosec
+ALLOWED_HOSTS = ["getgov-unstable.app.cloud.gov", "get.gov"]
```
SSPJ, line 46: I have discovered that this irritating* folding can be avoided by [use of a trailing comma](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#the-magic-trailing-comma).

Not something to fix here, as I am doing a bunch of work in this file, just sharing FYI.

\* I call it irritating because it breaks the fundamental code formatting (& design) principle of "similar things should look similar, different things should look different".

SSPJ (approved):
This looks really great, thank you for installing these! 👍

.github/workflows/test.yaml
```diff
@@ -0,0 +1,34 @@
+# test.yml
+name: Testing
+
+on:
+  push:
+    branches:
+    - main
+  pull_request:
+    branches:
+    - main
```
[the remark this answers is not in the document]
loganmeetsworld: Done!

.github/workflows/test.yaml
loganmeetsworld: Done!

PR #100, opened 2022-09-02
Write Django cmd for running lint

### LINKED ISSUE #82: Address local runs of tests in `dev`

SSPJ:
## 🗣 Description ##

A script to allow running linters on dev. And documentation.

Devs need to be able to check their work before pushing it to GitHub.

Closes #82

### DIFF
--- a/docs/developer/README.md
+++ b/docs/developer/README.md
@@ -34,20 +34,24 @@ You can change the logging verbosity, if needed. Do a web search for "django log
 
 ## Running tests
 
+Crash course on Docker's `run` vs `exec`: in order to run the tests inside of a container, a container must be running. If you already have a container running, you can use `exec`. If you do not, you can use `run`, which will attempt to start one.
+
+To get a container running:
+
+```shell
+cd src
+docker-compose build
+docker-compose up -d
+```
+
 Django's test suite:
 
 ```shell
-docker-compose up -d
 docker-compose exec app ./manage.py test
-docker-compose stop
 ```
 
 Linters:
 
 ```shell
-docker-compose up -d
-docker-compose exec app ./manage.py test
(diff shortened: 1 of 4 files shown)


docs/developer/README.md
```diff
@@ -34,4 +34,20 @@ You can change the logging verbosity, if needed. Do a web search for "django log
 
 ## Running tests
 
-tbd.
+Django's test suite:
```
loganmeetsworld, line 37: Add a note to make sure Docker is running and that this has to be run from `src`?

docs/developer/README.md
```diff
@@ -34,4 +34,20 @@ You can change the logging verbosity, if needed. Do a web search for "django log
...
+docker-compose up -d
+docker-compose exec app ./manage.py test
+docker-compose stop
+```
+
+Linters:
+
+```shell
+docker-compose up -d
+docker-compose exec app ./manage.py test
```
loganmeetsworld, line 49: Should this be `lint`?

loganmeetsworld (approved):
Works for me! A couple notes.

docs/developer/README.md
SSPJ: Added!

docs/developer/README.md
SSPJ: Yes! Vielen Dank!

PR #104, opened 2022-09-07
Add testing to the CI pipeline

### LINKED ISSUE #77: Add pytest to CI pipeline
Add pytest as runner to CI pipeline and confirm successfully runs. Not necessary to add individual tests, just prepare the runner.

SSPJ:
## 🗣 Description ##

Adding testing to the CI. Initial ticket specified to use pytest. However, integration of pytest with Django still relies on third party packages.

Therefore, path of least resistance is to use Django's own test runner, which uses [unittest](https://docs.python.org/3/library/unittest.html) under the hood.

Closes #77

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -33,3 +33,40 @@ jobs:
       - name: Run bandit security scanning
         working-directory: ./src
         run: bandit -r .
+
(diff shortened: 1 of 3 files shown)


loganmeetsworld (approved):
Looks great!

PR #116, opened 2022-09-12
Add compliance documentation to source control 

loganmeetsworld:
## 🗣 Description ##

This adds compliance documentation from the 18F [compliance template](https://github.com/GSA-TTS/compliance-template). 

We need some way to track our progress with the SSP and why not do it in the same place we develop! I have pulled in the NIST 800-53 moderate baseline profile as a start, but there may be other catalogues we want to add to this profile. We can start documenting them as a team in Markdown. 

This uses [trestle](https://ibm.github.io/compliance-trestle/) as the tool for translating between markdown and valid OSCAL in JSON. For developers, all you need to do is document in Markdown. For now, we can ignore all the extra baggage that comes with trestle and OSCAL. But it will hopefully help us create an SSP down the line. 

## For Reviewers

Most of this can be ignored – focus on the README and other top-level markdown files under docs/compliance.


docs/compliance/Pipfile
```diff
@@ -0,0 +1,12 @@
...
+verify_ssl = true
+name = "pypi"
+
+[packages]
+compliance-trestle = "*"
+
+[dev-packages]
+
+[requires]
+python_version = "3.9"
```
neilmb, line 12: I think that we are using 3.10 in our app, can we use the same major version here for consistency?

neilmb (approved):
Otherwise, this doesn't feel like it needs too much detailed review until we start using it, but hooray for compliance-template.

PR #123, opened 2022-09-14
Add asset compilation and collection before deploy

igorkorenfeld:
## 🗣 Description ##

This PR adds two steps to the deploy GitHub Action that compile the USWDS assets and collect them into the `/public` folder using our docker container before deploying them to staging. By using the GH action and the docker container we can avoid having to run `collectstatic` on the server upon deploying. 

This closes #120 

https://getgov-staging.app.cloud.gov/

PR #129, opened 2022-09-15
Add build and deploy scripts

igorkorenfeld:
## 🗣 Description ##

This PR adds a build and deploy scripts as suggested in the discussion in #123. 

The build script compiles the css, copies over the static assets, and collects into the `/public` folder. The deploy script runs the build script and then pushes the up the files to `unstable`.

The PR also updates the corresponding documentation. 

These two scripts make it easier to compile the assets and deploy them to unstable. They also remove the need to run `collectstatic` on the server.  In the future we could add the `build.sh` script to the deploy GitHub Action. 

## 🤔 Question ##
I created a symlink in the `src` folder to link to the scripts in the `ops/scripts` folder. This is because the paths used in the scripts assume they are run from the `src` folder, and as a convenience when trying to run them from that src (i.e. so that one doesn't have to type `../ops/<script>`.  Is there any issue with that? Should I use a different path in the scripts?

### DIFF
--- a/docs/operations/README.md
+++ b/docs/operations/README.md
@@ -49,9 +49,10 @@ cf target -o cisa-getgov-prototyping -s unstable
 cf push getgov-unstable -f ops/manifests/manifest-unstable.yaml
 cf run-task getgov-unstable --command 'python manage.py migrate' --name migrate
 ```
+Alternatively, you could run the `deploy.sh` script in the `/src` directory to build the assets and deploy to `unstable`. Similarly, you could run `bash.sh` script to just compile and collect the assets without deploying.
 
 
 ## Serving static assets
 We are using [WhiteNoise](http://whitenoise.evans.io/en/stable/index.html) plugin to serve our static assets on cloud.gov. This plugin is added to the `MIDDLEWARE` list in our apps `settings.py`.
 
-Note that it’s a good idea to run `collectstatic` locally or in the docker container before pushing files up to `unstable`. This is because `collectstatic` relies on timestamps when deciding to whether to overwrite the existing assets in `/public`. Due the way files are uploaded, the compiled css in the `/assets/css` folder on `unstable` will have a slightly earlier timestamp than the files in `/public/css`, and consequently running `collectstatic` on`unstable` will not update `public/css` as you may expect.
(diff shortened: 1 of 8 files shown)


docs/operations/README.md
```diff
@@ -49,9 +49,10 @@ cf target -o cisa-getgov-prototyping -s unstable
 cf push getgov-unstable -f ops/manifests/manifest-unstable.yaml
 cf run-task getgov-unstable --command 'python manage.py migrate' --name migrate
 ```
+Alternatively, you could run the `deploy.sh` script in the `/src` directory to build the assets and deploy to `unstable`. Similarly, you could run `bash.sh` script to just compile and collect the assets without deploying.
```
neilmb, line 52: I think you mean `build.sh` not `bash.sh`?
```suggestion
Alternatively, you could run the `deploy.sh` script in the `/src` directory to build the assets and deploy to `unstable`. Similarly, you could run `build.sh` script to just compile and collect the assets without deploying.
```

src/registrar/assets/sass/_theme/_uswds-theme-custom-styles.scss
```diff
@@ -26,3 +26,7 @@ i.e.
 p {
   color: color('blue-10v');
 }
+
```
neilmb, line 29: This doesn't seem as if it's related to the build scripts? It isn't a huge deal, but generally I would want every change that appears in a PR to be related somehow.

src/run.sh
```diff
@@ -15,5 +15,4 @@ else
   fi
 fi
 
-python manage.py collectstatic --settings=registrar.config.settings --noinput
```
neilmb, line 18: Maybe it's worth replacing this command with a comment that explains that the site will only work right if `collectstatic` has been run somewhere else, just as a breadcrumb for the future?

ops/README.md
```diff
@@ -3,4 +3,4 @@
 
 This directory contains files related to deploying or running the application(s).
 
-Documentation is in [docs/ops](../docs/ops).
\ No newline at end of file
+Documentation is in [docs/ops](../docs/operations/).
```
neilmb, line 6: Thank you for changing this, I noticed it was in error the other day but didn't take the time to fix it.

neilmb (approved):
LGTM, thanks for quickly doing all of the follow-on tickets for this. It's making our infrastructure much more solid.

src/registrar/assets/sass/_theme/_uswds-theme-custom-styles.scss
igorkorenfeld: That makes sense in general. For this case I did it to make sure the style change was actually applied when I ran the deploy script. So would you typically revert or just not commit the change?

neilmb: I would typically back out a change that I made just to test that something happens. I would either make another commit that undoes the change so that it doesn't show up as a change in the overall PR, or not commit it in the first place.

PR #139, opened 2022-09-22
Add Login.gov

SSPJ:
## 🗣 Description ##

Make Login.gov function in the Django backend.

This includes a wrapper application which handles the Private Key JWT method of authenticating via OpenID Connect using the underlying pyoidc library.

See ADR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,11 @@ docs/research/data/**
 public/
 credentials*
 
+*pem
+*crt
+
+*bk
+
 ### The usual garbage files ###
 
 # Byte-compiled / optimized / DLL files
--- a/.gitignore
+++ b/.gitignore
@@ -6,10 +6,10 @@ docs/research/data/**
 public/
 credentials*
 
-*pem
-*crt
+*.pem
+*.crt
 
-*bk
+*.bk
 
 ### The usual garbage files ###
 
--- a/src/djangooidc/views.py
+++ b/src/djangooidc/views.py
@@ -1,0 +1,116 @@
+# coding: utf-8
+
+import logging
+
+try:
+    from urllib.parse import parse_qs, urlencode
+except ImportError:
+    from urllib import urlencode
+    from urlparse import parse_qs
+
+
+from django.conf import settings
+from django.contrib.auth import logout as auth_logout
+from django.contrib.auth import authenticate, login
+from django.http import HttpResponseRedirect
+from django.shortcuts import redirect, render
+from djangooidc.oidc import Client
+from djangooidc import exceptions as o_e
+
+
+logger = logging.getLogger(__name__)
+
+# Initialize provider using pyOICD
+OP = getattr(settings, "OIDC_ACTIVE_PROVIDER")
+CLIENT = Client(OP)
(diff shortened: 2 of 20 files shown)


.gitignore
```diff
@@ -6,6 +6,11 @@ docs/research/data/**
 public/
 credentials*
 
+*pem
```
neilmb, line 9: **suggestion (non-blocking)**: If these are meant to match extensions, maybe include the dot in the pattern for microscopic improvement in false matches?

src/djangooidc/views.py
```diff
@@ -0,0 +1,120 @@
...
+        logger.info("Successfully logged out user %s" % username)
+        next_page = getattr(settings, "LOGOUT_REDIRECT_URL", None)
+        if next_page:
+            request.session["next"] = next_page
+
+
+def logout_callback(request):
+    """Simple redirection view: after logout, redirect to `next`."""
+    next = request.session["next"] if "next" in request.session.keys() else "/"
+    return redirect(next)
```
neilmb, line 120: **thought**: I have had problems not getting to the `next` page that I expected. I wonder if we can do any testing of this callback to ensure that it is working as expected.

src/docker-compose.yml
```diff
@@ -10,10 +10,6 @@ services:
       - db
     working_dir: /app
     entrypoint: python /app/docker_entrypoint.py
-    deploy:
```
neilmb, line 13: **question**: Is there a reason to delete this block in this PR?

src/registrar/config/urls.py
```diff
@@ -11,9 +11,12 @@
...
+    path("", index.index, name="home"),
     path("admin/", admin.site.urls),
-    path("", index.index),
     path("health/", health.health),
+    path("openid/", include("djangooidc.urls")),
+    # these views respect the DEBUG setting
+    path("__debug__/", include("debug_toolbar.urls")),
 ]
 
 if settings.DEBUG:
```
neilmb, line 22: **suggestion**: I think that we can get rid of the entire `if` block here because adding the url patterns above does the same thing in all cases.

neilmb (approved):
LGTM, so good to see this.

.gitignore
SSPJ: Good suggestion, done.

src/docker-compose.yml
SSPJ: No reason other than it annoys me. 😁  I've added it back.

src/registrar/config/urls.py
SSPJ: Good catch. This was leftover mess from a less-than-careful rebase.

src/djangooidc/views.py
SSPJ: I appreciate you asking for code tests. It will take a bit of time, but I will add them.

I'm unsure if it is possible to do arbitrary `next` redirection with Login.gov, due to the security implications. I'll have to do a bit of reading and experimentation.

SSPJ:
As discussed with Neil, going to go ahead and merge this and continue some follow up work on an additional PR. See #144 .

PR #140, opened 2022-09-23
Update product_strategy.md

calerubensteingsa:
#  #

### DIFF
--- a/docs/product/product_strategy.md
+++ b/docs/product/product_strategy.md
@@ -14,7 +14,9 @@ CISA intends to create a modern, user-centered, responsive web application to en
 
 
 ## Problem Statements
-CISA and U.S.-based government organizations and publicly controlled entities lack a clear, usable, and efficient way to register, maintain, secure, and operate a .gov domain.
+U.S.-based government organizations and publicly controlled entities lack a clear, usable, and efficient way to apply, register, and a .gov domain and related infrastructure in order to build public trust of their website and communications.
+
+CISA lacks a scalable, efficient, and secure method of managing the outreach and operations of .gov TLD program in order to facilitate government agencies building public trust of their website and communications.
 
 ## Short-term Success for .gov
 * A production-ready, modern .gov registrar that can replace the current system with improved user experience and operational efficiency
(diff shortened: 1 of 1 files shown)


docs/product/product_strategy.md
```diff
@@ -12,7 +14,7 @@ _TBD - once we synthesize initial research and align as a product team_
 
 
 ## Problem Statements
-_TBD - once we synthesize initial research and prioritize areas of need_
+The US federal, state, and local governments find the process of provisioning, managing, and securing .gov TLDS to be time consuming, confusing, risky, and inefficient.
```
h-m-f-t, line 17: I'd suggest something like:
> The .gov registrar should be the primary interaction point between _.gov registrants_ and _CISA’s .gov program_, and between _registrants_ and _their DNS_. It's not.

Or, ha, if that's too not-a-problem-statement-y:
> U.S.-based government organizations and publicly controlled entities lack a clear, usable, and efficient way to register, maintain, and operate a .gov domain.

docs/product/product_strategy.md
```diff
@@ -12,7 +14,7 @@ _TBD - once we synthesize initial research and align as a product team_
 
 
 ## Problem Statements
-_TBD - once we synthesize initial research and prioritize areas of need_
+CISA and U.S.-based government organizations and publicly controlled entities lack a clear, usable, and efficient way to register, maintain, secure, and operate a .gov domain.
```
calerubensteingsa, line 17: Incorporated Cameron's changes and updated slightly

calerubensteingsa (commented):
Added Cameron's suggested problem statement and updated slighlty

docs/product/product_strategy.md
h-m-f-t: > CISA and

I think your intent in adding this was to highlight the .gov registry's need to have a place to evaluate domain requests and interact with requestors – and that's certainly part of the problem we're solving for. But I think adding "CISA and" to this statement is potentially confusing, since we don't "register, maintain, secure, and operate" individual domains like gov't orgs do.

calerubensteingsa: Makes sense - my main reason for adding CISA and "secure" to the problem statement is that CISA's needs, both as users and the eventual security needs (like hosting DNS) are represented. As it reads to me now, the problem statement seemed focused primarily on the registrant users while I think there is some problem space to represent for CISA. We can always split the problem statement into two - however you think best represents the mission!

h-m-f-t: Yeah, I think splitting the problem statement into two actually would make the most sense, one registrant-side and one admin-centric.

PR #142, opened 2022-09-23
User and UserProfile data models

### LINKED ISSUE #50: Prototype database schema
_Outcome_: Update the current data modeling PR for review, and merge initial data models to support login.gov flow. 

- [ ] #92 
- [ ] #119 (not high-priority)

neilmb:
# Add custom User and related UserProfile model

## 🗣 Description ##

This adds data models for a User and a one-to-one related UserProfile. The User model is un-customized from Django's default, but Django recommends using a custom model because we can't change it after our very first set of migrations.

There is also a very, very basic page at `/edit_profile/` that allows users to edit just one field of the UserProfile model, their "Display Name". It isn't intended that this page will necessarily be part of the Get.gov product, that would take more product discussion and design work. Rather it exercises the patterns for making custom forms in Django that we are going to be using over and over again on this project.

Screen Shot 2022-09-23 at 13 47 22

Closes #50 for the very first stages of making a database schema.

This PR will require developers to re-run migrations, possibly on a clean database inside of their Docker container.  The command `docker compose run --rm db psql -d postgres://user@db/app` can be used to get access to a SQL shell in the local db where `DROP DATABASE` and `CREATE DATABASE` can be used.


src/registrar/admin.py
```diff
@@ -0,0 +1,15 @@
...
+from django.contrib.auth.admin import UserAdmin
+from .models import User, UserProfile
+
+
+# edit a user's profile on the user page
+class UserProfileInline(admin.StackedInline):
+    model = UserProfile
+
+
+class MyUserAdmin(UserAdmin):
```
SSPJ, line 11: Docstring.

src/registrar/apps.py
```diff
@@ -0,0 +1,17 @@
...
+    name = "registrar"
+
+    def ready(self):
+        """Runs when all Django applications have been loaded.
+
+        We use it here to load signals that connect related models.
+        """
+        # noqa here because we are importing something to make the signals
+        # get registered, but not using what we import
+        from . import signals  # noqa
```
SSPJ, line 17: Kudos for explaining the `#noqa`. ⭐

src/registrar/config/settings.py
```diff
@@ -54,6 +54,9 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = env_debug
 
+# Use our user model instead of the default
+AUTH_USER_MODEL = "registrar.User"
```
SSPJ, line 58: Put this in the section with the other database settings?

src/registrar/forms.py
```diff
@@ -0,0 +1,10 @@
+from django import forms
+
+from .models import UserProfile
+
+class EditProfileForm(forms.ModelForm):
```
SSPJ, line 5: Docstring.

It can feel redundant to put class documentation on a class which is so obviously self-explanatory, but future maintainers may not be as familiar with Django as you are. It is a kindness to them to include a brief but clear explanation of why a piece of code exists.

src/registrar/migrations/0003_alter_userprofile_street1.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 4.1.1 on 2022-09-22 16:53
+
+from django.db import migrations, models
```
SSPJ, line 3: As this is the first PR to include migrations, could I trouble you to squash them? (Just delete and regenerate.) Unless you believe they are logically better represented as three files. That's fine as well.

src/registrar/models/models.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class User(AbstractUser):
+    """
+    A custom user model that performs identically to the default user model
+    but can be customized later.
+    """
+
+    def __str__(self):
+        if self.userprofile.display_name:
```
SSPJ, line 12: What happens if `self.userprofile` is null?

src/registrar/models/models.py
```diff
@@ -0,0 +1,81 @@
...
+    fax = models.TextField(blank=True)
+    email = models.TextField(blank=True)
+
+    class Meta:
+        abstract = True
+        # don't put anything else here, it will be ignored
+
+
+class UserProfile(TimeStampedModel, ContactModel, AddressModel):
+    user = models.OneToOneField(User, null=True, on_delete=models.CASCADE)
```
SSPJ, line 74: Is the `on_delete` attribute uni-directional for a `OneToOneField`? (Deleting UserProfile deletes User, but deleting User does _not_ delete UserProfile.) I assume so, but want to confirm.

src/registrar/models/models.py
```diff
@@ -0,0 +1,81 @@
...
+    display_name = models.TextField()
+
+    def __str__(self):
+        if self.display_name:
+            return self.display_name
+        else:
+            return self.user.username
```
SSPJ, line 81: What will happen if `self.user` is null?

src/registrar/signals.py
```diff
@@ -0,0 +1,17 @@
+from django.db.models.signals import post_save
+from django.contrib.auth.models import User
+from django.dispatch import receiver
+
+from .models import UserProfile
+
+
+@receiver(post_save, sender=User)
+def create_profile(sender, instance, created, **kwargs):
```
SSPJ, line 9: Docstring.

src/registrar/signals.py
```diff
@@ -0,0 +1,17 @@
...
+    if created:
+        UserProfile.objects.create(user=instance)
+
+
+@receiver(post_save, sender=User)
+def save_profile(sender, instance, **kwargs):
```
SSPJ, line 15: Docstring.

src/registrar/signals.py
```diff
@@ -0,0 +1,17 @@
...
+    # instance is a User, it has a profile from the one-to-one relation
+    instance.userprofile.save()
```
SSPJ, line 17: Suggest checking for the existence of `userprofile` and creating it if it does not exist.

src/registrar/tests/test_views.py
```diff
@@ -9,3 +12,21 @@ def test_health_check_endpoint(self):
         response = self.client.get("/health/")
         self.assertEqual(response.status_code, 200)
         self.assertContains(response, "OK")
+
+
+class LoggedInTests(TestCase):
```
SSPJ, line 17: Thank you for including tests! 🎸

src/registrar/views/profile.py
```diff
@@ -0,0 +1,20 @@
+
+from django.shortcuts import redirect, render
+from django.contrib.auth.decorators import login_required
+from django.contrib import messages
+
+from ..forms import EditProfileForm
+
+
+@login_required
+def edit_profile(request):
```
SSPJ, line 10: Docstring.

src/registrar/models/models.py
neilmb: It is uni-directional, but it goes the opposite of the direction that you specify. Even though we are specifying it here on the `UserProfile` model, `on_delete` describes what should happen when we delete the matching `User`. 

In this case, a `User` who is deleted no longer needs a `UserProfile` so we want to also remove the `UserProfile` at that time which is what `CASCADE` indicates.

https://docs.djangoproject.com/en/4.1/ref/models/fields/#django.db.models.ForeignKey.on_delete for the formal documentation which just refers to SQL's behavior and https://stackoverflow.com/a/38389488 for a better explanation.

src/registrar/models/models.py
neilmb: In this case, `self.user` should never be null because of the foreign key constraint and the cascading deletes. But for safety, I can change this to a guarded `getattr` as well.

src/registrar/models/models.py
SSPJ: Ack, thank you for the docs link! I do tend to get confused by foreign key constraints since the key is so often held in the table which is opposite to the "natural" relationship (e.g. a car has wheels, but a Wheel has the key to the Car...)

I have a [draft ADR](https://docs.google.com/document/d/1ea04AvWeXwxelxmhD2rqMfcRBnKdnxvam6EyA8ehe7I/edit#bookmark=id.po74ml55c5lk) for the business logic decisions that are going into the design of the database models (e.g. "a User who is deleted no longer needs a UserProfile"), but it is kinda messy and has met with some objections from the team. Would you be good enough to start a new one in that document and include your thinking behind the models on this PR?

SSPJ (approved):
LGTM 👍  Nice work, Neil!

I was able to test this on dev.

As we discussed in chat, there is some concern about whether the signal handler is working as expected. We can test this again once it and #139 have merged and make any necessary adjustments.

Feel free to add you ADR to this PR or open a follow up one for it.

Issue #15, opened 2022-08-04
Proof of concept for login.gov use in authentication 


SSPJ:
Outcome: An endpoint that returns `whoami` to confirm that login was successful

Authentication use cases, Keep it simple.

- [x] Setup login.gov sandbox for team
- [x] Anyone can log in (for now) set in login config
- [x] Create user model
- [x] Implement the route for `whoami`(create views/templates)
- [x] #132 
- [x] Investigate and document options for role-based access control (RBAC) from the registrant perspective
- [x] Consider testing with team members for QA

This depends on having a codebase to integrate with, so finish #12 and #13 first.

loganmeetsworld:
Here is the developer documentation for login.gov: https://developers.login.gov/

loganmeetsworld:
Turns out the FAC, a project TTS is working on, also uses login.gov (with a Django backend): [Frontend PR](https://github.com/GSA-TTS/FAC-Frontend/pull/24) and [Backend PR](https://github.com/GSA-TTS/FAC/pull/282)

loganmeetsworld:
We now have a [login.gov](http://login.gov/) app and respective team in the non-production sandbox. @SSPJ, @colinmurphy01, and @h-m-f-t have access to our app [here](https://dashboard.int.identitysandbox.gov/service_providers/2640). Weirdly I could not find DHS under the agency dropdown so I chose GSA but I'm filing a ticket with the Login team. 

Things we should be able to access in our django app for now: `Attribute bundle: email, first_name, last_name`

I also turned on identity verification and [here is some info about attributes](https://developers.login.gov/attributes/). We don't need to mess with this now as this is just about authentication, but thought it would be interesting to take a look at.

loganmeetsworld:
The endpoints and flow may rely on USWDS implementation here: https://github.com/cisagov/getgov/issues/45

neilmb:
Django's `django.contrib.auth.UserModel` is already in place by default and has `username`, `email`, `first_name`, and `last_name` fields. I think that `djangooidc` will give us users from that model and fill in their attributes by default. If we ask for the `UUID` attribute from Login.gov, it will go into the `username` field (https://github.com/koriaf/django-oidc/blob/master/djangooidc/backends.py#L25).

[...]

PR #147, opened 2022-09-28
Add logged-in /whoami page

neilmb:
# Add a /whoami page for logged-in users #

## 🗣 Description ##

Now that we can have logged-in users via Login.gov, this adds a page that those users can see their who they are logged in as. It also adds an un-styled "Sign in" or the user's username to the header.

Screen Shot 2022-09-28 at 15 00 21

This is just a small part of the effort to integrate Login.gov as a proof-of-concept described in #15


SSPJ (approved):
Overall, LGTM 👍  Great work, Neil!

I found some type errors. These didn't fail the CI due to a mistake in the way I wrote the linter. I'll push a PR in a moment to fix that.

```
registrar/models/models.py:74: error: Definition of "Meta" in base class "TimeStampedModel" is incompatible with definition in base class "ContactModel"
registrar/models/models.py:74: error: Definition of "Meta" in base class "TimeStampedModel" is incompatible with definition in base class "AddressModel"
registrar/models/models.py:74: error: Definition of "Meta" in base class "ContactModel" is incompatible with definition in base class "AddressModel"
Found 3 errors in 1 file (checked 33 source files)
```

Other thing I noticed is that I am consistently getting `Received state not the same as expected` (401) while logging in, but I do not think this is an issue with your code. There is probably something wrong with the oidc callback view; I'll have a look after this merges.

The Basics
- [x] I read the code
- [x] I ran the code
 
Documentation
    - [x] using plain language
- [x] licenses for new dependencies are recorded, if any - n/a
- [x] licenses for updated dependencies have been checked for accuracy, if any - n/a
- [x] new components have been added to the system diagram - n/a
 
Security
- [x] security false positives are documented - n/a, no automated scanning
    We are heavily relying on Django template HTML entity escaping. So, for example, if there is malicious code in `request.path` which ends up in `CANONICAL_PATH`, that code will be rendered harmless by the templating engine.
- [x] significant user actions are logged to a database table - n/a
 
Reliability
- [x] good level of error handling
    - [x] interactions with external systems are wrapped in try/except - n/a
- [x] functionality is tested with unit or integration tests ⭐ 

Accessibility
- [ ]  New pages have been added to .pa11yci file so that they will be tested with our automated accessibility testing ❌ 
- [ ] Meets WCAG 2.0 AA or 2.1 AA for Section 508 compliance ❓ 
    I am skipping manual testing, because the whois page is never intended to be part of the public facing site
 
Device Matrix
- [ ] I am skipping device testing, because the whois page is never intended to be part of the public facing site

PR #148, opened 2022-09-29
Ensure manage.py lint fails when linters fail

SSPJ:

Previously, the `manage.py lint` command would fail if any linter failed to successfully complete, but would silently pass if the linter returned a non-zero exit code. In other words, allowing any linting errors to pass the CI as long as they weren't egregious enough to throw an exception in the linter.


neilmb (approved):
That seems right now to me.

neilmb:
And of course now the tests are failing because the type errors that were merged to main are still there.

PR #150, opened 2022-09-30
Login.gov docs and tests

SSPJ:
## 🗣 Description ##

Adds documentation and tests to previous PR #139 

Due to time constraints, only tests the views, not the underlying OIDC wrapper code. It may be necessary in the future to go back and add additional tests to satisfy a coverage metric.


neilmb (changes requested):
Would love to see the `contextmanager` change, otherwise looks great to me.

neilmb (approved):

PR #160, opened 2022-10-04
Add security scanning to Github Actions

### LINKED ISSUE #80: Add initial security scanning for Django application code
- Turn on dependabot alerts 
- Investigate various options for security scanning through github actions https://github.com/marketplace/actions/django-security-check
- Contemplate adding Zap scanning:  https://github.com/18F/django-template/blob/main/django_template/templates/github/workflows/owasp-scan.yml.jinja#L44-L51

neilmb:
# Add security checks to Github actions

## 🗣 Description ##

This adds two basic security checking procedures to our CI pipeline, Django's `check --deploy` and OWASP Zap's Baseline scan. They run in a single Github workflow, but as two separate jobs that each leverage our Docker compose setup.

We want to use automated tools that monitor different aspects of our security posture. These tools aren't a panacea, but getting continuous feedback on various rule sets throughout development is a direct benefit.

Should close #80 when merged.


.github/workflows/security-check.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+env:
+  FAIL_LEVEL: WARNING
+  ENV_TYPE: pipenv
+  DEP_PATH: src/
+  APP_PATH: src/
+  EXTRA_ARGS: "--settings=registrar.config.settings"
+  DJANGO_SECRET_KEY: not-a-secret-jw7kQcb35fcDRIKp7K4fqZBmVvb+Sy4nkAGf44DxHi6EJl
+  DATABASE_URL: "postgres://not_a_user:not_a_password@not_a_host"
+  DJANGO_BASE_URL: "https://not_a_host"
```
SSPJ, line 19: Comment vars, please. I spent a while trying to figure out why OWASP wasn't failing on warnings before realizing that `FAIL_LEVEL` was intended for victoriadrake/django-security-check

.github/workflows/security-check.yaml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Check out
+        uses: actions/checkout@v3
+      - name: OWASP scan
+        run: docker compose run owasp
+        working-directory: ./src
+      - name: Upload output
+        uses: actions/upload-artifact@v2
+        with:
+          name: owasp-scan-output
+          path: ./src/zap_report.html
```
SSPJ, line 51: This is not working as intended, perhaps because the path is wrong or maybe because the file is only being written within a container.

src/docker-compose.yml
```diff
@@ -71,3 +71,11 @@ services:
       - app
     profiles:
       - pa11y
+
+  owasp:
+    image: owasp/zap2docker-weekly
```
SSPJ, line 76: Stable vs weekly?

src/zap.conf
```diff
@@ -0,0 +1,121 @@
...
+10052	WARN	(X-ChromeLogger-Data (XCOLD) Header Information Leak - Passive/beta)
+10053	WARN	(Apache Range Header DoS (CVE-2011-3192) - Active/beta)
+10054	WARN	(Cookie without SameSite Attribute - Passive/release)
+10055	WARN	(CSP - Passive/release)
+10056	WARN	(X-Debug-Token Information Leak - Passive/release)
+10057	WARN	(Username Hash Found - Passive/release)
+10058	FAIL	(GET for POST - Active/beta)
+10061	WARN	(X-AspNet-Version Response Header - Passive/release)
+10062	FAIL	(PII Disclosure - Passive/beta)
+10095	IGNORE	(Backup File Disclosure - Active/beta)
```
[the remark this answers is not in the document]
trevorbryant-cisa: This is my typical experience with false positives:

1. Document any and all false positives to start,
3. Technical demonstration that the false positive cannot be exploited (ie, existing mitigations),
4. Justify the exception to the false positives (further supported documentation).

Sometimes the false positive flags can occur on inadequate rule development and conditions from the scanner's database (like wildcarding packages). In which case the demonstration includes a fault in the scanning capability itself and not the targeted technology that is scanned against.

.github/workflows/security-check.yaml
neilmb: Re-arranged so that these env vars were under the particular job they were needed for.

.github/workflows/security-check.yaml
neilmb: Removed, this seems to be an issue with the users in the OWASP Zap docker setup, but I didn't seek a workaround very hard.

src/docker-compose.yml
neilmb: Changed to `-stable`.

neilmb:
@SSPJ this should be ready for a re-review now.

SSPJ:
Fantastic work, @neilmb . 💯 Thanks for making those fixes. I'll merge as we discussed on Friday.

SSPJ (approved):

PR #162, opened 2022-10-07
Run tests behind logged in pages

### LINKED ISSUE #141: Create a method of disabling login required during testing and development
⚠ Any method of disabling login creates a "backdoor". ⚠ 

- [ ] document in an ADR security considerations and how risks have been mitigated

SSPJ:
## 🗣 Description ##

Introduces a schema for running CI scans on pages which require login.

The way this works is by having a middleware class `registrar.tests.common.MockUserLogin` which checks every request received by the application. If the user is anonymous (as a scanning tool would be, upon first accessing the application), the middleware class creates a test user and logs the requester in.

This middleware class is physically copied into the list by the CI tool, meaning that it will not be present in the running code outside of the CI environment.

Closes #141 

Here's a screen shot taken by pa11y-ci, verifying a log in.

```
{
    "urls": [
        "http://app:8080/",
        "http://app:8080/health/",
        {
            "url": "http://app:8080/whoami/",
            "screenCapture": "screenshot.png"
        }
    ]
}
```

screenshot

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -35,6 +35,14 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Disable Login
+        working-directory: ./src
+        # by adding MockUserLogin to settings.MIDDLEWARE
+        run: |
+          perl -pi \
+          -e 's/"csp.middleware.CSPMiddleware",/$&"registrar.tests.common.MockUserLogin",/' \
+          src/registrar/config/settings.py
+
       - name: Accessibility Scan
         working-directory: ./src
         # leverage the docker compose setup that we already have for local development
--- a/src/registrar/tests/common.py
+++ b/src/registrar/tests/common.py
@@ -1,6 +1,7 @@
 from django.conf import settings
 from django.contrib.auth import get_user_model, login
 
+
 class MockUserLogin:
     def __init__(self, get_response):
         self.get_response = get_response
--- a/src/registrar/tests/common.py
+++ b/src/registrar/tests/common.py
@@ -1,0 +1,28 @@
+from django.conf import settings
+from django.contrib.auth import get_user_model, login
+
+class MockUserLogin:
+    def __init__(self, get_response):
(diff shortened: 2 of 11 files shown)


src/registrar/tests/common.py
```diff
@@ -0,0 +1,29 @@
...
+    def __call__(self, request):
+        if request.user.is_anonymous:
+            user = None
+            UserModel = get_user_model()
+            username = "Testy"
+            args = {
+                UserModel.USERNAME_FIELD: username,
+            }
+            try:
+                user = UserModel.objects.get(**args)
```
neilmb, line 18: **suggestion (non-blocking):** I think that this `try/except` block could idiomatically be replaced with `UserModel.objects.get_or_create()`?

src/registrar/tests/common.py
```diff
@@ -0,0 +1,29 @@
+from django.conf import settings
```
neilmb, line 1: **question:** Is there a meaningful improvement in security if we keep this file outside of `src/` (which is what gets uploaded to CF) and using both a `cp` and the `perl` invocation to turn it on?

.github/workflows/test.yaml
```diff
@@ -35,6 +35,14 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Disable Login
```
neilmb, line 38: **question:** Is there a recommended way to use this locally to run these same tests? If it is just "add this line to settings.py", could we document that in e.g. a (new) `docs/developer/testing.md`?

neilmb (approved):
LGTM. We should add that `perl` invocation to the OWASP tests whenever they or this PR lands.

src/registrar/tests/common.py
SSPJ: Good point. I've changed it.

.github/workflows/test.yaml
SSPJ: It is just "add this line to settings.py"... and I have updated the docs. :)

PR #165, opened 2022-10-13
Add User Fixtures

SSPJ:
## 🗣 Description ##

In order to make data reproducible in our various environments, this PR starts the process by including a User model fixture and instructions for its use.

Additional considerations are for local users: although we currently do not allow logging via any method except Login.gov, if we were at some future point to decide on the creation of local users (for demo purposes), we would need to update the DJANGO_SECRET_KEY to be the same between all environments. This would be required because Django uses the secret key for password encryption and a password stored in a fixture file will only work with the secret key which created it. This is not an issue for Login.gov users, as they have an empty password field.

### DIFF
--- a/.github/workflows/loaddata.yaml
+++ b/.github/workflows/loaddata.yaml
@@ -1,0 +1,40 @@
+name: Reset database
+
+# This workflow can be run from the CLI
+#     gh workflow run loaddata.yaml -f environment=staging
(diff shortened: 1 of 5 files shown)


loganmeetsworld (approved):
This looks good! I would like to be added to the fixtures as well

PR #197, opened 2022-10-21
#174 Setup form pages for sidebar sections

### LINKED ISSUE #174: Setup pages for sidebar sections
- [x] Set up sidebar sections for registrant form utilizing USWDS
~- [ ] Add mechanism for locking future pages~

neilmb:
# Setup form pages for sidebar sections

## 🗣 Description ##

The new domain application form has a number of separate forms that are displayed on different pages with a navigation sidebar on the left. This sets up that structure using Django's `formtools` form wizard.
 
Screen Shot 2022-10-21 at 11 08 13

It also starts using a Django `widget_tweaks` library for formatting the various parts of forms and form fields with enough flexibility to work with USWDS and (hopefully) our needs for interactivity.

This doesn't include the entire content of the form, just the first two sections, nor does it have any of the interactivity, but we need this first to establish the structure of the form so that other people can add the remaining parts in other PRs.

Closes #174


src/registrar/forms/application_wizard.py
```diff
@@ -0,0 +1,71 @@
...
+    form_list = FORMS
+
+    def get_template_names(self):
+        """Template for the current step.
+
+        The return is a singleton list.
+        """
+        return [TEMPLATES[self.steps.current]]
+
+    def done(self, form_list, **kwargs):
```
loganmeetsworld, line 70: Do we want to add a comment about our future thoughts about `done()` or just leave for now?

loganmeetsworld (approved):
Looks good! Would be curious about if @igorkorenfeld has any questions about the looping and tools used for generating this.

igorkorenfeld:
> Looks good! Would be curious about if @igorkorenfeld has any questions about the looping and tools used for generating this.

I get the general gist though some of the implementation details I have some questions about (mostly being unfamiliar with Django forms). 

One thing to put on your radar, though no action is necessary right now, is that Michelle and I have been talking about the possibility of having sub steps under a larger step in the sidenav. I am currently leaning to only have one level and not going that route, but thought I would mention it.

src/registrar/forms/application_wizard.py
```diff
@@ -0,0 +1,71 @@
...
+
+# Dict to match up the right template with the right step. Keys here must
+# match the first elements of the tuples above
+TEMPLATES = {
+    ORGANIZATION_TITLE: "application_organization.html",
+    CONTACT_TITLE: "application_contact.html",
+}
+
+
+class ApplicationWizard(NamedUrlSessionWizardView):
```
SSPJ, line 67: Docstring.

This might be a good place to briefly explain what a `NamedUrlSessionWizardView` is and how it differs from `forms.Form` (i.e. why we are using it).

src/registrar/config/urls.py
```diff
@@ -18,6 +24,8 @@
     path("health/", health.health),
     path("edit_profile/", profile.edit_profile, name="edit-profile"),
     path("openid/", include("djangooidc.urls")),
+    path("register/", application_wizard, name="application"),
+    re_path(r"^register/(?P<step>.+)/$", application_wizard, name=APPLICATION_URL_NAME),
```
SSPJ, line 28: I think this would work fine as

```
    path("register/<step>/", application_wizard, name=APPLICATION_URL_NAME),
```

src/registrar/forms/application_wizard.py
```diff
@@ -0,0 +1,71 @@
...
+    form_list = FORMS
+
+    def get_template_names(self):
+        """Template for the current step.
+
+        The return is a singleton list.
+        """
+        return [TEMPLATES[self.steps.current]]
```
SSPJ, line 75: **Suggestion:** Consider whether there are cases when `self.steps` may not exist or `self.steps.current` may not have a valid integer or `TEMPLATES` may not include the desired index. If so, is there any error handling we should do or is it okay to fail?

SSPJ (approved):
Great work, Neil! Using the PR checklists, I've identified some areas that could be improved, but overall really nice work! 🎉 Thank you for doing this.

Notes from aXe:
- Header can't be navigation
- Investigate duplicate header elements

Accessibility
- ❌  New pages have been added to .pa11yci file so that they will be tested with our automated accessibility testing
    - N/A All relevant images use an img tag
    - N/A All images have alt attributes
    - N/A Multimedia is tagged. All multimedia has appropriate captioning and audio description
    - N/A Text has sufficient color contrast. All text has a contrast ratio of 4.5:1 with the background
    - [x] Tab order is logical
    - ❌  Form instructions are associated with inputs
    - N/A Site doesn’t timeout unexpectedly. Identify elements that may “timeout” and verify that the user can request more time
    - N/A Tables are coded properly. Tables have proper headers and column attributes
    - N/A Frames are named. All frames have a name element
    - N/A Flashing elements are compliant. Elements that flash on screen do so at a rate of less than 3 Hz
       This is `<html lang="en-us"`
    - N/A The language for sections on the page that differ from the site language are set
       This was tested by deleting the CSS tags from the `<head>`
    - N/A Required plugins are linked on the page
 
Device Matrix
- [x] firefox
    - [x] web pages render correctly
    - [x] interactive functionality is operative
- [x] chrome/chromium/edge
    - [x] web pages render correctly
    - [x] interactive functionality is operative
- ❌  safari
    - I am having issues with `CSRF cookie not set`. This may just be me.
- [ ] web page is readable and usable 
    - ❌  at 480px
    - [x] at 1024px

There's some overlapping text with the sidebar on small mobile devices. It doesn't render the page unusable, merely hard to read.

image

The Basics
- [x] I read the code
- [x] I ran the code
 
Documentation
- ❌  all new functions and methods are commented
    - [ ] using plain language
- ❌  licenses for new dependencies are recorded, if any
- N/A licenses for updated dependencies have been checked for accuracy, if any
- N/A new components have been added to the system diagram
 
Security
- [x] security false positives are documented: None
- N/A data from external sources is cleaned and clearly marked: we have not accepted the form data yet
- N/A significant user actions are logged to a database table
 
Reliability
- [ ] good level of error handling
    - N/A interactions with external systems are wrapped in try/except
- ❌  functionality is tested with unit or integration tests

src/registrar/forms/application_wizard.py
neilmb: `self.steps` is an instance of the Wizard's StepsHelper class that has a `current` property which always returns one of the wizard's steps which are the keys in our `TEMPLATES` dictionary. If that isn't true, then I think that the resulting exception is what we would want to happen.

src/registrar/config/urls.py
neilmb: Great catch. I got this from their documentation, but it appears that the `re_path` version is old-fashioned Django style for the much better `path` version.

neilmb:
@SSPJ I think that this is probably deserving of a re-review at this point, maybe?

SSPJ (approved):
@neilmb I've taken another pass; thanks for making these fixes! 💯 

When I was testing with Firefox, I noted that the text input labels were not being correctly associated with the fields. I think this might be due to the generated IDs violating the characters permitted by the W3 spec ([see this SO answer](https://stackoverflow.com/a/79022)). If there's a quick fix, that would be great; if not, we'll tackle it later.

PR #212, opened 2022-10-28
Show and Hide Questions Dynamically

### LINKED ISSUE #175: Develop infrastructure for dependent answers
- [ ] Develop infrastructure for dependent answers.
- [ ] y/n to question and something else pops up - if any particular page ended up having more than two levels of dependencies - break out

SSPJ:
## 🗣 Description ##

Includes Django and JavaScript code to allow one set of radio buttons to trigger a question containing another set of radio buttons.

The overall design is that radio buttons are rendered by Django code in templates and then (using HTML attributes to signal which belongs to which) a JavaScript script runs to attach event handlers that do the actual work of show/hide.

To test, go to http://localhost:8080/register and select radio buttons in different combinations. Questions should appear as shown in the wireframe.

Here are some of the design considerations that came into play with this PR:

What should AT users hear when a question appears or vanishes?

A: For the moment, they hear "Previous follow-up question removed" and "New follow-up question required".

What happens if a hidden question is already answered when the page loads (for example, restoring a saved application or going backward from a later step)?

A: The hidden question is shown.

What happens if a hidden question is not answered, but the triggering question is already answered with the correct answer to trigger the follow up question when the page loads?

A: The hidden question is shown.

What happens to the answer if a hidden question is answered but the triggering question's answer changes?

A: The answer is ~left intact and will reappear~ cleared. (This may not be the desired behavior.)

What happens to the answer if a hidden question is answered but the triggering question is no longer answered with the correct answer to trigger that follow up when the form is submitted?

A: The answer is ~saved and sent to the server~ discarded. (This may not be the desired behavior.)

Are hidden questions required?

A: ~Yes.~ Depends. Although, this causes a problem in that the "next" button prevents submitting the form if required questions are left unanswered. The solution was to ~remember which questions are required~ make hidden questions required only if the triggering question is also required and remove/restore the `required` attribute when the questions are hidden/shown.

**Update:** Responding to PR feedback and bug fixes caused me to rethink some implementation details. Edits are shown above.

Closes #175

### DIFF
--- a/src/registrar/assets/js/get-gov.js
+++ b/src/registrar/assets/js/get-gov.js
@@ -1,0 +1,169 @@
+/** Strings announced to assistive technology users. */
+var ARIA = {
+  QUESTION_REMOVED: "Previous follow-up question removed",
+  QUESTION_ADDED: "New follow-up question required"
+}
+
+/** Helper function. Makes an element invisible. */
+function makeHidden(el) {
+  el.style.position = "absolute";
+  el.style.left = "-100vw";
+  // The choice of `visiblity: hidden`
+  // over `display: none` is due to
+  // UX: the former will allow CSS
+  // transitions when the elements appear.
+  el.style.visibility = "hidden";
+}
+
+/** Helper function. Makes visible a perviously hidden element. */
+function makeVisible(el) {
+  el.style.position = "relative";
+  el.style.left = "unset";
+  el.style.visibility = "visible";
+}
+
+/** Inputs which are required (if visible). */
+var requiredInputs = {}
+
+/** Helper function. Removes `required` attribute. */
+function removeRequired(el) {
+  const inputs = el.querySelectorAll("input");
+  for (const input of inputs) {
+    if (input.required) {
+      requiredInputs[input.id] = true;
+      input.removeAttribute("required");
+    }
+  }
+}
+
+/** Helper function. Adds `required` attribute. */
+function markRequired(el) {
+  const inputs = el.querySelectorAll("input");
+  for (const input of inputs) {
+    if (input.id in requiredInputs) {
+      input.setAttribute("required", "");
+    }
+  }
+}
+
+/** Helper function. Creates and returns a live region element. */
+function createLiveRegion(id) {
+  const liveRegion = document.createElement("div");
+  liveRegion.setAttribute("role", "region");
+  liveRegion.setAttribute("aria-live", "polite");
+  liveRegion.setAttribute("id", id + "-live-region");
+  liveRegion.classList.add("sr-only");
+  document.body.appendChild(liveRegion);
+  return liveRegion;
+}
+
+/** Currently selected radio buttons. */
+var selected = {};
+
+/** Mapping of radio buttons to the toggleables they control. */
+var radioToggles = {};
+
+/**
+ * Helper function. Tracks state of selected radio button.
+ * 
+ * This is required due to JavaScript not having a native
+ * event trigger for "deselect" on radio buttons. Tracking
+ * which button has been deselected (and hiding the associated
+ * toggleable) is a manual task.
+ * */
+function setSelected(radioButton) {
+  selected[radioButton.name] = radioButton;
+}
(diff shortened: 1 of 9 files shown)


neilmb:
The OWASP Zap test failure is a false positive due to the user of "suspicious comments" in `get-gov.js`. You can add an exception in `src/zap.conf` similar to this one: https://github.com/cisagov/getgov/blob/main/src/zap.conf#L28-L29

neilmb:
## Accessibility
- N/A New pages have been added to .pa11yci file so that they will be tested with our automated accessibility testing
    - N/A All relevant images use an img tag
    - N/A All images have alt attributes
    - N/A Multimedia is tagged. All multimedia has appropriate captioning and audio description
    - [X] Tab order is logical
    - N/A Site doesn’t timeout unexpectedly. Identify elements that may “timeout” and verify that the user can request more time
    - N/A Tables are coded properly. Tables have proper headers and column attributes
    - N/A Frames are named. All frames have a name element
    - N/A Flashing elements are compliant. Elements that flash on screen do so at a rate of less than 3 Hz
    - N/A The language for sections on the page that differ from the site language are set
    - N/A Required plugins are linked on the page
 
## Device Matrix
- [X] firefox
    - ❌ interactive functionality is operative
- [X] chrome/chromium/edge
    - ❌ interactive functionality is operative
- ❌ safari - couldn't get Login to work with Safari
    - [ ] interactive functionality is operative
    - [X] at 480px
    - [X] at 1024px

neilmb:
I found one situation where the hiding and revealing didn't work right.

Screen Shot 2022-10-28 at 12 12 24

To reproduce, select "Federal" and then "Executive" in the revealed question. Click "Next" to go to the next form. On the next form, click "About your organization" in the sidebar to go back to the first form. Then all of the sub-questions are shown at the same time. (To make them go away, click between "Federal", "Interstate", and "Tribal" a couple of times and various revealed questions hide themselves.)

src/registrar/assets/js/get-gov.js
```diff
@@ -0,0 +1,169 @@
...
+
+  for(const toggleable of toggleables) {
+    // get the (comma-seperated) list of radio button ids
+    // which trigger this toggleable to become visible
+    const attribute = toggleable.getAttribute("toggle-by") || "";
+    if (!attribute.length) continue;
+    const radioIDs = attribute.split(",");
+
+    createLiveRegion(toggleable.id)
+    console.log('toggleable :', toggleable);
```
neilmb, line 154: **suggestion**: I think that we shouldn't keep debugging console statements in our Javascript when we get to production.

neilmb (changes requested):
This is so great to see, very impressive combined front- and backend work.

I found one situation where the hiding and revealing didn't work as I expected. I'm not sure why it is a problem to re-load the first page of the form, but I would guess it has something to do with how formtools renders the form on a re-load.

src/registrar/assets/js/get-gov.js
SSPJ: Good catch, thank you!

SSPJ:
@neilmb I think this is ready for another pass when you have a moment.

src/registrar/templates/application_organization.html
```diff
@@ -1,6 +1,8 @@
 <!-- Test page -->
 {% extends 'application_form.html' %}
 {% load widget_tweaks %}
+{% load debugger_tags %}
```
neilmb, line 4: **question (non-blocking)**: Is there anything in these debugger tags that would make us not even want to include the code into this template in production? I think there is nothing bad that can happen if we don't use the tags, but I thought I would ask.

neilmb (approved):
Works a treat and looks a bit simplified inside the JS code. 5️⃣ 🌟 ✔️

src/registrar/templates/application_organization.html
SSPJ: Good catch here as well. I don't think there is any harm, but there's also no point in loading more tags than we need.

PR #215, opened 2022-10-28
Use a demo site banner to avoid confusion

### LINKED ISSUE #152: Add a DEMO SITE banner to unstable and staging
As we develop on public cloud.gov sites, the USWDS theme encourages people to believe that this is an official government site. But of course it is not. We should  add a LARGE RED BANNER with text explaining that these are only demo sites. And configure that banner to appear in the `unstable` and `staging` environments, but it won't appear on a production environment.

neilmb:
# Add a demo site banner so we don't claim to be a real government site #

## 🗣 Description ##

As we add more functionality to our development sites running on cloud.gov, we need an indication that this is not the actual site for registering dotgov domains. This adds a configuration setting `IS_DEMO_SITE` and a template variable to control whether or not we include a big red banner that says "TEST SITE".

Screen Shot 2022-10-28 at 14 37 28

Currently we only have an unstable and staging environment so this config variable is always `True`. When we get close to having a production site, we will have to put in a place a method for modifying this setting to make it `False` on the production app.

This closes #152.

### DIFF
--- a/src/registrar/templates/base.html
+++ b/src/registrar/templates/base.html
@@ -106,7 +106,7 @@
   {% if IS_DEMO_SITE %}
     <section
     class="usa-site-alert usa-site-alert--emergency usa-site-alert--no-icon"
-    aria-label="Site alert,,,,,"
+    aria-label="Site alert"
     >
       <div class="usa-alert">
         <div class="usa-alert__body">
--- a/src/registrar/templates/base.html
+++ b/src/registrar/templates/base.html
@@ -103,6 +103,20 @@
       </div>
     </div>
   </section>
+  {% if IS_DEMO_SITE %}
+    <section
+    class="usa-site-alert usa-site-alert--emergency usa-site-alert--no-icon"
+    aria-label="Site alert,,,,,"
+    >
+      <div class="usa-alert">
+        <div class="usa-alert__body">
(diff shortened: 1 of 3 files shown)


src/registrar/templates/base.html
```diff
@@ -103,6 +103,20 @@
       </div>
     </div>
   </section>
+  {% if IS_DEMO_SITE %}
+    <section
+    class="usa-site-alert usa-site-alert--emergency usa-site-alert--no-icon"
+    aria-label="Site alert,,,,,"
```
SSPJ, line 109: Lots and lots of `,`'s?

SSPJ (approved):
Nice work, Neil! I've done an abbreviated review, since we don't expect this banner to be shown to users in the finished site, but everything looks great. 👌 

The Basics
- [x] I read the code
- [x] I ran the code
 
Documentation

Accessibility
- [x] banner content is audible to AT users (tested via VO on Chrome)

src/registrar/templates/base.html
neilmb: Copied and pasted from https://designsystem.digital.gov/components/site-alert/ but now I see they were using the commas to differentiate between many different components with the same label. Fixed.

PR #233, opened 2022-11-03
Add a nightly cron to our deploy action

loganmeetsworld:
To ensure any updates to the buildpack get incorporated on a regular basis, we rebuild and deploy nightly our application to staging. 

I asked the cloud.gov team about the difference between this and a custom nightly restage and they said: 

> cf deploy action is just a cf push . Both push and restage will recompile the application basically. Restage will recompile with the code that is existing and Push will compile with any changes to code that is pushed.


neilmb (approved):
Amazingly simple. Could we add a note to our developer documentation that this happens and a link to the Github Actions "dashboard" where someone could look to make sure that it was happening correctly every night?

PR #236, opened 2022-11-04
Add JavaScript to check domain availability

### LINKED ISSUE #223: Make use of "/available/" API endpoint in domain registrant form
_Outcome_: Users can see if a domain is available when filling out the domain request form.

We need to hook up the on_change property to javascript that XHR calls the API ("/available/") endpoint and puts feedback about availability on the page.

This builds ontop of #213 (issue #176)

SSPJ:
## 🗣 Description ##

When a user types characters into the domain validation box, this code waits until they appear to be done typing and then calls the API to determine availability.

For testing/demo purposes, I have transformed the existing "Street address" field into a domain validator.

image

This PR does not include any styling elements or user messaging, aside from the bare minimum to make the code functional.

This is one possible solution to doing the frontend domain validation check. Another way, likely simpler and with better user experience is to place an explicit button to allow checking.

Nevertheless, this implementation includes some JavaScript code (and patterns) which may be useful for validation elsewhere in the application.

It may, at some point, be worthwhile to write tests for our JavaScript. That will be no simple undertaking however, because they can only be run e2e (in other words, with Django running, similar to how pa11y works). In the meantime, I have been concentrating effort on ensuring that the pages are _fully functional_ even if the user has JavaScript disabled or the scripts crash/fail to load.

Closes #223

### DIFF
--- a/src/registrar/assets/js/get-gov.js
+++ b/src/registrar/assets/js/get-gov.js
@@ -1,13 +1,26 @@
+/**
+ * @file get-gov.js includes custom code for the .gov registrar.
+ *
+ * Constants and helper functions are at the top.
+ * Event handlers are in the middle.
+ * Initialization (run-on-load) stuff goes at the bottom.
+ */
+
 /** Strings announced to assistive technology users. */
 var ARIA = {
   QUESTION_REMOVED: "Previous follow-up question removed",
   QUESTION_ADDED: "New follow-up question required"
 }
 
+var DEFAULT_ERROR = "Please check this field for errors.";
+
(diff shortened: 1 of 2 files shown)


neilmb (approved):
I tested this and it does what it says on the tin using the validator API. 

When I was using it, I wished there was more feedback on the page as to why what I entered was invalid. I can see where we could change _checkDomainAvailability to put text into a DOM element that says "The domain <fda.gov> is already taken". But, until we have an actual field to apply it to and some design for how we want the interaction to go, I'm fine with this as-is.

PR #247, opened 2022-11-09
Create a new DomainApplication in database on form submission

### LINKED ISSUE #222: Save a new domain application on form submission

neilmb:
# Create a new DomainApplication on form submission #

## 🗣 Description ##

When a registrant completes our form, we need to record all of their submissions and make an entry in our database. This adds a `done` method to our registration form that takes all of the form answers, puts them into the data model for the domain application and `save`s that data in the database.

It works with the current two-page form that we initially developed to figure out the infrastructure. As we add additional form pages, we will need to store the new data from those forms in the `forms_to_object()` method on the `ApplicationWizard` class. Also, it currently redirects to the home page without any feedback that the submission was correct. We still need to design and implement what we want to happen after a successful form submission.

Closes #222


loganmeetsworld:
If we want to add the concept of saving progress prior to submitting will we repopulate the form with answers previously given and then use the same form submit? Is it then possible that a DomainApplication exists when opening the form that we use to repopulate? I don't think this is a requirement for this ticket but just thinking about how that will factor in.

src/registrar/forms/application_wizard.py
```diff
@@ -113,5 +116,34 @@ def get_context_data(self, form, **kwargs):
...
+        requested_site, _ = Website.objects.get_or_create(
+            website=contact_data["organization_name"] + ".gov"
+        )
+        application.requested_domain = requested_site
+        return application
+
+    def done(self, form_list, form_dict, **kwargs):
         logger.info("Application form submitted.")
+        application = self.forms_to_object(form_dict)
+        application.submit()  # change the status to submitted
```
SSPJ, line 146: **Suggestion:** Seeing this function used now in its natural state and how the naming of states could end up being quite confusing, I wonder if we'd do better with a setter method to wrap the actual transition methods. Not a real Python setter (unless you want), just something like `application.set_status(to=DomainApplication.SUBMITTED)`.

src/registrar/forms/application_wizard.py
```diff
@@ -113,5 +116,34 @@ def get_context_data(self, form, **kwargs):
...
+        contact_data = form_dict["contact"].cleaned_data
+        application.organization_name = contact_data["organization_name"]
+        application.street_address = contact_data["street_address"]
+        # TODO: add the rest of these fields when they are created in the forms
+
+        # This isn't really the requested_domain field
+        # but we need something in this field to make the form submittable
+        requested_site, _ = Website.objects.get_or_create(
+            website=contact_data["organization_name"] + ".gov"
+        )
```
SSPJ, line 139: Your choice on whether you'd like to take care of it now or make a follow up ticket, but any data validation which is performed on the client side as a connivence to the user must be re-performed during saving. In other words, the user should get back a nice, readable message, not a random 500 error.

The steps to reproduce are 1. put "my domain name" in the organization name field; 2. click submit.

src/registrar/forms/application_wizard.py
```diff
@@ -113,5 +116,34 @@ def get_context_data(self, form, **kwargs):
...
+        application.requested_domain = requested_site
+        return application
+
+    def done(self, form_list, form_dict, **kwargs):
```
SSPJ, line 143: Docstring.

SSPJ (approved):
LGTM 👍 

I left a few discretionary comments, but feel free to merge as-is.

The Basics
- [x] I read the code
- [x] I ran the code
 
Documentation
    - [x] using plain language
 
Security
 
Reliability
- ❌ good level of error handling
    As mentioned, okay to defer this work to a follow up ticket.

SSPJ:
@loganmeetsworld I'll share my thoughts and let @neilmb respond.

> If we want to add the concept of saving progress prior to submitting will we repopulate the form with answers previously given and then use the same form submit?

At the moment, I believe answers are stored in the user's session (Neil can correct me.) -- but yes, we'd want to make a new row in the DomainApplication table as soon as a user starts and we'd want to renew their progress whenever they return.

> Is it then possible that a DomainApplication exists when opening the form that we use to repopulate? I don't think this is a requirement for this ticket but just thinking about how that will factor in.

Yup, loading the form from the database would normally be handled by the view code. This library does its own routing and views, so I'm not sure exactly what needs done.

Issue #14, opened 2022-08-04
Background dev research for admin experience


SSPJ:
- [x] Meet with cloud.gov to ask questions about networking for two applications
- [x] ~Draft a diagram of how the backend application might be deployed as two applications, split between the CISA admin and registrant experiences~ N/A, going to develop integrated application for the moment
- [x] Investigate and document options for role-based access control (RBAC) from the CISA admin perspective
    Also see #137 
- [x] ~Investigate and document how both applications might talk to the database~
- [x] Investigate and document how Django Admin might be leveraged in an MVP
- [x] From #277. There is seemingly a bug with FSMField and choices: admin says "Select a valid choice. False is not one of the available choices."
- [x] From #277. Admin does not work with FSMFields if `protected=True`.

~~Do not draft an ADR. There are currently too many unknowns to commit to a decision. The outcome of this ticket will be further tickets (which may include ADRs).~~

See also [this note](https://github.com/cisagov/getgov/pull/248#discussion_r1019578536) on #248 and make sure that testing is part of the A/C of any follow tickets.

loganmeetsworld:
Some questions I've been thinking about:
* Do we have a split between superadmins and admins? 
* How customized does the view for the admin interface need to be?
* Can you put a Django app and Django admin on two different subdomains?

loganmeetsworld:
Next steps for this is a meeting with cloud.gov to discuss our options here and get some questions answered:

* We have one set of routes and interface that might need to go behind a VPN and another that is public. Both need to communicate with the same DB. How should we go about this?
* Admin access vs. registrant access
* What is currently the best option for email notifications?
* What is currently the best option for log storage? / searching?
* We might need a task queue, would CF tasks be sufficient for our use cases or do we need to explore another option?

loganmeetsworld:
Something that might be useful to come out of this ticket as well is an evaluation of what we would need from a CRM, ticketing software, or communications software that would need to be integrated into the analyst/admin side.

SSPJ:
Met with Igor on Friday to review existing features of the Django admin. He suggested a list of items to research for implementation:
- [ ] location of search box / search capabilities
- [ ] ability to run reports
- [ ] ability to export data
- [ ] sorting ability
- [ ] (less important) ability to theme

Next steps are to find potential implementations to those features and make a prototype for Igor to review. The goal is to understand the trade-off in effort to create a new admin interface vs. customizing the existing admin.

[...]

PR #248, opened 2022-11-10
Include audit logging

### LINKED ISSUE #218: Research django-audit and how audits of models like Application work
_Outcome_: Select a process for auditing events in our Django application, investigate django-audit 

- [ ] Write up ADR

# Background

It is a good idea to have traditional domain models like “Application” and some way to tie LoginUser to Application. Whereas the registry will be complete source of truth for Domains, we will have to track and keep the truth for Applications.

There are two areas event sourcing or even tracking could bring benefit - Application (just a series of events) and changes to users permissions/roles/domains/etc. We believe this will be best modeled in the db with traditional modeling for it and decouple the audit logs - a separate table that is an event source but it not the model.

The benefits to this approach: 
- For analysts working on an application, see a history of how and when and who has changed the state of that application 
- The detail page for a domain application would have a table for past changes - SELECT FROM events WHERE app.id == app.id

SSPJ:
## 🗣 Description ##

The purpose of this PR is to automatically log any model changes to an audit table in the database.

It also includes a refactoring of models.py. To make this work easier to review, changes to the models are in one commit and then moving the models to their new file locations are in a subsequent commit. (Except for a couple of tiny bug fixes while I was getting the tests to run.)

@neilmb I'm tagging you as primary reviewer, but @igorkorenfeld this also touches on your work and I'd welcome any feedback.

As implemented, this logs _all_ changes, which includes keeping an audit log of the `last_login` field of the User table. This may not be desired.

Closes #218

## How to test ##

This requires a full dev reset. The suggested way to do this:

```
docker-compose down
docker-compose build --no-cache
docker-compose up
(in other window) docker-compose exec app ./manage.py loaddata registrar/fixtures/*
```

Go to http://localhost:8080/edit_profile and edit the display name. Go to http://localhost:8080/admin/auditlog/logentry/ and verify that the change was recorded.

Go to http://localhost:8080/admin/registrar/domainapplication/add/ and add a domain application. Go to the detail view for the newly added application and click "History". Verify that you see the changes you've made.

image

image


src/registrar/models/domain_application.py
```diff
@@ -0,0 +1,227 @@
...
+        blank=True,
+        help_text="Acknowledged .gov acceptable use policy",
+    )
+
+    def __str__(self):
+        try:
+            if self.requested_domain and self.requested_domain.website:
+                return self.requested_domain.website
+            else:
+                return f"{self.status} application created by {self.creator}"
```
neilmb, line 210: This is a lovely default for when the site is not set.

src/registrar/admin.py
```diff
@@ -18,3 +37,5 @@ class MyUserAdmin(UserAdmin):
 
 
 admin.site.register(User, MyUserAdmin)
+admin.site.register(DomainApplication, AuditedAdmin)
```
neilmb, line 40: **question (non-blocking):** Do we need/want tests on our custom admin views? Is it easy enough to make a `TestCase` that makes an admin user, uses `force_login`, and loads one or more of these admin pages?

neilmb (approved):
LGTM, great work on the `models` module and a truly non-blocking question about testing admin views.

src/registrar/admin.py
SSPJ: I think that's a great idea, but I'm just going to make a note on #14 about it for the moment.

PR #277, opened 2022-11-17
Add Domain model

### LINKED ISSUE #252: Create a basic domain model
_Outcome_: A new domain model that would be created after ApplicationDomain moves to `Accepted`. 

A/C: We can keep this model simple but it should contain the string of the domain, a status, and be attachable to a user as the domain owner. 

We would like to document the Domain model state changes somewhere and have a model file added.

SSPJ:
## 🗣 Description ##

This PR adds a Domain model. It is heavy on design but light on implementation, as there are many details which are still tbd on how the registry will integrate with the registrar.

The overall idea is that other parts of the application (views, forms, etc) will be able to transparently use the Domain model to create, update, renew, etc. Meanwhile, the Domain model will do the heavy lifting behind the curtain to determine whether to contact the registry (for live/active domains) or to store changes locally (for draft/unapproved domains).

There are still many things which need to be done, but not on this PR.

* Wrap the EPP calls in audit logging.
* Flesh out how hostnames, ip address, nameservers, and domain contacts are handled
* Determine how/when/why the `owners` field on Domain gets populated
* Add code to deal with the asynchronous nature of DNS updates
    * (e.g. what to do if the registry does not complete the update within a reasonable timeout period)
* There is seemingly a bug with FSMField and choices: admin says "Select a valid choice. False is not one of the available choices."
* Admin does not work with FSMFields if `protected=True`.

I will capture these in existing or new tickets following PR review.

Closes #252


src/registrar/models/domain_application.py
```diff
@@ -147,12 +147,12 @@ class DomainApplication(TimeStampedModel):
         related_name="current+",
     )
 
-    requested_domain = models.ForeignKey(
-        Website,
+    requested_domain = models.OneToOneField(
```
neilmb, line 150: **question (non-blocking)**: I wonder if a `OneToOneField` might not be right here. @h-m-f-t has mentioned before that we should be ready for multiple applications for the same domain name to be opened, then one of them could be awarded. I think that this could still work here since the `name` field of `Domain` is not constrained to be `unique`, but I'm also not sure if that is how we want it or not.

**suggestion (non-blocking)**: Could we maybe have a test case of the models to explicitly show what we want to be possible? Create two domain applications for the same domain at the same time? I know that we do it in one of the tests, but since that test is about something else, it isn't as clear to me that we are defining in the tests how we specifically want it to be possible.

src/registrar/models/host_ip.py
```diff
@@ -0,0 +1,31 @@
...
+        null=False,
+        blank=False,
+        default=None,  # prevent saving without a value
+        validators=[validate_ipv46_address],
+        help_text="IP address",
+    )
+
+    host = models.ForeignKey(
+        Host,
+        on_delete=models.PROTECT,
```
neilmb, line 29: **question (non-blocking)**: This seems maximally safe to force us to manually clean up a host's IP addresses before we delete the host, but I wonder if it might be reasonable to just `models.CASCADE` here and auto-delete these IP addresses when we delete the host?

src/registrar/models/host.py
```diff
@@ -0,0 +1,32 @@
...
+        default=None,  # prevent saving without a value
+        unique=True,
+        help_text="Fully qualified domain name",
+    )
+
+    domain = models.ForeignKey(
+        Domain,
+        on_delete=models.PROTECT,
+        help_text="Domain to which this host belongs",
+    )
```
neilmb, line 32: **suggestion (non-blocking)**: For models like this that are foreign keys for other things like `HostIP`, they get useful reverse lookup properties that might be used quite frequently (i.e.  Host.hostip_set). Could we have a comment here in this model about the existence of that reverse relation and its name to improve locality-of-reference for future developers?

src/registrar/models/nameserver.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class Nameserver(Host):
+    """
+    A nameserver is a host which has been delegated to respond to DNS queries.
+
+    The registry is the source of truth for this data.
+
+    This model exists ONLY to allow a new registrant to draft DNS entries
+    before their application is approved.
```
neilmb, line 11: **suggestion (non-blocking)**: It isn't always obvious to me in Django models that use class inheritance that I should look at the super class to learn about the fields that are allowed here. Could we have a comment to that effect, maybe above  the `pass` that says "there is nothing here because all of the fields are defined over there on the Host class"?

src/registrar/tests/common.py
```diff
@@ -1,7 +1,57 @@
+import os
+import logging
+
+from contextlib import contextmanager
```
neilmb, line 4: **question**: It seems very unfortunate to have to repeat this exact code across the `api` and `registrar` Django apps. Is it possible to define it in one of the apps and find an appropriate `import` statement to bring it into the other one? If not, can we put it somewhere else and use it from there in both apps?

src/registrar/tests/test_models.py
```diff
@@ -56,9 +57,49 @@ def test_status_fsm_submit_fail(self):
...
+        """Returns proper status based on `is_active`."""
+        domain = Domain.objects.create(name="igorville.gov")
+        domain.save()
+        self.assertEquals(None, domain.status)
+        domain.activate()
+        domain.save()
+        self.assertIn("ok", domain.status)
+
+    def test_fsm_activate_fail_unique(self):
+        # can't activate domain if name is not unique
```
neilmb, line 89: **suggestion (non-blocking)**: Can we make these initial comments into docstrings so that the test runner can display them when appropriate?

src/registrar/tests/test_models.py
```diff
@@ -56,9 +57,49 @@ def test_status_fsm_submit_fail(self):
...
+        d1, _ = Domain.objects.get_or_create(name="igorville.gov")
+        d2, _ = Domain.objects.get_or_create(name="igorville.gov")
+        d1.activate()
+        d1.save()
+        with self.assertRaises(ValueError):
+            d2.activate()
+
+    def test_fsm_activate_fail_unapproved(self):
+        # can't activate domain if application isn't approved
```
neilmb, line 98: suggestion (non-blocking): As above, docstring instead of comment?

SSPJ: Yep.

src/registrar/tests/test_models.py
SSPJ: I had no idea you could do that! Done and thank you.

src/registrar/tests/common.py
SSPJ: I considered that because I am also not a fan of the code duplication, but Django apps are intended to be plug-and-play with each other. So I think avoiding non-functional dependencies between them is the lesser of two evils in this case.

src/registrar/models/nameserver.py
SSPJ: Done.

src/registrar/models/host.py
SSPJ: Done.

src/registrar/models/host_ip.py
SSPJ: I'll leave it as `models.PROTECT` for the moment, because I believe this mirrors what EPP protocol specifies for the registry. (IP addys must be deleted before their hosts can be.) However, since this model is only a scratchpad for draft data that has not been published to the registry yet, I think we could absolutely change it later, if it ends up being more useful as `models.CASCADE`.

src/registrar/models/domain_application.py
SSPJ: 1. I see what you are saying. I think `OneToOneField` is the right choice for now because we've agreed that an applicant can only apply for one domain at a time (one domain per application) and, while a domain _string_ may not be unique during the application phase, the Domain model and its associated Foreign Key models track additional information, like nameserver data, which is uniquely tied to one organization (one application per domain).

1. Done-ish. I've sketched out some test cases for the interaction between the DomainApplication model and the Domain model, but have made use of `@skip` on them, since there's no working implementation for it.

src/registrar/models/domain.py
```diff
@@ -0,0 +1,223 @@
...
+                # get info from registry
+                self.info = domain_info(self.name)
+            except Exception as e:
+                logger.error(e)
+                # TODO: back off error handling
+                return None
+        if hasattr(self, "info") and (property in self.info):
+            return self.info[property]
+        else:
+            return None
```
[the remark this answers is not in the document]
SSPJ: Done. I implemented `KeyError` for when `self.info` exists but `property` doesn't, however this function still needs additional error handling for network failures. I've added more explicit TODOs for that.

src/registrar/models/domain.py
```diff
@@ -0,0 +1,223 @@
...
+            # once approved, they must be globally unique
+            models.UniqueConstraint(
+                fields=["name"],
+                condition=models.Q(is_active=True),
+                name="unique_domain_name_in_registry",
+            ),
+        ]
+
+    class Status(models.TextChoices):
+        """The status codes we can receive from the registry."""
```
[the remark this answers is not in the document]
SSPJ: Can't add an exact link because they are buried deep in a long document, but I've added a reference and a link to the document. (Could probably find a linkable version of that RFC, but it makes more sense I think to have the whole entire thing in context, if someone were to go hunting for more information.)

src/api/tests/test_available.py
```diff
@@ -104,10 +105,12 @@ def test_available_get(self):
 
     def test_available_post(self):
         """Cannot post to the /available/ API endpoint."""
-        response = self.client.post(API_BASE_PATH + "nonsense")
+        with less_console_noise():
```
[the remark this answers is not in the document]
SSPJ: Done.

PR #278, opened 2022-11-17
List users on homepage

### LINKED ISSUE #261: DomainApplication list on logged in user homepage
Outcome: A list (limited to one now) of domain applications on the homepage of a logged in user. 

A/C: As a user I want to see my in progress applications for a domain. As a user I want to be able to select an application that is `In Progress` and continue my progress. As a user I want to be able to view submitted applications.

SSPJ:

<strike>This PR conflicts with #277 in the use of `Website` vs `Domain` for the `requested_domain` field. This must be updated before merging.</strike> Will address on #277 

Closes #261 

## 🧪  Testing steps ##

1. Use the [admin](http://localhost:8080/admin/) to add a few domain applications, some with your user as a the creator and some with another user (`docker-compose exec app ./manage.py loaddata registrar/fixtures/*` if you have no other users).
2. Verify that all your applications show up and no others
3. Use the frontend testing checklist to verify no accessibility issues


src/registrar/templates/home.html
```diff
@@ -17,9 +17,37 @@ <h2>
...
+    {% endfor %}
+  </tbody>
+</table>
+{% endif %}
+
+<p><a href="{% url 'application' %}" class="usa-button">Apply</a></p>
+
+<p><a href="{% url 'edit-profile' %}">Edit profile</a></p>
+
+<p><a href="{% url 'admin:index' %}">CISA admin panel</a></p>
```
neilmb, line 46: **suggestion**: I don't think that we want all logged-in users to see this, can we wrap it in a conditional on whether the user `is_staff`?

neilmb (changes requested):
I like how straightforwardly this works to put this content on the home page.

I would like to see a new test case that exercises this view. It could create a DomainApplication object, save it, then visit the page as the logged-in user and check that there is a table with one row and the correct name.

neilmb:
## Accessibility
- ~New pages have been added to .pa11yci file so that they will be tested with our automated accessibility testing~
    - ~All form inputs have explicit labels~
    - ~All relevant images use an img tag~
    - ~All images have alt attributes~
    - ~Multimedia is tagged. All multimedia has appropriate captioning and audio description~
    - [X] Tab order is logical
    - ~Form instructions are associated with inputs~
    - ~Site doesn’t timeout unexpectedly. Identify elements that may “timeout” and verify that the user can request more time~
    - ~Tables are coded properly. Tables have proper headers and column attributes~
    - ~Frames are named. All frames have a name element~
    - ~Flashing elements are compliant. Elements that flash on screen do so at a rate of less than 3 Hz~
    - ~The language for sections on the page that differ from the site language are set~
    - ~Links are unique and contextual. All links can be understood taken alone, e.g., ‘Read more - about 508’~
    - ~Required plugins are linked on the page~
 
## Device Matrix
- [X] firefox
    -  ❌ interactive functionality is operative: tab key doesn't cycle through page elements!?
- [X] chrome/chromium/edge
- ❌ safari: I can't log in successfully in Safari. There is an error in djangooidc.oidc "Received state not the same as expected for"
    - [X] at 480px
    - [X] at 1024px

src/registrar/templates/home.html
```diff
@@ -17,9 +17,37 @@ <h2>
...
+  <thead>
+    <tr>
+      <th>Name</th>
+      <th>Status</th>
+    </tr>
+  </thead>
+  <tbody>
+    {% for application in domain_applications %}
+      <tr>
+        <th>{{ application.requested_domain.website }}</th>
```
neilmb, line 34: **question (non-blocking)**: I invented the URL `/application/<id>` for the detail view of each domain application. Do we want to make this a link to that URL?

SSPJ: I'm a bit uncertain about this change. I'm not opposed, but I can't find `/application/<id>` in `urls.py`, so I can't add a link to it yet. Otherwise `reverse` will complain `NoReverseMatch`.

src/registrar/templates/home.html
SSPJ: Done.

neilmb (approved):
Test case looks good, thanks.

PR #302, opened 2022-12-07
Use email instead of username UUID

neilmb:
# Use email instead of ugly UUID username #

## 🗣 Description ##

Our Login.gov integration puts the Login.gov UUID into the `username` of the Django user. That means that it shows up in the top-right corner of our navigation bar. Which is (IMO) ugly. 

A recent conversation emphasized the primacy of email address for identifying logged-in users, so this changes the page template to display the logged-in email address in the top-right.

Screen Shot 2022-12-07 at 10 23 01

Just scratching an itch to have the page look more pleasing with less gobbledygook.

### DIFF
--- a/src/registrar/templates/base.html
+++ b/src/registrar/templates/base.html
@@ -145,7 +145,8 @@
         <ul class="usa-nav__primary usa-accordion">
           <li class="usa-nav__primary-item">
             {% if user.is_authenticated %}
-            <a href="/whoami"><span>{{ user.email }}</span></a>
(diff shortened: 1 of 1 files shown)


SSPJ (approved):
LGTM 👍 

I have not tested this, since the change is very small and the code appears to be correct.

igorkorenfeld:
This is looking great. Can be a follow up, but if it's not a lot of effort would you be able to add a `  |  Sign out ` button after the name?

Screenshot (you can ignore the search box)

image

PR #323, opened 2022-12-12
Review application

### LINKED ISSUE #263: Implement "Review application" page with real data
_Outcome_: The "review application" page shows actual data inputs prior to submission. 

A/C: As a user I want to be able to review all the data in my application prior to submitting. 

Builds on #257

SSPJ:
## 🗣 Description ##

Edit the review page to see the information the user entered in previous steps.

image

Due to the way the current application auto-saves progress to the database each time the user clicks "next", it was possible to implement this page in the style of an old-time telephone conversation: first the information is written down; then it is read back.

This provides a slightly stronger guarantee that what the user intended to submit has successfully made it past all validation checks and will be identical to the information seen by the analyst reviewing the case.

There are two areas of this page which could use a bit of workshopping for accessibility. CC @igorkorenfeld and @boonerang 

1. The page is sub-divided by pseudo-headings (such as "Purpose of your domain") which are apparent to sighted users, but are not semantically present for screenreader users. (Users will hear "...Organization name and mailing address City of Anyville...".) Impact is likely minor.
2. The page has several visually identical "Edit" buttons. This poses a problem both for screenreader users, who will tab through the page hearing "Edit, Edit, Edit, Edit, Edit. . .", and also for voice control users, who cannot easily target a link by saying "click on Edit"; I've included the section title as hidden text in the edit link, which will fix this for screenreader users, but I'm uncertain if this is sufficient for voice control users.

Closes #263


src/registrar/forms/application_wizard.py
```diff
@@ -79,17 +79,19 @@ class OrganizationContactForm(RegistrarForm):
...
+        label="Street address line 2",
     )
+    city = forms.CharField(label="City")
     state_territory = forms.ChoiceField(
-        label="State/territory",
+        label="State, territory, or military post",
         choices=[("", "--Select--")] + DomainApplication.StateTerritoryChoices.choices,
     )
     zipcode = forms.CharField(label="ZIP code")
+    urbanization = forms.CharField(label="Urbanization (Puerto Rico only)")
```
neilmb, line 94: **question**: I think that we want `required=False` on this field?

src/registrar/migrations/0006_alter_domainapplication_federal_agency.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 4.1.3 on 2022-12-12 14:48
```
neilmb, line 1: **question (non-bloking)**: I'm wondering if there is a merge conflict here. Did this migration already get made?

src/registrar/models/domain_application.py
```diff
@@ -28,62 +28,69 @@ class DomainApplication(TimeStampedModel):
...
-        WASHINGTON = "WA", "Washington"
-        WEST_VIRGINIA = "WV", "West Virginia"
-        WISCONSIN = "WI", "Wisconsin"
-        WYOMING = "WY", "Wyoming"
-        AMERICAN_SAMOA = "AS", "American Samoa"
-        GUAM = "GU", "Guam"
-        NORTHERN_MARIANA_ISLANDS = "MP", "Northern Mariana Islands"
-        PUERTO_RICO = "PR", "Puerto Rico"
-        VIRGIN_ISLANDS = "VI", "Virgin Islands"
+        ALABAMA = "AL", "Alabama (AL)"
```
neilmb, line 31: **comment:** This is great to see for keyboard accessibility of the states select box, thank you.

neilmb (approved):
Looks great, just a few questions.

## Accessibility
- ~New pages have been added to .pa11yci file so that they will be tested with our automated accessibility testing~
    - ~All form inputs have explicit labels~
    - ~All relevant images use an img tag~
    - ~All images have alt attributes~
    - ~Multimedia is tagged. All multimedia has appropriate captioning and audio description~
    - [X] Tab order is logical
    - ~Form instructions are associated with inputs~
    - ~Site doesn’t timeout unexpectedly. Identify elements that may “timeout” and verify that the user can request more time~
    - ~Tables are coded properly. Tables have proper headers and column attributes~
    - ~Frames are named. All frames have a name element~
    - ~Flashing elements are compliant. Elements that flash on screen do so at a rate of less than 3 Hz~
    - ~The language for sections on the page that differ from the site language are set~
    - ❌ CSS is not required to use the page. The page makes sense with or without CSS: Without the CSS, it feels pretty difficult to distinguish the section headers from their values and the following Edit buttons. Maybe some colons or blank lines? Or maybe it's unsolvable? 
    - [X] Links are unique and contextual. All links can be understood taken alone, e.g., ‘Read more - about 508’: hidden link text is very helpful
    - ~Required plugins are linked on the page~
 
## Device Matrix
- [X] firefox
- [X] chrome/chromium/edge
- ❌ safari: still can't log in
    - [X] at 480px: this is intensely better!
    - [X] at 1024px

src/registrar/forms/application_wizard.py
SSPJ: Good catch, fixed!

src/registrar/migrations/0006_alter_domainapplication_federal_agency.py
SSPJ: I am not sure about any merge conflict. I looked on `main` and didn't see any `0005` or `0006`.

I do see that I've included two migration files on this PR and there really only needed to be one, so I've condensed them.

SSPJ:
@neilmb Thanks for the review! 🙇 

The comment about it looking bad with no CSS relates to the structure of the elements on the page that I mentioned in the PR description. I'll try to grab some time with @igorkorenfeld tomorrow to work out a solution.

igorkorenfeld:
Looking nice! 

For the section `Requirements for registration and operation of .gov domains` could we put in some value in there if they have checked the box? I'm thinking "Agree"

Thanks for adding in the hidden text for SR, I think that's working well. 

I am far from an expert, but I did some manual testing with Voice Control, and it does seem like "Edit" + Field name selects the correct button, or at least presents the right options to choose from. And the number overlays that pop up seem to work well as a fallback. 

I did get a value error when entering "current sites". I typed in "nah" and got the output below. Might not be related to the PR itself though 
 
<h1 style="padding: 0px; margin: 0px; font-weight: normal; color: rgb(0, 0, 0); font-family: sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">ValueError at /register/current_sites/</h1><pre class="exception_value" style="padding: 0px; margin: 10px 0px; font-size: 1.5em; white-space: pre-wrap; word-break: break-word; font-family: sans-serif; color: rgb(87, 87, 87); font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">You must specify a tld for nah</pre>

Request Method: | POST
-- | --
http://localhost:8080/register/current_sites/
4.1.3
ValueError
You must specify a tld for nah
/app/registrar/models/domain.py, line 143, in normalize
registrar.forms.application_wizard.ApplicationWizard
/usr/local/bin/python
3.10.8
['/app',  '/usr/local/lib/python310.zip',  '/usr/local/lib/python3.10',  '/usr/local/lib/python3.10/lib-dynload',  '/usr/local/lib/python3.10/site-packages']

SSPJ:
> I did get a value error when entering "current sites". I typed in "nah" and got `You must specify a tld for nah`.

@neilmb The error that Igor surfaced is coming from `normalize` that is called when the website is being saved to the database. That saving is not performed unless the form's `is_valid` returns `True`, so am I correct in thinking that it will be taken care of by the work you are currently doing?

neilmb:
> The error that Igor surfaced is coming from normalize that is called when the website is being saved to the database. That saving is not performed unless the form's is_valid returns True, so am I correct in thinking that it will be taken care of by the work you are currently doing?

That is my goal, that we shouldn't ever see that exception because the form won't be valid if that method isn't satisfied.

Issue #284, opened 2022-11-22
User unable to progress form if Selecting "No" For Election Office


calerubensteingsa:
## Issues

- [ ] On the Type of organization — Election board Page, attempting to select No for an answer results in the current form page being refreshed when the user clicks next, user is unable to proceed

SSPJ:
Added an additional issue I found today.

loganmeetsworld:
Fixed in #292

PR #331, opened 2022-12-14
Application form field validation

### LINKED ISSUE #262: Add field validation for all form fields in the domain application request form
_Outcome_: Ensure all required fields are required and any validation we want on individual fields is implemented. 

*Requires design work*

We should ensure browser/frontend error handling is correct as well as validation for the state transition to submit. When there is an error on submission, we should give helpful instruction as to how to fix the error.

Should include the URL validation for available, unavailable, and non-compliant URLs

A/C: As a user I want timely and clear response when I input something that is not allowed on submission.

neilmb:
# Validate form fields with error messages #

## 🗣 Description ##

This adds field validation and the associated error messages for the form wizard. 

Screen Shot 2022-12-13 at 14 28 00

We get lots of this for "free" from the Django fields that we use, the only cost is to make sure we include the `{{ form.errors }}` content in our templates. There is a new template tag in `field_helpers` called `input_with_errors` to help with that for text boxes. For now, other field types should include the error information manually.

Validation unit tests are in a new `tests/test_forms.py` file. We got a new dependency on a Django phonenumber library to make the handling of all the different phone number formats much easier. The only programming change is that phone numbers are stored in the database in the full `+12015551234` format and then rendered onto the page using `{{ phone.as_national }}` to get the appropriate US format. It also checks phone numbers for valid area codes, so for testing we need to use `201-555-XXXX`.

Hopefully closes #262.


igorkorenfeld:
This is awesome work! 

It looks like maybe validation is missing on the `organization_contact` page? I was able to get to the next page with a few bad inputs, namely not choosing the agency and having a 1-digit zip code:
Screen Shot 2022-12-15 at 5 59 31 PM

src/registrar/forms/application_wizard.py
```diff
@@ -188,13 +207,42 @@ def from_database(self, obj):
...
+        if not requested:
+            # none or empty string
+            raise forms.ValidationError(
+                "Please enter the .gov domain that you are requesting.", code="invalid"
+            )
+        if requested.endswith(".gov"):
+            requested = requested[:-4]
+        if "." in requested:
+            raise forms.ValidationError(
+                "Please enter a top-level domain name without any periods.",
```
igorkorenfeld, line 235: This can a be a todo for later, if Michelle is going to do some content work on the error messages, but I think `top-level` might not be quite clear here

h-m-f-t: True. .gov is the top-level domain, everything else is just a "domain" (or a second-level domain, but we don't need to use the jargon).

SSPJ (changes requested):
Neil, great job taking on this work! This was a huge lift and I can tell you've worked very hard on it!🥇

I've put overall feedback at the top and then page-by-page checklists underneath. Validation is a large amount of work, so if you would like to break any of these items out into their own tickets, please let Cale know.

- [ ] Client side validation is causing some problems.
    * It does not display any error messages for radio buttons.
    * It does not apply the correct visual styling of the invalid form inputs via CSS.
    * The native validation message in Chrome for a required field is "Please fill out this field", but screen reader users are given no indication of _which_ field they are now on.

    My recommendation is to turn off client side validation with the use of `novalidate` HTML attribute, allowing the user to submit the form with errors (including missing "required" fields) validate on the server, and (re-)render the template with the correct visual styling and aria-invalid attributes.
- [ ] Missing aria: invalid fields need `aria-invalid="true"`. It wouldn't hurt to set `aria-errormessage` as well, although support for this attribute is still spotty.

    
    (in this image, "invalid user entry" is False, despite the email address being invalid)
- [ ] Inconsistent instructional text.

    Every page needs to inform users of which fields are required. This can be accomplished one of two ways. cc @michelle-rago and @igorkorenfeld 
        1. (current style) Put "All fields are required unless they are marked optional." Then put `(optional)` on each field which is not required.
        2. ([recommended](https://designsystem.digital.gov/templates/form-templates/address-form/)) Use

            <p>
              Required fields are marked with an asterisk (<abbr
                title="required"
                class="usa-hint usa-hint--required"
              >*</abbr>).
            </p>

    and

      <label>
        Label <abbr title="required" class="usa-hint usa-hint--required">*</abbr>
      </label>

- [ ] Save button integration with validators: on each page, if I press "Save", the page I am currently working on is refreshed from the database. No message appears to explain what has happened: "Saving" simply and mysteriously erases my work.

    This is likely due to the fact that saving is not performed if a form is invalid. Desired behavior is likely to see the same errors as with "Next" (and get my WIP back!) with a message that "Some information could not be saved due to errors."

    (Perhaps we'd like saving logic to go field-by-field, rather than form-by-form, but that is out of scope for this validation work.)

**Organization type**
* Page is missing instructional text

**Organization type - Federal**
* Page is missing instructional text

**Organization type - Election Board**
* Selecting "No" and then clicking "Next" results in an error.
Type of organization - Election board page

**Organization name and mailing address**
* Federal agency dropdown does not have the `required` attribute.
* Street address line 2 does not have `(optional)` designation as required by "All fields are required unless they are marked optional."

**Authorizing official**
* Phone validator library is kind of impressive.
* Random note: the box is not big enough for extensions after formatting:
Phone number box

**Organization website**
* Any website should be allowed.
Organization website page
(I am using http://www.hartford-hwp.com/Taino/jatibonicu.html as test data.)

**.gov domain**
* The validator from #236 needs to be hooked up, per the issue ticket.
* The error message could use a bit of wordsmithing
What  gov domain do you want page
* The page did not perform an availability check when I clicked next. I'm not sure if this is a requirement or not... if a user does _not_ click the "Check availability" button, are they allowed to request an already taken domain? I think this is probably not desirable.
* Alternative domain allowed me to save "-okyeah" and "ok.yeah" without an error.

**Purpose of your domain**
* Page is missing instructional text

**Security email for public use**
* Page is missing instructional text
* Security email is missing `(optional)` designation

**Anything else we should know?**
* Page is missing instructional text
* text box is missing `(optional)` designation

**Requirements**
* Page is missing instructional text
* Checkbox is missing `required` attribute

neilmb:
From @SSPJ on a related PR:

Steps to reproduce

Select "Federal", click "next"
Answer the question, click "next"
Fill out the org contact, click "next"
Go back to step 1, change to "City", click "save"
Skip question number 2 ("Election board")
Continue filling out the form. On the review page, click "Submit"

Expected result: I am returned to the Election Board page and presented with an error message. The radio button inputs are marked invalid.

Actual result: I am returned to the Election Board page. No error appears. Radio buttons are required but not aria-invalid.

src/registrar/forms/application_wizard.py
neilmb: Changed to just say "Please enter a domain without any periods."

SSPJ (approved):
As discussed in Engineering Huddle, the scope of ticket #262 was too broad and this PR is to focus on scaffolding error messaging code.

Reviewing from that perspective, this looks good to me. :+1: Let's get it merged!

Below is the summary of validation issues from my second review pass. Before merging, Neil, please have a conversation with @calerubensteingsa to group and scope this (plus any others you've got) into follow up tickets (if they don't already exist).

Follow up list:
* Instructional text missing on Type of organization - Federal, .gov domain, Purpose of your domain, and Requirements for registration and operation of .gov domains
* Organization name and mailing address - Federal agency dropdown is not marked required and does not get an error message; it is possible to bypass that question
* Organization website - Currently allows entry of websites, but strips path data when saving to the database (e.g. `http://www.hartford-hwp.com/Taino/jatibonicu.html` becomes `hartford-hwp.com`)
* Integrate the domain validation JavaScript code into the "Check availability" button.
* Check availability of requested domain on saving.
* Decide upon correct level of validation for alternative domain requests and implement it.
* Remove client-side validation from the .gov domain requested domain.
* Purpose of your domain page is missing `aria-invalid`.
* Save button integration with validators
* Decide how to integrate the final review page submission with error handling (send users back, show errors on the review page, etc) and implement it.

Type of organization - Election board
- [x] helpful error text, yay :tada:
- [x] "select no" bug is fixed, yay :tada: -- Neil, you can update this PR description to close #284 as well 🙂 

Organization name and mailing address
- [x] field width of Urbanization code is fixed, yay :tada:
- [x] 1 digit zip codes are no longer allowed, yay :tada:

neilmb:
Follow-ups:

- Organization website - Currently allows entry of websites, but strips path data when saving to the database (e.g. http://www.hartford-hwp.com/Taino/jatibonicu.html becomes hartford-hwp.com)

This is still the case. I think that it's correct behavior, but I don't think that we have discussed it yet as a team.

- Integrate the domain validation JavaScript code into the "Check availability" button.
- Check availability of requested domain on saving.

These are now in #358.

- Decide upon correct level of validation for alternative domain requests and implement it.

This still needs to be discussed with design how we want it to function.

- Save button integration with validators
- Decide how to integrate the final review page submission with error handling (send users back, show errors on the review page, etc) and implement it.

After the change in our forms, we still need to figure these two out.

PR #344, opened 2022-12-21
Refactor fixtures from JSON to Python

SSPJ:
Does what it says on the tin. I've also added some new fixtures.

The motivation was that JSON fixtures are a bit fragile, seeing that they rely on hardcoded primary keys.


src/registrar/fixtures.py
```diff
@@ -0,0 +1,256 @@
...
+            "username": "d4c3bd84-dc3a-48bc-a3c3-f53111df2ec6",
+            "first_name": "Igor",
+            "last_name": "",
+        },
+        {
+            "username": "ee80bfe0-49ad-456d-8d82-e2b608a66517",
+            "first_name": "Logan",
+            "last_name": "",
+        },
+    ]
```
SSPJ, line 44: @michelle-rago @boonerang @calerubensteingsa @neilmb @h-m-f-t 

Currently refactoring (and expanding) our data fixtures and thought I'd ask again if any of you would care to have your account ID from the Login.gov sandbox included here.

Also, if you want some custom fixture data (beyond being an admin and having a few example domain applications), I'm happy to take requests. CC @loganmeetsworld  @igorkorenfeld on that last sentence.

michelle-rago: @SSPJ What does adding our sandbox account here get us?

loganmeetsworld: I'm good with what is there for me.

src/registrar/fixtures.py
```diff
@@ -0,0 +1,256 @@
...
+        },
+        {
+            "status": "submitted",
+            "organization_name": "Example - Submitted but pending Investigation",
+        },
+        {
+            "status": "investigating",
+            "organization_name": "Example - In Investigation",
+        },
+    ]
```
loganmeetsworld, line 114: These are a little funky when loaded for me - the fake data doesn't seem to be loaded when I go to the form.

loganmeetsworld (approved):
The Faker library is cool! I noted that I saw some things might've been missing. I wonder too if how the validation PR might work with this because of that but overall LGTM

src/registrar/fixtures.py
SSPJ: @michelle-rago Unsure if you are asking what _this file_ does currently or what _data fixtures_ in general do, so answering both.

**Data fixtures:** We periodically have to recreate the database in staging or unstable. When we do that, we can run the code in this file to generate data. If you'd like to have fake data associated with your user, this is where we'd put it. An example might be a page that you can only access once you have a domain approved: if you'd like to check the wording on that page, you'd need to create an application, submit it, then approve it. A data fixture speeds the process by giving you an already approved domain.

**Currently:** The `DomainApplicationFixture` (further down in the file) randomly generates about 4 applications per user, in various states. If you have more specific requests, I'm happy to add them!

PR #350, opened 2023-01-03
add no-input flag to flush so it can pass

loganmeetsworld:


SSPJ (approved):
LGTM 👍 Thanks for this fix, Logan !

neilmb (approved):

Issue #321, opened 2022-12-09
Add ability for additional items under certain fields


loganmeetsworld:
Currently "other contacts", organization website, and alternative domains have no mechanism to add more than one. We'd like to be able to allow users to add additional of these items in the form.

calerubensteingsa:
12/20:

Ran into blocker where library doesn't allow multiple forms to one page - needs to be separate form objects

calerubensteingsa:
12/22:

- Possible merge conflicts to work out with changes to form library

calerubensteingsa:
1/17:

- Review feedback

loganmeetsworld:
#365

PR #353, opened 2023-01-04
Remove formtools

SSPJ:
## 🗣 Description ##

This PR moves the relevant formtools library code into our own codebase.

Doing this, we can more easily extend the code with needed features, such as multiple Django forms per page.

<details>
<summary>⚠️ Caveat: this PR introduces a feature regression.</summary>

Applications are no longer checked for validity on the final review step before submission.

The reason is in how formtools did this vs how the new code does it. Formtools used saved POST data to revalidate each form _as the user had entered it_. This implementation discards that data and tries to instantiate the forms from the database in order to perform validation.

This must be possible in Django (after all, that is how ModelForms work), but is presently not working: the form claims it is invalid, even when careful checking via `breakpoint()` shows that the form object contains valid data.

The plan is to fix this in the PR for #321 after #331 merges.
</details>


src/registrar/forms/application_wizard.py
```diff
@@ -39,10 +34,14 @@ def to_database(self, obj: DomainApplication | Contact):
             setattr(obj, name, value)
         obj.save()
 
-    def from_database(self, obj: DomainApplication | Contact):
+    @classmethod
+    def from_database(cls, obj: DomainApplication | Contact | None):
         """Initializes this form's fields with values gotten from `obj`."""
```
neilmb, line 39: **question**: It seems like the contract for this method changed from initializing form fields to returning a dict of fields. Does this comment need to be updated to match?

src/registrar/forms/application_wizard.py
```diff
@@ -112,11 +111,11 @@ def to_database(self, obj):
             obj.authorizing_official = contact
             obj.save()
 
-    def from_database(self, obj):
+    @classmethod
+    def from_database(cls, obj):
         """Initializes this form's fields with values gotten from `obj`."""
```
neilmb, line 116: **suggestion (non-blocking):** It seems like the docstrings for these subclass overrides are copy-pasted from the original method. I think that we could either leave them out if they are the exact same, or use them instead to explain what is specifically different about this subclass's version of the method.

src/registrar/templates/base.html
```diff
@@ -129,7 +129,7 @@
         {% block logo %}
         <div class="usa-logo" id="extended-logo">
           <em class="usa-logo__text">
-            <a href="/" title="Home" aria-label="Home">
+            <a href="{% url 'home' %}" title="Home" aria-label="Home">
```
neilmb, line 132: **comment**: Thank you for all of these changes to using named URLs in the templates!

src/registrar/views/application.py
```diff
@@ -0,0 +1,421 @@
...
+
+from registrar.forms import application_wizard as forms
+from registrar.models import DomainApplication
+from registrar.utility import StrEnum
+from registrar.views.utility import StepsHelper
+
+logger = logging.getLogger(__name__)
+
+
+class Step(StrEnum):
```
neilmb, line 18: **suggestion (non-blocking)**: Because the `StrEnum` class is non-standard, how about a comment here that the order in which these constants are defined will be the order that the steps appear? At least I think that is the way that it will work?

src/registrar/views/application.py
```diff
@@ -0,0 +1,421 @@
...
+    def application(self) -> DomainApplication:
+        """
+        Attempt to match the current wizard with a DomainApplication.
+
+        Will create an application if none exists.
+        """
+        if self._application:
+            return self._application
+
+        if self.has_pk():
```
neilmb, line 115: **suggestion (non-blocking):** For locality of reference, could we move the definition of the `has_pk` method above this property? It mixes an ordinary method in with a section full of properties, but it felt slightly strange to have to look forward to find out what that method does.

src/registrar/views/application.py
```diff
@@ -0,0 +1,421 @@
...
+        )
+        self.storage["application_id"] = self._application.id
+        return self._application
+
+    @property
+    def storage(self):
+        # marking session as modified on every access
+        # so that updates to nested keys are always saved
+        self.request.session.modified = True
+        if self.prefix not in self.request.session:
```
neilmb, line 137: **suggestion (non-blocking):** I'm not sure if it's any better that your explicit version here, but this pattern is Python's `setdefault` method: `return self.request.session.setdefault(self.prefix, {})`.

src/registrar/views/application.py
```diff
@@ -0,0 +1,421 @@
...
+        """This method handles GET requests."""
+
+        current_url = resolve(request.path_info).url_name
+
+        # if user visited via an "edit" url, associate the id of the
+        # application they are trying to edit to this wizard instance
+        if current_url == self.EDIT_URL_NAME and "id" in kwargs:
+            self.storage["application_id"] = kwargs["id"]
+
+        # if accessing this class directly, redirect to the first step
```
neilmb, line 189: **suggestion (non-blocking)**: I think that this comment could be even more explicit about how this won't apply to the rest of our views which are sub-classes (?) and so GET requests for the other views won't return us to the first step?

src/registrar/views/application.py
```diff
@@ -0,0 +1,421 @@
...
+    def get_context_data(self):
+        context = super().get_context_data()
+        context["Step"] = Step.__members__
+        context["application"] = self.application
+        return context
+
+    def goto_next_step(self):
+        return self.done()
+        # TODO: validate before saving, show errors
+        # forms = self.get_all_forms()
```
neilmb, line 404: **suggestion (non-blocking)**: Say more about why this is a TODO, why the code commented out doesn't just work. Maybe link to a ticket that explains more?

neilmb (approved):
It is so great to see this working and passing the tests. Lots of suggestions for minor improvements, but overall, I think it will be great to see this merged so we can move forward. 

Is there an ADR that should go along with this same PR, or some additional documentation in `docs/developer` on how to use our version of `ApplicationWizard`?

src/Pipfile
```diff
@@ -17,7 +17,6 @@ oic = "*"
 pyjwkest = "*"
 psycopg2-binary = "*"
 whitenoise = "*"
-django-formtools = "*"
```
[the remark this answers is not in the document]
SSPJ: Yikes! Good catch, thank you.

src/registrar/config/urls.py
```diff
@@ -9,30 +9,62 @@
...
-APPLICATION_URL_NAME = "application_step"
-application_wizard = ApplicationWizard.as_view(
-    url_name=APPLICATION_URL_NAME,
-    done_step_name="finished",
-    condition_dict=WIZARD_CONDITIONS,
+application_urls = (
+    [
+        path("", views.ApplicationWizard.as_view(), name=""),
+        # dynamically generate the other paths
+        *[
```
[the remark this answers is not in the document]
SSPJ: I've restructured it as you suggested, hopefully will be easier for future maintainers now.

src/registrar/forms/application_wizard.py
SSPJ: They were copy-pasta; I've removed them from all except the base class.

src/registrar/views/application.py
SSPJ: Comment added.

src/registrar/views/application.py
SSPJ: Moved `has_pk`; that was a good suggestion.

src/registrar/views/application.py
SSPJ: I do prefer idiomatic code; I've updated it to use `setdefault`.

src/registrar/views/application.py
SSPJ: Inserted longer comment here.

src/registrar/views/application.py
SSPJ: Done.

Issue #306, opened 2022-12-08
Update new domain request form contacts language to reflect decision around "key personnel"


calerubensteingsa:
See context for decision on [how to handle contacts 🔒](https://gsa-tts.slack.com/archives/C03QM0JGSQG/p1670516467308479). 

Per standup 12/15: https://docs.google.com/document/d/1jWjJdi6LlLMTGMN0KPpN9BieqDn-DypoyI8NIFXPhaI/edit?usp=sharing

Acceptance Criteria

- [ ] Update Registrant form design to ask for key contacts and updated language: https://docs.google.com/document/d/1ieO5Qm18dMCS04v-HdF6zKSDdzDWSmAyVjUAI21RyRU/edit#heading=h.2dydexc00jeq
- [ ] Add a field for registrants to provide reasoning if they don't add key contacts

PR #365, opened 2023-01-13
Add ability for additional items under certain fields

### LINKED ISSUE #321: Add ability for additional items under certain fields

SSPJ:
## 🗣 Description ##

When the user clicks "Add an extra \<thing\>" (wording not exact), their current progress is saved and an additional blank field is loaded into the page.

Discussion with the team confirmed that deletion of extra items is an enhancement that we'll look to add later.

Additionally, the linked issue specifies that organization website should allow additional entries. I was not sure that was correct, @calerubensteingsa or @michelle-rago, is that true? For now, this PR includes other contacts and alternative domains.

Lastly, some funkiness remain around how required fields are handled. I can work that out here or do it as part of #306 -- my preference is to understand what changes need to be made on #306 first, but I can do the actual work in this PR, if that is sensible.

Closes #321


michelle-rago:
Yes, we do ask users if they have more than one website. Here's the text:

> Enter your organization’s public website, if you have one. For example, www.city.com. 
> [url field]
> 
> If you have more than one website, add another.

neilmb (commented):
Some general observations:

- Clicking "add another" is just another way of saving and so it validates the fields on the page and gives error messages for missing required fields, etc. If there are errors, it doesn't add another field.

- The error handling on the "Other Contacts" page seems to be broken by the switch to formsets there. Clicking "Next" on a completely empty form doesn't provide any error messages for required fields, but it also doesn't go to the next page.

- First name/given name seems to have lost its asterisk. `label_suffix=REQUIRED_SUFFIX`?
 
- There is an extra www .gov under Alternative domains because forms.0.alternative_domain doesn't exist any more in that form, it's all in forms.1.
Screen Shot 2023-01-13 at 11 38 06


## Accessibility
- [ ] ~New pages have been added to .pa11yci file so that they will be tested with our automated accessibility testing~
    - [ ] ~All relevant images use an img tag~
    - [ ] ~All images have alt attributes~
    - [ ] ~Multimedia is tagged. All multimedia has appropriate captioning and audio description~
    - [X] Tab order is logical
    - [ ] ~Site doesn’t timeout unexpectedly. Identify elements that may “timeout” and verify that the user can request more time~
    - [ ] ~Tables are coded properly. Tables have proper headers and column attributes~
    - [ ] ~Frames are named. All frames have a name element~
    - [ ] ~Flashing elements are compliant. Elements that flash on screen do so at a rate of less than 3 Hz~
    - [ ] ~The language for sections on the page that differ from the site language are set~
    - [ ] CSS is not required to use the page. The page makes sense with or without CSS
       
       Without CSS, the Add a Contact button is disruptively large. Can we add an explicit `width` in the HTML to make it smaller in that situation?
    
        Screen Shot 2023-01-13 at 11 24 38

    - [ ] ~Required plugins are linked on the page~
 
## Device Matrix
- [X] firefox
- [X] chrome/chromium/edge
- [ ] safari: Still can't log in
    - [X] at 480px
    - [X] at 1024px

neilmb (changes requested):
This is fantastic functionality, but I would love to see a unit test for at least one of these forms to verify that it works as expected with saving into the database. 

Somthing like: Open the register URL, fill in the pages up to .gov domain, fill in the .gov domain and first alternative domain, click "Add another" and verify that there are two Alternative domain fields on the resulting page and that there is a new `Website` object in the database matching the one that was entered and that it is correctly referenced in the ManyToOne relationship of the DomainApplication.

There are more versions of this test that could be possible to verify the deleting of form fields and corresponding database entries, but I think that those could be optional.

neilmb:
And finally, error messages don't work, probably because formsets carry their error attributes in different places. The base `application_form.html` checks in `form.errors` but with a formset that isn't the right place to look to see if there are errors.

This could be fixed in follow-up work, but I think that we want error handling and validation to work on these pages as well. I think that it's less important for alternative domains since they can be blank, but for Other Contacts I think that we should want it to work.

SSPJ (commented):
@neilmb I believe this is ready for another pass.

Validation on the fields of the formset is working; deletion is TBD (can delete non-required fields by erasing them, but that measure is not sufficient for contacts); tests are still admittedly sparse, but I've included one for current sites. The issue with being unable to "add another" while fields are invalid is not fixed. I can think of a potential fix, but it would not be simple to implement.

The refactoring work I've done on `input_with_errors` is stashed. I didn't need it for this branch (although we'd fail a rigorous 508 audit because some fields are missing their `aria-describedby` attributes), but perhaps I will include it on the domain availability check PR, as that will require a custom attribute `validate="domain"`.

neilmb (approved):
LGTM, thanks for all the changes. So long "Test site"!

PR #366, opened 2023-01-17
Change next and save button text

igorkorenfeld:
## 🗣 Description ##
This PR change the "Next" button text in the form to "Save and continue", and "Save" to "Save progress"

This changes comes from a discussion in our design huddle a few weeks ago. Our hunch is that the updated text will be clearer, and we'd like to test the new language proposed in this PR with users to see if they understand that their application is saved as they progress and that they have the option to pause and pick up the application again.


neilmb (approved):
LGTM, easy shmeezy.

PR #377, opened 2023-01-20
fix issue confusing head vs base branch syntax

loganmeetsworld:


SSPJ (approved):
Without knowing if this will work, it looks good to me on a quick inspection of the code. 👍

Issue #330, opened 2022-12-14
Add dynamic helper text based on organization for AO and .gov pages


calerubensteingsa:
**Acceptance Criteria**

Per slack: https://gsa-tts.slack.com/archives/C03QM0JGSQG/p1671035829341489

- Add logic to show dynamic helper text based on organization type
- Add text/files changes @igorkorenfeld

loganmeetsworld:
first part complete in #378, .gov pages work to come

PR #378, opened 2023-01-20
Dynamic helper text for Authorizing Official

igorkorenfeld:
## 🗣 Description ##

This PR adds content for the authority official for each type of government organization, and and implements logic to display the relevant text based on the registrant's organization type. 

Part of #330 . If this approach and implementation looks good, I will extend it to the .gov page as well.

### DIFF
--- a/src/registrar/templates/application_authorizing_official.html
+++ b/src/registrar/templates/application_authorizing_official.html
@@ -13,7 +13,29 @@
   </p>
 
   <div class="ao_example">
-    {% include "includes/ao_example__school-district.html" %}
+    {% if application.organization_type == 'federal' %}
+      {% if application.federal_type == 'executive' %}
+        {% include "includes/ao_example__exec.html" %}
+      {% elif application.federal_type == 'judicial' %}
+        {% include "includes/ao_example__judicial.html" %}
+      {% elif application.federal_type == 'legislative' %}
+        {% include "includes/ao_example__legislative.html" %}
+      {% endif %}
+    {% elif application.organization_type == 'city' %}
+        {% include "includes/ao_example__city.html" %}
+    {% elif application.organization_type == 'county' %}
+        {% include "includes/ao_example__county.html" %}
+    {% elif application.organization_type == 'interstate' %}
+        {% include "includes/ao_example__interstate.html" %}
+    {% elif application.organization_type == 'school_district' %}
+        {% include "includes/ao_example__school-district.html" %}
+    {% elif application.organization_type == 'special_district' %}
+        {% include "includes/ao_example__special-district.html" %}
+    {% elif application.organization_type == 'state_or_territory' %}
+        {% include "includes/ao_example__state-territory.html" %}
+    {% elif application.organization_type == 'tribal' %}
+      {% include "includes/ao_example__tribal.html" %}
+    {% endif %}
   </div>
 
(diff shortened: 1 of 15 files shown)


src/registrar/templates/application_authorizing_official.html
```diff
@@ -12,8 +12,30 @@ <h2>Who is the authorizing official for your organization?</h2>
   <p>Your authorizing official is the person within your organization who can authorize your domain request. This is generally the highest ranking or highest elected official in your organization. Read more about <a href="{% url 'todo' %}">who can serve as an authorizing official</a>.
   </p>
 
-  <div class="ao-example">
-    {% include "includes/ao_example__city.html" %}
+  <div class="ao_example">
+    {% if is_federal %}
+      {% if federal_type == 'executive' %}
```
SSPJ, line 17: For code organization, I might have favored putting this `if` block in an include and all of the example text inside of it over multiple smaller files. That's a matter of preference and this is fine.

src/registrar/templates/includes/ao_example__exec.html
```diff
@@ -0,0 +1,3 @@
+<p>Domain requests from executive branch agencies must be authorized by <strong>CIOs</strong> or <strong>agency heads</strong>.</p>
```
SSPJ, line 1: I would very much prefer that all abbreviations be [eliminated or at least defined on first usage](https://www.plainlanguage.gov/guidelines/words/minimize-abbreviations/) (within a page):

`<strong>Chief Information Officers</strong>` or 

`<strong>Chief Information Officers (<abbr>CIO</abbr>s)</strong>`

CC @michelle-rago

src/registrar/tests/test_views.py
```diff
@@ -740,6 +740,70 @@ def test_application_type_of_work_interstate(self):
 
         self.assertContains(contact_page, self.TITLES[Step.TYPE_OF_WORK])
 
+    def test_application_ao_dynamic_text(self):
+        type_page = self.app.get(reverse("application:")).follow()
```
SSPJ, line 744: Kudos for inclusion of tests 💯

SSPJ (approved):
LGTM, @igorkorenfeld ! 👍 

The Basics
- [x] I read the code
- [x] I ran the code

Accessibility
- [x] Meets WCAG 2.0 AA or 2.1 AA for Section 508 compliance
    - ❌  Form instructions are associated with inputs
    We are missing some `aria-describedby` attributes; but these are out of scope of this PR (your code does not introduce any new issues, I'm noting an existing issue for awareness).

src/registrar/views/application.py
```diff
@@ -365,6 +365,13 @@ class AuthorizingOfficial(ApplicationWizard):
     template_name = "application_authorizing_official.html"
     forms = [forms.AuthorizingOfficialForm]
 
+    def get_context_data(self):
+        context = super().get_context_data()
+        context["is_federal"] = self.application.is_federal()
```
neilmb, line 370: **thought (non-blocking)**: This is the second form  that needed `context["is_federal"]`. Maybe that is common enough that we could move it into the base class and provide that context in all of the forms? The change doesn't need to be made here, just noticing that we reached the second gate in the "Rule of Three".

igorkorenfeld: Ah good point, and we will likely reach the 3rd gate when doing the .gov domain examples. I can change it as part of that work.

PR #411, opened 2023-02-13
Send confirmation email on application submission

### LINKED ISSUE #319: Proof of concept derisking for email notifications
_Outcome_: An end-to-end proof of concept sending a simple email from our application to a test user account. For example, emails the applicant when their domain request is submitted.

This covers the ability for us to send existing registrants and domain applications emails to notify them of actions taken on their applications. We have many use cases for being able to email applicants and domain managers. This ticket doesn't include implementing any of those specific emails, but rather derisking the infrastructure needed for it. 

- [ ] Set up access for dev team to CISA AWS account and ensure SES is present (blocked pending access)
- [ ] Setup SES to send emails
- [ ] Find a client that works for connecting to SES through our application
- [ ] Figure out a way to store SMTP credentials so our application has access
- [ ] Related, connect SES to our cloud.gov account [through a brokerpak](https://github.com/GSA-TTS/datagov-brokerpak-smtp) or some other means. 

Helpful resources for figuring this out will be the data.gov team that wrote the broker, the Notify team, and other folks that have made this work with cloud.gov. 

This will be blocked on our access to AWS accounts.

neilmb:
# Use AWS SES to send a confirmation email on application submission  #

## 🗣 Description ##

This sends a very simple confirmation email when a domain application is submitted. It uses the CISA AWS SES infrastructure that has been made available to us. 

This change introduces two new secret credentials for our environments for access to AWS.

(black also wanted some whitespace-only changes too.)

Closes #319.


src/registrar/utility/email.py
```diff
@@ -0,0 +1,34 @@
...
+
+    template = get_template(template_name)
+    email_body = template.render(context=context)
+
+    ses_client = boto3.client(
+        "sesv2",
+        region_name=settings.AWS_REGION,
+        aws_access_key_id=settings.AWS_ACCESS_KEY_ID,
+        aws_secret_access_key=settings.AWS_SECRET_ACCESS_KEY,
+    )
```
loganmeetsworld, line 24: What will happen in environments where there are not these variables set?

src/registrar/utility/email.py
```diff
@@ -0,0 +1,34 @@
+"""Utilities for sending emails."""
+
+import boto3
+
+from django.conf import settings
+from django.template.loader import get_template
+
+
+def send_templated_email(template_name: str, to_address: str, context={}):
```
SSPJ, line 9: **Non-blocking:** Thinking ahead to the future, how is this signature going to accommodate TXT + HTML messages and messages with To/Cc/Bcc?

It's fine as is, but if you want to improve it before it is used in many other places, it could save your future self some refactoring headache.

src/registrar/utility/email.py
SSPJ, line 24: I'd suggest the use of `max_retries` and `mode` as illustrated [in the docs](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html#defining-a-retry-configuration-in-a-config-object-for-your-boto3-client). The default is legacy mode with 5 tries, which is probably too many tries for a synchronous request. The docs say that standard mode has 3 tries, which seems reasonable, but I'd use `max_retries` as well, just on the principle that explicit is better than implicit.

src/registrar/utility/email.py
```diff
@@ -0,0 +1,34 @@
...
+    ses_client.send_email(
+        FromEmailAddress=settings.DEFAULT_FROM_EMAIL,
+        Destination={"ToAddresses": [to_address]},
+        Content={
+            "Simple": {
+                "Subject": {"Data": "Thank you for applying for a .gov domain"},
+                "Body": {"Text": {"Data": email_body}},
+            },
+        },
+    )
```
SSPJ, line 34: If SES is unreachable when the applicant submits their domain application, they are going to get a generic 500 page. 🟥  😢 

The error handling strategy depends on the message*, I think, but the details of why a message failed to send are unlikely to be of interest to the caller, so I recommend logging the real error and then propagating a more generic error upward.

The boto3 library has a pretty useful [page on error handling](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html).

\* For example, in the case of domain application submitted confirmation, it probably doesn't matter at all. In the case of notifying the authorizing official, maybe the system should leave a note for the analysts on that domain request about the failure to send. In the case of a soon-to-expire notice, maybe the domain gets flagged for attention. Etc...

SSPJ (changes requested):
Great work here, Neil! Our registrar is really starting to feel like a real web service!

I see that Michelle wasn't able to test it in the sandbox so I'm going to hold off on that for a bit.

The Basics
- [x] I read the code
- [ ] I ran the code
 
Documentation
    - [x] using plain language
- ❌  licenses for new dependencies are recorded, if any
- [x] licenses for updated dependencies have been checked for accuracy, if any
- ❌  new components have been added to the system diagram
    I'm honestly not sure that we've been maintaining a "system diagram", but I'd like to see an ADR on our choice to use SES.
 
Security
- [x] significant user actions are logged to a database table
 
Reliability
- ❌  good level of error handling
    - ❌  interactions with external systems are wrapped in try/except

src/registrar/utility/email.py
neilmb:
```
[14/Feb/2023 15:39:55] WARNING [registrar.models.domain_application:490] Failed to send email, no credentials are present
```

loganmeetsworld: I think that is the case - I don't think every environment needs to have these credentials.

src/registrar/utility/email.py
neilmb: I'll catch the error in the `_send_confirmation_email` method and log that it happened and then ignore it there because it seems like the right choice for this particular situation.

neilmb:
> I'm getting an error when I submit a request in your sandbox.

@michelle-rago @SSPJ I think you should be able to submit successfully on my sandbox now.

src/registrar/utility/email.py
neilmb: Personally, I'm happy with us refactoring this in the future when we need more complicated messages.

SSPJ (approved):
LGTM 👍 Nice work, @neilmb !

src/registrar/templates/emails/submission_confirmation.txt
```diff
@@ -0,0 +1,4 @@
+Thank you for submitting an application for the domain name "{{ domain_name }}".
+
+If you need to make changes to your application, visit
+<http://registrar.get.gov/application/{{ id }}/edit>.
```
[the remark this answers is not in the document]
michelle-rago: Yes, the submission worked and I got the email!!

Issue #398, opened 2023-02-06
Add confirmation on each page when there is a successful save with no errors


calerubensteingsa:
*User Story:* As a user of the registrar I want to know when my data is saved so I can be assured it's not lost
*Clarify:* A successful save is all required information on a question is valid upon save

**Acceptance Criteria**

When a there is a successful save, a [confirmation message](https://www.figma.com/proto/ninNjo0bTg7XfB4mvTp4hG/GetGov%2FDesigns?page-id=2798%3A31463&node-id=2798%3A32317&viewport=-133%2C960%2C0.37&scaling=min-zoom&starting-point-node-id=2798%3A32317&show-proto-sidebar=1) appears at the top of the form confirming the save was successful and there were no errors. 

![Image](https://user-images.githubusercontent.com/10510566/218600729-55ec9c01-852d-4221-9ae6-0b46730eaf31.png)

loganmeetsworld:
This is related to but does not include when there is _invalid data_ or errs, which is handled in #359

kewlguy781:
This is almost done. May will be ready for review tomorrow (Wednesday 2/15)

kewlguy781:
Done.

Issue #359, opened 2023-01-10
If there is an error on Save for a given page, save any valid data to the database, and give a caution message with context about what was saved


calerubensteingsa:
*Context*: Right now if you fill out a bunch of data and hit “Save Progress” and there is invalid data, there will be that error, but also it won’t save anything that you entered that was valid. We need to save anything on the page that is valid even when there is an error and give context that partial information has been saved but anything that erred has not been saved.

**Acceptance Criteria**

When there is an error on the page after hitting "Save Progress"... 
- Only save valid data entered
- Add a visual indicator with copy describing that valid information has been saved but any fields with errors have not been saved
- Show in-line errors on the page

See the example here from [Figma](https://www.figma.com/proto/ninNjo0bTg7XfB4mvTp4hG/GetGov%2FDesigns?page-id=2798%3A31463&node-id=2800%3A33123&viewport=-133%2C960%2C0.37&scaling=min-zoom&starting-point-node-id=2800%3A33123&show-proto-sidebar=1):
![Image](https://user-images.githubusercontent.com/10510566/218601246-14bea912-583a-420d-8565-faacc55a597b.png)

There is an in-line error above for the field that is incorrect (`Street address`) but other information (`Federal Agency` and `Organization Name`) should be saved as they are valid data. Right now nothing would be saved including the valid data. Then there is a caution box explaining that anything with an error was not saved.

igorkorenfeld:
Designs for the indicators:

[Success](https://www.figma.com/proto/ninNjo0bTg7XfB4mvTp4hG/GetGov%2FDesigns?page-id=2798%3A31463&node-id=2798%3A32317&viewport=-133%2C960%2C0.37&scaling=min-zoom&starting-point-node-id=2798%3A32317&show-proto-sidebar=1) – Use the success alert

[Invalid data](https://www.figma.com/proto/ninNjo0bTg7XfB4mvTp4hG/GetGov%2FDesigns?page-id=2798%3A31463&node-id=2800%3A33123&viewport=-133%2C960%2C0.37&scaling=min-zoom&starting-point-node-id=2800%3A33123&show-proto-sidebar=1)

kewlguy781:
The caution dialog have been added. now researching on method of this and see if this is possible.

kewlguy781:
[...]

PR #414, opened 2023-02-16
Add confirmation and caution message to the forms

kewlguy781:
This complete ticket #398 . Also in addition to this, I have added caution message to this. As well as "commenting" out the error form on the top of the form. The error will still show under the field. 

## 🗣 Description ##

* Added messages to forms
* Added confirmation message when saved progress
* Changed the error message in form to caution
* Commented out the error message on the top of form
* Changed the message in base to only show if it have extra_tag base

Closes ticket #398. Also did caution message which is already shown according to UX design (we already have ticket (359) for caution but that ticket focus on be able to save partial fields instead of the UI)


src/registrar/templates/includes/form_messages.html
```diff
@@ -0,0 +1,11 @@
+{% if messages %}
+{% for message in messages %}
```
neilmb, line 2: **question**:This will show even messages that are tagged as "base", leading to duplicates with any that show up in `base.html`. Do you want to only show messages here with `'base' not in message.extra_tags`?

src/registrar/views/application.py
```diff
@@ -319,13 +321,26 @@ def post(self, request, *args, **kwargs):
...
             # unless there are errors
+            # no sec because this use of mark_safe does not introduce a cross-site
+            # scripting vulnerability because there is no untrusted content inside.
+            # It is only being used to pass a specific HTML entity into a template.
+            messages.warning(
+                request,
+                mark_safe(  # nosec
+                    "<b>We could not save all the fields.</b><br/> The highlight "
+                    + "fields below <b>could not be saved</b> because they have "
+                    + "missingor invalid data. All other information on this page "
```
neilmb, line 332:
```suggestion
                    + "missing or invalid data. All other information on this page "
```

src/registrar/views/application.py
```diff
@@ -319,13 +321,26 @@ def post(self, request, *args, **kwargs):
...
+                    + "has been saved."
```
neilmb, line 333: **question**: This says that "All other information on this page has been saved." but I don't think that is true right now. Is there something else we could say that would be more accurate?

src/registrar/views/application.py
```diff
@@ -319,13 +321,26 @@ def post(self, request, *args, **kwargs):
             self.save(forms)
         else:
             # unless there are errors
+            # no sec because this use of mark_safe does not introduce a cross-site
+            # scripting vulnerability because there is no untrusted content inside.
+            # It is only being used to pass a specific HTML entity into a template.
+            messages.warning(
+                request,
+                mark_safe(  # nosec
+                    "<b>We could not save all the fields.</b><br/> The highlight "
```
neilmb, line 330:
```suggestion
                    "<b>We could not save all the fields.</b><br/> The highlighted "
```

src/registrar/templates/includes/form_errors.html
```diff
@@ -1,5 +1,9 @@
+{% comment %}
```
neilmb, line 1: **question**: @igorkorenfeld Can you confirm that we don't want the individual field errors to appear at the top of the form page any longer? I think that we have heard from some of our user interviews that people find it helpful to know at the top of the page what they need to fix. As it is, this PR looks like:

Screen Shot 2023-02-17 at 12 11 31

src/registrar/views/application.py
```diff
@@ -319,13 +321,26 @@ def post(self, request, *args, **kwargs):
...
+                ),
+            )
             context = self.get_context_data()
             context["forms"] = forms
             return render(request, self.template_name, context)
 
         # if user opted to save their progress,
         # return them to the page they were already on
         if button == "save":
+            messages.success(request, "Your progress has been saved!")
```
neilmb, line 343: **comment**: This is so nice to see after clicking Save!

Screen Shot 2023-02-17 at 12 13 20

neilmb (approved):
Just a few typos and questions, but otherwise it looks good to me.

src/registrar/views/application.py
kewlguy781: This is correct, I plan to work on #359 next. If we cannot complete that ticket for any reason, we will need to revisit and reword this.

src/registrar/templates/includes/form_messages.html
kewlguy781: Right now my understanding of base messages is for debug only. My main concern if we do the above is, supposedly we add more tags to the extra_tags, that will lead to issues because of how I use the tag (success/caution) for CSS/USWDS styling. Something I plan to keep eye on during the project.

src/registrar/templates/includes/form_errors.html
igorkorenfeld: I was hoping to have two behaviors:
1. On "Save progress" - show warning,  but don't list out the errors.
2. On "Save and continue" - show error and list out the errors. 

I think "Save progress" should has a softer warnings because the user isn't trying to advance yet. But on "Save and continue" we do want to be more detailed so that the user can fix the errors before they advance. 

Does that make sense?

kewlguy781: This will require new ticket. It requires more logical in coding.

PR #434, opened 2023-02-21
Add confirmation and caution message to the forms

kewlguy781:
This is same PR of #414, just with cleaner branch with all changes "signed" (Github wasnt playing nice with me with the older branch because it had my personal email which caused github to "unsigned" the first two changes which was then corrected in the commits after these two. However I could not rebase the first two commit in that PR, and I fear making any further changes/modify to the branch will make problem than benefit.

## 🗣 Description ##

Complete #398 
Based on PR #414 

- Added messages to forms
- Added confirmation message when saved progress
- Changed the error message in form to caution
- Commented out the error message on the top of form
- Changed the message in base to only show if it have extra_tag base

Closes ticket https://github.com/cisagov/getgov/issues/398. Also did caution message which is already shown according to UX design (we already have ticket (359) for caution but that ticket focus on be able to save partial fields instead of the UI)

### DIFF
--- a/src/registrar/templates/application_form.html
+++ b/src/registrar/templates/application_form.html
@@ -18,6 +18,10 @@
         </a>
         {% endif %}
 
+{% block form_messages %}
+          {% include "includes/form_messages.html" %}
+{% endblock %}
+
 {% block form_errors %}
         {% comment %}
            to make sense of this loop, consider that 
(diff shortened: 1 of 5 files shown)


loganmeetsworld (approved):
Approving based on comments on the other PR!

PR #439, opened 2023-02-27
Mr/legal

michelle-rago:
# Text updates based on user feedback and latest decisions #

## 🗣 Description ##

- Major change: Removed AO and naming guidance from legal
- Updated some form field labels and section intro text

### DIFF
--- a/src/registrar/forms/application_wizard.py
+++ b/src/registrar/forms/application_wizard.py
@@ -672,11 +672,11 @@ OtherContactsFormSet = forms.formset_factory(
 
 class NoOtherContactsForm(RegistrarForm):
     no_other_contacts_rationale = forms.CharField(
-        required=False,
+        required=True,
         # label has to end in a space to get the label_suffix to show
         label=(
-            "If you can’t provide other contacts for your organization,"
-            " please explain why."
+            "Please explain why there are no other employees from your organization"
+            " that we can contact."
         ),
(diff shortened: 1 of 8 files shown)


src/registrar/forms/application_wizard.py
```diff
@@ -180,10 +180,10 @@ def clean(self):
             or self.cleaned_data["state_recognized_tribe"]
         ):
             raise forms.ValidationError(
-                "Only tribes recognized by the U.S. federal government or by a U.S."
-                " state government are eligible for .gov domains. Please email"
-                " registrar@dotgov.gov to tell us more about your tribe and why you"
-                " want a .gov domain.",
+                "Only tribes recognized by the U.S. federal government
+                " or by a U.S. state government are eligible for .gov
+                " domains. Please <a href="{% url 'todo' %}">tell us more about your tribe and why
```
neilmb, line 185: **comment (non-blocking)**: This might not look right on the page because Django might escape the HTML codes from the error message specified here. `mark_safe` is how we've dealt with this in other similar situations and I'm happy to help work it out.

@michelle-rago How does this look in the sandbox?

michelle-rago: It didn't deploy to my sandbox ☹️

neilmb (approved):
Language changes are wonderful, but we might need some more technical work to get the link in an error message working.

Issue #448, opened 2023-03-05
Remove caution box from pages with errors


loganmeetsworld:
We've decided to remove the caution box with copy about partial saves and revert to previous error messaging. In the future we want to also: 

* Remove the “Save progress” button from the form
* Pages which registrant has completed (including blanks for optional questions) will have a check icon in the left nav, similar to the lock icon
* When a registrant has incomplete fields or invalid data, clicking on “Save and continue” will bring them to the same page with an error message box at the top, and the erroneous fields presented in the error state (i.e red border on input field, red left bar, error text). We may modify the error message text at the top of the page to indicate that none of the fields on this page have been saved

But these do not have to be accomplished as part of this task.

kewlguy781:
Done merge this into main.

PR #449, opened 2023-03-06
Caution Box removed. 

kewlguy781:
# Reverting changes of caution box

## 🗣 Description ##

Reverting change of caution box, as well as adding back in error message on the top of the message. 

closes ticket #448 :D


neilmb (approved):
LGTM, does what it says on the tin.

Issue #450, opened 2023-03-06
Remove "save progress" button at bottom of form


loganmeetsworld:
Follow up to #448, we want to remove the "Save progress" button on the form so that we only have one button to progress. 

Reviewer: @igorkorenfeld

kewlguy781:
I ran into some concerns with the test suite. How its written is that once it hit the "save progress" it does the data validation. I do not want to scrap them and felt that data validation should be in. I am going to talk with @SSPJ about it and see if there are any way we can do data validation without save progress feature. 

cc: @loganmeetsworld

PR #452, opened 2023-03-06
Save Progress button Removed

kewlguy781:
# Bye, bye Save Progress button!

## 🗣 Description ##

This remove the save progress, as well as making some changes to the test cases. In test cases execution, the data validation happens after "Save Progress" were executed. In that case, I have moved the data validation to after "Save and Continute" instead of removing them along with "Save Progress" code. This is why you see lot of changes in the test cases part.

* Save Progress button removed (#450)
* Redid test cases to do data validation after each step instead of within step

### DIFF
--- a/src/registrar/templates/application_form.html
+++ b/src/registrar/templates/application_form.html
@@ -70,18 +70,6 @@
               class="usa-button"
             >Submit your domain request</button>
             {% endif %}
-            {% comment %}
-            Commenting this section as we are "removing"
-            the save progress button for now. 
-            If confirmed to remove this, please reivist and
-            remove in application.py "if button == "save":"
-            <button 
-              type="submit"
-              name="submit_button"
-              value="save"
-              class="usa-button usa-button--outline"
-            >Save progress</button>
-            {% endcomment %}
           </div>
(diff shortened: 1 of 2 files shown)


src/registrar/templates/application_form.html
```diff
@@ -70,12 +70,18 @@ <h1> {{form_titles|get_item:steps.current}} </h1>
               class="usa-button"
             >Submit your domain request</button>
             {% endif %}
+            {% comment %}
```
neilmb, line 73: **Suggestion**: In order to keep our templates simple, I would like to see this code removed rather than commented out. If we want to use this again in the future, we have Git history to get it back.

neilmb (changes requested):
LGTM. Just one suggestion, to delete rather than comment out code that we aren't using anymore.

src/registrar/templates/application_form.html
kewlguy781: Done. 👍

neilmb (approved):
Perfect, thanks for the change.

kewlguy781:
this resolves #450

PR #459, opened 2023-03-10
Role-based Access Control

### LINKED ISSUE #297: Initialize RBAC for User model and add permissions decorator to user actions
_Outcome_: We have a permissions model with only one role (domain `Manager`) and can use a decorator on user actions.

neilmb:
# Role-based Access Control Research #

## 🗣 Description ##

This branch sets up an experiment with an access-control system. It isn't complete and ready to be used as-is, but it is intended to lay the groundwork for an MVP system. It introduces a data model for users to have a role on a particular domain and then uses that data model to control access to a domain's detail page. On their dashboard, users will see only domains on which they have some role.

When a logged-in user visits the detail page for a domain that they don't have access to, they get a (completely unstyled) 403 Forbidden error. When a user who hasn't logged in visits a detail page, they are redirected to login.gov.

domain_permissions mov

**Limitations:** There is currently only one role called `ADMIN`. Since there's only one role, I didn't build a complex decorator for allowing particular roles to do particular things. The domain detail page is _extremely_ basic.

Closes #297


SSPJ (approved):
Haven't finished, just wanted to get my tentative 👍 over to you before EOD.

This looks really great so far! Thank you for taking on this ticket, Neil!

The Basics
- [x] I read the code
- ❌  I ran the code -- will do this ASAP

Documentation
- ❌  changes to “how we do things” are documented in READMEs
    This feels significant enough that I'd like to see a bit of developer documentation (no more than a paragraph or two). Ideally, this would the plain language answers to common edits that future maintainers might like to make, such as "How do I add another role?" ... "How do I give users read-only access to X?" ... "How/where do users get assigned roles?"

Security

Reliability
- ❌  error handling exists for unusual or missing values -- still testing this one
- [x] functionality is tested with unit or integration tests 🎉

src/registrar/views/index.py
```diff
@@ -9,4 +10,13 @@ def index(request):
...
         context["domain_applications"] = applications
+
+        domains = request.user.permissions.values(
+            "role",
+            pk=F("domain__id"),
+            name=F("domain__name"),
+            created_time=F("domain__created_at"),
+            application_status=F("domain__domain_application__status"),
+        )
+        context["domains"] = domains
```
SSPJ, line 21: OKAY, the comment I left on this code which was more erudite before GitHub erased it for me was that I'd love to see this done in a less opt-in way, possibly with models.Manager or models.QuerySet.as_manager(), such that calling `Domain.objects.all()` (or the like) would return `[]` unless you were a superuser. BUT that this is fine as-is if you don't have time for that right now.

src/registrar/models/domain_application.py
```diff
@@ -502,6 +502,25 @@ def submit(self):
         # This is a side-effect of the state transition
         self._send_confirmation_email()
 
+    @transition(field="status", source=[SUBMITTED, INVESTIGATING], target=APPROVED)
+    def approve(self):
+        """Approve an application that has been submitted.
```
SSPJ, line 507: This doesn't appear to have run when I changed the state of a domain application from the admin dashboard. While trying to debug this, I've realized that we need to add the new UserDomainRole to the admin, because otherwise I can't see whether I've been added or not.

neilmb: We used `protected=False` when we defined this FSMField so the admin is able to change the status field without going through the state transition methods. It seems now like this might not be what we want? There is also https://github.com/gadventures/django-fsm-admin which is an FSM-aware module for the Django admin that I think will run the transition methods when the state is changed in the admin.

Issue #477, opened 2023-03-20
Set Up Functionality to Authenticate and Assign User to an Accepted Domain


calerubensteingsa:
**Acceptance Criteria**

- Add functionality to User Management so that if a user is added to an accepted domain a email is generated prompting them to set up a login.gov account and then be assigned to the domain

Doc: https://docs.google.com/document/d/1vmVKmxQ7G4uJ7S6c27RNpTDKbAjCxt3PQ-qgCb5Uhq4/edit?usp=sharing

See thread: https://gsa-tts.slack.com/archives/C03QM0JGSQG/p1679344501659889

neilmb:
Django models and views are now done in the domain's user management along with their unit tests on https://github.com/cisagov/getgov/tree/nmb/user-invitations

Still need to make emails be sent when we create invitations, and "retrieve" existing invitations when a user logs in.

PR #480, opened 2023-03-21
Simple user management

### LINKED ISSUE #461: Existing Login - Create Happy Path for Adding a User to an Accepted Domain
**Acceptance Criteria**

- Develop a happy path from:

1. Domain is accepted
2. User clicks on managing a domain
3. User is able to add other users with a login on get.gov to domain

See: #460 for comp

neilmb:
# Simple user management #

## 🗣 Description ##

This adds the initial UI for adding users to an approved domain. Using our previous RBAC work, we add a `/domain/<id>/users` view to show the users on the domain and a `/domain/<id>/users/add` form where people can enter an email address to add a user with the "admin" role on that domain. (There is no other role than "admin" right now.)

Screen Shot 2023-03-21 at 2 55 06 PM

Currently, users can only add emails that have already been seen as users in our registrar system. In #477 we will extend this functionality to send emails to users who haven't logged in before to invite them to log in. Another limitation is that styling does not yet exactly match what Igor has drawn in Figma, primarily in the user table and breadcrumbs.

Closes #461

### DIFF
--- a/src/registrar/templates/domain_base.html
+++ b/src/registrar/templates/domain_base.html
@@ -12,12 +12,22 @@
     <div class="grid-col-9">
       <main id="main-content" class="grid-container">
 
-	  {% block domain_content %}
-	  <h1>Domain {{ domain.name }}</h1>
+      {% if messages %}
+      {% for message in messages %}
+      <div class="usa-alert usa-alert--{{ message.tags }} usa-alert--slim margin-bottom-2">
+          <div class="usa-alert__body">
+          {{ message }}
+          </div>
+      </div>
+      {% endfor %}
+      {% endif %}
 
-	  {% endblock %}  {# domain_content #}
+      {% block domain_content %}
+      <h1>Domain {{ domain.name }}</h1>
 
-	  </main>
+      {% endblock %}  {# domain_content #}
+
+      </main>
     </div>
   </div>
 </div>
--- a/src/registrar/templates/domain_base.html
+++ b/src/registrar/templates/domain_base.html
@@ -1,0 +1,24 @@
+{% extends "dashboard_base.html" %}
+
+{% block title %}Domain {{ domain.name }}{% endblock %}
+
+{% block content %}
+<div class="grid-container">
+  <div class="grid-row grid-gap">
+    <div class="grid-col-3">
+      {% include 'domain_sidebar.html' %}
+    </div>
+
+    <div class="grid-col-9">
+      <main id="main-content" class="grid-container">
+
+	  <h1>Domain {{ domain.name }}</h1>
+
+	  {% block domain_content %}
+	  {% endblock %}  {# domain_content #}
+
+	  </main>
+    </div>
+  </div>
+</div>
+{% endblock %}  {# content #}
--- a/src/registrar/templates/domain_base.html
+++ b/src/registrar/templates/domain_base.html
@@ -1,4 +1,4 @@
-{% extends "dashboard_base.html" %}
+{% extends "base.html" %}
 
 {% block title %}Domain {{ domain.name }}{% endblock %}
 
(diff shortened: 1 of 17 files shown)


src/registrar/templates/domain_base.html
```diff
@@ -0,0 +1,42 @@
...
+    <div class="grid-col-3">
+      {% include 'domain_sidebar.html' %}
+    </div>
+
+    <div class="grid-col-9">
+      <main id="main-content" class="grid-container">
+
+      {% if messages %}
+      {% for message in messages %}
+      <div class="usa-alert usa-alert--{{ message.tags }} usa-alert--slim margin-bottom-2">
```
SSPJ, line 18: It continues to be a slight annoyance that blind users will not get notification of the styling. (A red warning message will "look" the same as a green success message when read by the screen reader.)

I think it's fine; we've just got to be careful when crafting our messages that the meaning and severity is completely conveyed by the text itself. CC @michelle-rago

src/registrar/views/domain.py
```diff
@@ -11,3 +17,60 @@ class DomainView(DomainPermission, DetailView):
...
     context_object_name = "domain"
+
+
+class DomainUsersView(DomainPermission, DetailView):
+    model = Domain
+    template_name = "domain_users.html"
+    context_object_name = "domain"
+
+
+class DomainAddUserForm(DomainPermission, forms.Form):
```
SSPJ, line 28: **Non-blocking** Consider scooting this over into `registrar/forms`.

src/registrar/views/domain.py
```diff
@@ -11,3 +17,60 @@ class DomainView(DomainPermission, DetailView):
...
+            return self.form_valid(form)
+        else:
+            return self.form_invalid(form)
+
+    def form_valid(self, form):
+        """Add the specified user on this domain."""
+        requested_email = form.cleaned_data["email"]
+        # look up a user with that email
+        # they should exist because we checked in clean_email
+        requested_user = User.objects.get(email=requested_email)
```
SSPJ, line 65: Personally philosophy: if it is worth writing a note to explain how an error _couldn't_ occur, it is worth writing a try/catch block for that error. 👼

As a more practical note, I think once the email sending is implemented, that clean function can maybe be deleted, since the email will be sent from here, once the form is otherwise fully valid and submitted.

src/registrar/templates/domain_sidebar.html
```diff
@@ -0,0 +1,61 @@
...
+        {% url 'domain' pk=domain.id as url %} 
+        <a href="{{ url }}" 
+          {% if request.path == url %}class="usa-current"{% endif %}
+        >
+          Domain Overview
+        </a>
+      </li>
+
+      <li class="usa-sidenav__item">
+        {% url 'todo' as url %}
```
SSPJ, line 16: Clever. 💯

SSPJ (approved):
Looks great, Neil! 🥇 Thanks for adding this feature.

Lots of comments, but no major issues. Merge when ready!

The Basics
- [x] I read the code
- [x] I ran the code

Documentation
- ❌  all new functions and methods are commented using plain language

Security
- [x] data from external sources is cleaned and clearly marked 💯 

Reliability
- [x] error handling exists for unusual or missing values -- ❗ see review comment

Accessibility
    - [x] Tab order is logical

Device Matrix
    - [x] at 480px -- ❗ overall, yes, but long domain names are going to be an issue
        image
    - [x] at 1024px

src/registrar/templates/domain_base.html
SSPJ: Unsure. We aren't making use of a messages file, so I think they are currently scattered around the codebase. Essentially, the errors are written wherever they are raised.

Use of a messages file to contain them would be a good idea, but I think our reasoning for not doing so was that we don't have any website translation planned.

neilmb: @michelle-rago These messages are produced in the code by `messages.success`, `messages.warning`, etc. in the appropriate flow in the code. See e.g. https://github.com/cisagov/getgov/blob/nmb/approved-users/src/registrar/views/domain.py#L75

src/registrar/views/domain.py
neilmb: This is about to change in follow-on work, but I agree with the sentiment and it will show up a couple times in that follow-on work.

PR #484, opened 2023-03-23
Domain name and back link

igorkorenfeld:
## 🗣 Description ##

This PR adds the domain name above the main content on the page so users always see which domain they are working on. 

Also moves the back link to the the base template and adds styling to match the form. 
image


neilmb (approved):
LGTM, thank you!

PR #485, opened 2023-03-27
Fix deprecation warnings

neilmb:
# Fix deprecation warnings #

## 🗣 Description ##

In preparation for the upcoming Django 4.2, their docs encourage fixing deprecation warnings. Running our test suite with Python's warnings turned on (`docker compose run app python -Wa manage.py test`) produces many warnings that can fairly trivially be fixed. This PR fixes all of them. (Actually, there is one left: `RemovedInDjango50Warning: The USE_L10N setting is deprecated.` and I can't find any guidance on how to satisfy that deprecation warning. We appear to be correctly using `USE_L10N = True` in our project's settings file.)

Good hygiene, I guess?

### DIFF
--- a/src/api/tests/common.py
+++ b/src/api/tests/common.py
@@ -47,3 +47,5 @@ def less_console_noise():
         # restore the streams
         for handler in handlers.values():
             handler.setStream(restore[handler.name])
+        # close the file we opened
+        devnull.close()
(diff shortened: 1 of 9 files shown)


SSPJ (approved):
LGTM! 👍 Thanks for catching those file handle leaks!

To check the new form renderer, I tested only a representative sampling of pages, not every form page in the app.

The Basics
- [x] I read the code
- [x] I ran the code

Documentation

Security

Reliability

Accessibility
    - [x] Tab order is logical
    - [x] Links are unique and contextual. All links can be understood taken alone, e.g., ‘Read more - about 508’ -- ❗ Axe did flag this one for the domain management dashboard, but Neil had already tested with voice input control and there are no problems.

Device Matrix
    - [x] at 480px
    - [x] at 1024px

PR #488, opened 2023-03-28
remove temporary landing page

### LINKED ISSUE #486: Remove landing sign in page on Django app
We will remove the landing page with the sign in button on the django app

A/C: All users get a page where they can either log in or start their login account creation

loganmeetsworld:
# What
Fixes #486 

# Side quest
The reset_db script was broken due to a typo so fixed that. I've also noticed the cloud.gov github action doesn't like `--wait` anymore (a new CF version thing?) so I've removed that and migrate worked for my environment.

### DIFF
--- a/src/djangooidc/views.py
+++ b/src/djangooidc/views.py
@@ -109,7 +109,9 @@ def logout(request, next_page=None):
         # Don't wait for the callback as it may never come.
         auth_logout(request)
         logger.info("Successfully logged out user %s" % username)
-        return HttpResponseRedirect(redirect_to="https://get.gov")
+        next_page = getattr(settings, "LOGOUT_REDIRECT_URL", None)
+        if next_page:
+            request.session["next"] = next_page
 
 
 def logout_callback(request):
--- a/src/djangooidc/views.py
+++ b/src/djangooidc/views.py
@@ -109,12 +109,10 @@ def logout(request, next_page=None):
         # Don't wait for the callback as it may never come.
         auth_logout(request)
         logger.info("Successfully logged out user %s" % username)
-        next_page = getattr(settings, "LOGOUT_REDIRECT_URL", None)
-        if next_page:
-            request.session["next"] = next_page
+        return HttpResponseRedirect(redirect_to="https://get.gov")
 
 
 def logout_callback(request):
     """Simple redirection view: after logout, redirect to `next`."""
-    next = request.session.get("next", "/")
+    next = request.session.get("next", "https://get.gov")
     return redirect(next)
--- a/src/registrar/tests/test_views.py
+++ b/src/registrar/tests/test_views.py
@@ -35,7 +35,7 @@ class TestViews(TestCase):
     def test_home_page(self):
         """Home page should not be available without a login."""
         response = self.client.get("/")
-        self.assertContains(response, "registrar", status_code=302)
+        self.assertEqual(response.status_code, 302)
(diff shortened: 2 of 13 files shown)


src/registrar/tests/test_views.py
```diff
@@ -24,10 +24,10 @@ def test_health_check_endpoint(self):
         self.assertContains(response, "OK", status_code=200)
 
     def test_home_page(self):
-        """Home page should be available without a login."""
+        """Home page should not be available without a login."""
```
neilmb, line 27: Love a test that embodies our intent.

src/registrar/views/index.py
```diff
@@ -1,9 +1,11 @@
 from django.db.models import F
 from django.shortcuts import render
+from django.contrib.auth.decorators import login_required
 
 from registrar.models import DomainApplication
 
 
+@login_required
 def index(request):
     """This page is available to anyone without logging in."""
```
neilmb, line 10: **suggestion**: We should change this docstring to accurately match the new decorator.

neilmb (approved):
LGTM, just one docstring suggestion.

src/zap.conf
```diff
@@ -78,6 +78,8 @@
 10062	FAIL	(PII Disclosure - Passive/beta)
 10095	FAIL	(Backup File Disclosure - Active/beta)
 10096	FAIL	(Timestamp Disclosure - Passive/release)
+# We have timestamps on the main page for sorting applications
+10096	OUTOFSCOPE	http://app:8080.*
```
loganmeetsworld, line 82: @neilmb confirming this works for you

src/djangooidc/views.py
```diff
@@ -109,12 +109,10 @@ def logout(request, next_page=None):
         # Don't wait for the callback as it may never come.
         auth_logout(request)
         logger.info("Successfully logged out user %s" % username)
-        next_page = getattr(settings, "LOGOUT_REDIRECT_URL", None)
-        if next_page:
-            request.session["next"] = next_page
+        return HttpResponseRedirect(redirect_to="https://get.gov")
```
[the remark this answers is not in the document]
SSPJ: The subtle behavior here is that a return statement in a finally block will preempt a return statement in the try clause. In other words, this will replace `return HttpResponseRedirect(url)` with `return HttpResponseRedirect(redirect_to="https://get.gov")` (skipping Login.gov entirely).

If you'd like folks to end up at "https://get.gov" after logging out, there's no real need for them to even return to the Django site at all. Ensure that the partner dashboard of Login.gov is configured to allow get.gov as a post-logout redirect destination and then send get.gov as the value of the GET parameter `next` in the logout request.

The prior code ought to work, if you set `LOGOUT_REDIRECT_URL` to what you'd like, although I haven't tested this.

SSPJ:
@loganmeetsworld I found some pages that were missing their login required decorator. Since we are moving away from having any non-public content on the Django site, it makes sense to me to protect the whole app (with something like [this login middleware](https://stackoverflow.com/questions/8377086/set-all-pages-to-require-login-globally)).

This could easily be its own ticket, but could also be a bolt on to this PR. Happy to help or pair, if you are interested. Also happy to pair on the OWASP issue, although I don't have as much context as Neil so I might not be very helpful.

loganmeetsworld:
We are going to supercede this with a couple changes to how we are doing authentication generally. I will reference this in a new ticket related to those tasks.

PR #491, opened 2023-03-30
Update review page style 

igorkorenfeld:
## 🗣 Description ##

This PR adds a graphic to the review page title and makes the submit button larger and greener :) 
alt text is intentionally empty because the image is decorative. 

image
- - -
image

The goal is to differentiate this page a bit more from the other form pages and give the last step a little more weight. These changes are in response some of the feedback we got during our user research sessions.


src/registrar/templates/application_review.html
```diff
@@ -5,6 +5,18 @@
...
 
+{% block form_page_title %}
+  <span class="display-flex flex-align-start" > 
+    <img 
+        class= "margin-right-105" 
+        src="{%static 'img/registrar/dotgov_review_magnify.svg' %}" 
+        alt=""
+        width="72"
+        />
+    <h1> Review and submit your domain&nbsp;request </h1>
```
SSPJ, line 16: The non-breaking space is causing the layout to look a bit wonky on smaller screens:

image

SSPJ (approved):
LGTM, Igor! 👍 

The Basics
- [x] I read the code
- [x] I ran the code

Documentation

Security

Reliability

Accessibility
    - [x] Tab order is logical

Device Matrix
    - ❌  at 480px
    - [x] at 1024px

src/registrar/templates/application_review.html
igorkorenfeld: Thanks Seamus! Tried a different approach with a title instead

PR #495, opened 2023-03-31
Begin using requirements.txt

SSPJ:
What it says on the tin.

> The requirements.txt is used by Cloud.gov. It is needed to work around a bug in the CloudFoundry buildpack version of Pipenv that breaks on installing [packages] from a git repository.

### DIFF
--- a/src/requirements.txt
+++ b/src/requirements.txt
@@ -1,0 +1,49 @@
+-i https://pypi.python.org/simple
+asgiref==3.6.0 ; python_version >= '3.7'
+boto3==1.26.69
+botocore==1.29.69 ; python_version >= '3.7'
+cachetools==5.3.0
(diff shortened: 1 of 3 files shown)


loganmeetsworld (approved):
Looks great! Thanks for updating documentation.

src/requirements.txt
neilmb, line 1: **question**: I don't see a git URL in here that necessitated adding the requirements.txt. Does that come in a separate PR?

neilmb (approved):
LGTM.

src/requirements.txt
SSPJ: Yes. I'm currently working from the CISA fork in my sandbox and I don't want to deploy (possibly broken) changes to other environments.

PR #514, opened 2023-04-10
Improve request received email

### LINKED ISSUE #416: Update Confirmation Email for Submitted Domain

neilmb:
# Improve request received email #

## 🗣 Description ##

This improves the plain text email that we send to registrants when they submit a new domain application. It includes a summary of their application along with an explanation of next steps.

Screen Shot 2023-04-10 at 1 42 49 PM

Formatting plain text documents using Django's template system can be tricky as it was designed for HTML which is whitespace-insensitive. This PR uses Django's `{% spaceless %}` tag along with tests that explicitly check how many newlines end up between items.

Closes #416

### DIFF
--- a/src/registrar/models/domain_application.py
+++ b/src/registrar/models/domain_application.py
@@ -478,7 +478,7 @@ class DomainApplication(TimeStampedModel):
                 "emails/submission_confirmation.txt",
                 "emails/submission_confirmation_subject.txt",
                 self.submitter.email,
-                context={"id": self.id, "domain_name": self.requested_domain.name},
+                context={"application": self},
             )
         except EmailSendingError:
             logger.warning("Failed to send confirmation email", exc_info=True)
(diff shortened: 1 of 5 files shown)


kewlguy781 (approved):
Codewise, Look good to me. The only question I have about this is, do we need to have "do not reply to this email as it is done automated" text in the email. Something to consider @michelle-rago @igorkorenfeld

michelle-rago:
> Codewise, Look good to me. The only question I have about this is, do we need to have "do not reply to this email as it is done automated" text in the email. Something to consider @michelle-rago @igorkorenfeld

Good question. We should check with @h-m-f-t on whether or not this email should receive replies.

michelle-rago (commented):
Looking good, @neilmb! I'm seeing some font color differences. Is that just something happening here or is it intentional?

h-m-f-t:
> > Codewise, Look good to me. The only question I have about this is, do we need to have "do not reply to this email as it is done automated" text in the email. Something to consider @michelle-rago @igorkorenfeld
> 
> Good question. We should check with @h-m-f-t on whether or not this email should receive replies.

We chatted about this question a bit at https://cisa-corp.slack.com/archives/C03QM0JGSQG/p1679670277935509. We should plan for a Reply-To: header, and the address should probably be dotgov@cisa.dhs.gov for now (eventually a help desk system).

Issue #328, opened 2022-12-13
Manage Submitted Domain - Implement Withdraw Domain Functionality


calerubensteingsa:
**Acceptance Criteria**

see below

igorkorenfeld:
Design for the [confirmation page ](https://www.figma.com/proto/ninNjo0bTg7XfB4mvTp4hG/GetGov%2FDesigns?page-id=3124%3A32294&node-id=3194-35153&viewport=158%2C512%2C0.17&scaling=min-zoom&starting-point-node-id=3194%3A35488)after "Withdraw request" is pressed. 

[Artboard](https://www.figma.com/file/ninNjo0bTg7XfB4mvTp4hG/GetGov%2FDesigns?node-id=3194%3A35153&t=We0rj9wgKCR2m5Gf-1)

Figma [prototype of the flow](https://www.figma.com/proto/ninNjo0bTg7XfB4mvTp4hG/GetGov%2FDesigns?page-id=3124%3A32294&node-id=3194-35488&viewport=158%2C512%2C0.17&scaling=min-zoom&starting-point-node-id=3194%3A35488) to help make it clearer:
1. Click "Manage" for the submitted domain
2. Click the "Withdraw request" button
3. Click "Withdraw request" on the confirmation page.

igorkorenfeld:
Updated the date to be "Last updated"

loganmeetsworld:
https://github.com/cisagov/getgov/pull/525

PR #525, opened 2023-04-13
New State: Withdrawn / Withdraw Confirmation Page

kewlguy781:
New State: Withdrawn for these who withdraw the application that are in either "Submitted" and "Investigating" State only

## 🗣 Description ##

Added State: `WITHDRAWN`
Added Page: Withdraw Confirmation Page
Updated Application Status to have link to Withdraw Confirmation Page
Added Unit Tests for Withdraw
If withdrawn, Will display "Edit" instead of "Manage" under "Domain Application Summary" 
Added State Translation of Withdrawn to Submitted (so people can open and edit and submitted)

this resolves and closes ticket #328.
I have to add a transition for withdrawn to be able to accepted and be re-submitted.


src/registrar/views/application.py
```diff
@@ -483,3 +483,18 @@ class ApplicationStatus(generic.DetailView):
...
+
+    def get_context_data(self, **kwargs):
+        context = super(ApplicationWithdraw, self).get_context_data(**kwargs)
+        return context
+
+    def updatestatus(request, pk):
+        application = DomainApplication.objects.get(id=pk)
+        application.status = "withdrawn"
+        application.save()
+        return HttpResponseRedirect(reverse("home"))
```
SSPJ, line 500: Should we use Django messages to show a success message to the user?

src/registrar/templates/application_withdraw_confirmation.html
```diff
@@ -0,0 +1,20 @@
+{% extends 'base.html' %}
+
+{% block title %}Withdraw request for {{ domainapplication.requested_domain.name }}{% endblock %}
+{% load static url_helpers %}
+
+{% block content %}
+<div class="grid-col desktop:grid-offset-2 desktop:grid-col-8">
```
igorkorenfeld, line 7: For alignment and spacing, this whole block should be wrapped in `<div class="grid-container">`

To clarify the container div would start right after `{% block content %}` and end right before `{% endblock %}`

This should resolve the padding issue.

src/registrar/templates/application_withdraw_confirmation.html
```diff
@@ -0,0 +1,20 @@
...
+ 
+
+<p><h1>Withdraw request for {{ domainapplication.requested_domain.name }}?</h1></p>
```
igorkorenfeld, line 10: Don't need to wrap the `h1` in a `p` tag (causing some typographic misalignment)

src/registrar/templates/application_withdraw_confirmation.html
```diff
@@ -0,0 +1,20 @@
...
+
+If you withdraw your request we won't review it. Once you withdraw your request you'll be able to edit it or completely remove it. 
+
+<p><a href="{% url 'application-withdrawn' domainapplication.id %}" class="usa-button withdraw">Withdraw request</button>
```
igorkorenfeld, line 14: The space the two buttons is a little tight, and I think also causing the no-CSS issue Seamus noted.  e could either have them as just anchor tags, or as individual `p` tags, that are then grouped together and styled with `flex` or `inline-blocks`

src/registrar/assets/sass/_theme/_uswds-theme-custom-styles.scss
```diff
@@ -133,11 +133,20 @@ a.breadcrumb__back {
...
 
+a.withdraw_outline:visited {
+  box-shadow: inset 0 0 0 2px color('error');
+  color: color('error');
+}
+
+a.withdraw {
+  background-color: color('error');
+}
+
```
igorkorenfeld, line 149: We need to add a couple of more rules for the `hover`, `focus`, and `active` states, because otherwise it's using the `usa-button` versions, which are the wrong hue. This was underspecified on my part, sorry about that.

On `.withdraw_outline`, could you add for `:hover` and `:focus` :
`color: color('error-dark');`

and for `:active`:
`color: color('error-darker');`

Similarly  for `withdraw` for or `:hover` and `:focus` :
`background-color: color('error-dark');`

and for `:active:`
`background-color: color('error-darker');`

igorkorenfeld (changes requested):
Looking good, thanks for your work on this. I noticed a couple of styles changes that need to be adjusted a bit.

kewlguy781:
Re: unit test, I thought I have had this push since I have been using this test in the last 5 commits. so I re-push this into the last commit. You should see the unit tests now.

src/registrar/config/urls.py
```diff
@@ -57,6 +57,16 @@
...
+    path(
+        "application/<int:pk>/withdraw",
+        views.ApplicationWithdraw.as_view(),
+        name="application-withdraw-confirmation",
+    ),
+    path(
+        "application/<int:pk>/withdrawconfirmed",
+        views.ApplicationWithdraw.updatestatus,
+        name="application-withdrawn",
+    ),
```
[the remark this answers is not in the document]
kewlguy781: I looked at .pa11y file and I see some problem.
1) There are no examples of it using the primary key in order to pull out respective information
2) this page is now protected by `DomainPermission` How does that for the program to access and view to work?

src/registrar/templates/application_withdraw_confirmation.html
kewlguy781: fixed.

src/registrar/templates/application_withdraw_confirmation.html
kewlguy781: i found a problem, i fixed it.

src/registrar/views/application.py
```diff
@@ -481,5 +483,24 @@ class ApplicationStatus(generic.DetailView):
...
+        """Get context details to process information from application"""
         context = super(ApplicationStatus, self).get_context_data(**kwargs)
         return context
+
+
+class ApplicationWithdraw(generic.DetailView, DomainPermission):
+    model = DomainApplication
+    template_name = "application_withdraw_confirmation.html"
+    """ The page above will display asking user to confirm if they want to withdraw;
+
```
SSPJ, line 495: @kewlguy781 I went to retest this for the login issue. I cannot withdraw as a logged out user 👏 but I am able to see the content on both https://getgov-jon.app.cloud.gov/application/3 and https://getgov-jon.app.cloud.gov/application/3/withdraw. This might be a bug in `DomainPermission` itself, or it might be that additional mixins are needed.

SSPJ (changes requested):

src/registrar/views/application.py
kewlguy781: this ticket focus only on the withdraw, the ticket for the application part is already merged into main. I will make change under different branch.

src/registrar/templates/application_withdraw_confirmation.html
igorkorenfeld: This one need to be tweaked just slightly — the news div does need to be inside the `{% block content %}`  (both the start and end of it) otherwise things getting a little off due to the wrapper.

src/registrar/templates/application_withdraw_confirmation.html
igorkorenfeld: Might be the committing issue you mentioned in slack, but this change still doesn't show up for me – i.e. the `h1` is still in a `p` tag

src/registrar/templates/application_withdraw_confirmation.html
```diff
@@ -0,0 +1,22 @@
...
+{% load static url_helpers %}
+
+<div class="grid-container">
+{% block content %}
+<div class="grid-col desktop:grid-offset-2 desktop:grid-col-8">
+ 
+
+<p><h1>Withdraw request for {{ domainapplication.requested_domain.name }}?</h1></p>
+
+If you withdraw your request we won't review it. Once you withdraw your request you'll be able to edit it or completely remove it.
```
igorkorenfeld, line 13: Missed this before, but this line should be wrapped in a `p` tag

igorkorenfeld (changes requested):
Great work. So odd with the commits not showing up – a couple of issues to clear up because of that.

igorkorenfeld (approved):
Changes look good, thank you!

SSPJ (approved):
It looks like solving the issue of why logged out users can access the page will be a bit more complicated. I'm approving so this can get merged. Jon is creating #537 as follow on work.

@loganmeetsworld This perhaps dovtails with the comment I left on your PR, #488 .

PR #535, opened 2023-04-17
Fix Pa11y Github Action

### LINKED ISSUE #518: Automated accessibility tests aren't running and aren't failing
Our pa11y tests that run as part of Github actions have not been running, but the tests have also not been failing, so we didn't notice that they weren't running.

@igorkorenfeld has a fix in progress based on some things that other 18F teams have done.

igorkorenfeld:
## 🗣 Description ##

This PR updates our GH action for pa11y-ci, running pa11y outside the docker container. This gets pa11y to run again, as it has been quietly failing in our workflow (failing to run, but showing up as a checkmark in the workflow). 

A slight disadvantage of this approach is that there is potential for the environment the test is run in to be less consistent than it would have been inside of a docker container. The advantage of this approach is that we are able to actually run it without the puppeteer issues that are present when trying to run it from within the container.

A good indication that the tests are running properly is that we did get some errors showing up, prior to adding an ignore rule to the configuration file:

Screen Shot 2023-04-18 at 4 55 58 PM

Special thanks to the engineers over on the FAC project [documenting the problem](https://github.com/GSA-TTS/FAC/issues/711) in their issues and having a [config](https://github.com/GSA-TTS/FAC/pull/862) we can reference.

Closes #518

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -43,9 +43,15 @@ jobs:
         run: |
           perl -pi \
           -e 's/"csp.middleware.CSPMiddleware",/$&"registrar.tests.common.MockUserLogin",/' \
-          src/registrar/config/settings.py
+          /registrar/config/settings.py
 
-      - name: Accessibility Scan
+      - name: Start container
         working-directory: ./src
         # leverage the docker compose setup that we already have for local development
-        run: docker compose run pa11y npm run pa11y-ci
+        run: docker compose up -d 
+
+      - name: run pa11y
+        working-directory: ./src
+        run: |
+          npm i -g pa11y-ci
+          pa11y-ci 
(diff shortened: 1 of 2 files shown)


SSPJ:
Nice work, Igor! 🏆 

In order to test this, I will need to add some failures in a commit to your branch. I will remove that commit once done testing.

CC: @neilmb

SSPJ (approved):
LGTM! 👍 Thanks for figuring this out, Igor.

The Basics
- [x] I read the code
- [x] I ran the code

Documentation

Security
- [x] security false positives are documented -- I'm not certain that we want to keep an exception for the previously failing rules, but no need to address that on this PR

Reliability
- [x] functionality is tested with unit or integration tests 🥇 

---

errors proving test works

PR #538, opened 2023-04-20
Add hyphens for highest-*official

michelle-rago:
# We're using hyphens for highest-elected something or highest-ranked something #

## 🗣 Description ##

This style is documented in the .gov content guide (which is still being developed).


igorkorenfeld (approved):
Looks good!

PR #539, opened 2023-04-20
Style tuning for domain request status page

igorkorenfeld:
## 🗣 Description ##
This PR adjusts the styles on the status page for a submitted domain request – aligning the container and making sure the typography matches the review page. In order to not repeat class names for each data point in the application I have also set up an `include` so that the basic structure could be reused. 

Additional the PR includes a fix for an a11y error – a missing `main` for the skip link.


src/registrar/templates/application_status.html
```diff
@@ -1,114 +1,99 @@
 {% extends 'base.html' %}
 
-{% block title %}Domain request status- {{ domainapplication.requested_domain.name }}{% endblock %}
+{% block title %}Domain request status- {{ domainapplication.requested_domain.name }} | {% endblock %}
```
neilmb, line 3: **comment (non-blocking)**: For consistency, should we use the vertical bar for separating "Domain request status" from the domain name as we have on other pages?

src/registrar/templates/includes/summary_item.html
```diff
@@ -0,0 +1,49 @@
+<section class="summary-item margin-top-3">
```
neilmb, line 1: **comment**: I like this simplification very much to make the shared component here.

neilmb (approved):

src/registrar/templates/application_status.html
igorkorenfeld: Great point, updated.

PR #551, opened 2023-04-28
Making field labels consistent

michelle-rago:
# Update field label in request summary #

## 🗣 Description ##

Make field labels consistent

@igorkorenfeld I also noticed "Authorizing Official" in the app, but I can't find that instance in the code. (It should be "Authorizing official.") Can you take a look?
image


neilmb:
@michelle-rago I think that header is in `registrar/templates/application_status.html`?

michelle-rago:
> @michelle-rago I think that header is in `registrar/templates/application_status.html`?

The correct heading is in that file. It says "official." I can't find the offending "Official." 😕

igorkorenfeld (approved):
Looks good! Thanks for keeping things consistent!

PR #574, opened 2023-05-04
Update Github Action to fix conditionals

loganmeetsworld:
#  #

## 🗣 Description ##

This updates the conditional github action so that it actually deploys correctly. 

Github action for sandbox deploys has not been running correctly on prepended branches.

### DIFF
--- a/.github/workflows/deploy-sandbox.yaml
+++ b/.github/workflows/deploy-sandbox.yaml
@@ -5,18 +5,18 @@ run-name: Build and deploy developer sandbox for branch ${{ github.head_ref }}
 
 on:
   pull_request:
-    branches:
-      - 'ik/**'
-      - 'rjm/**'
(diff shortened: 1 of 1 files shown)


abroddrick (approved):

brandonlenz (approved):
I was curious why the original syntax didn't work as it looks correct to me based on[ this blog post](https://dev.to/github/running-github-actions-ci-cd-triggers-on-specific-branches-5e6m) outlining a similar tactic.

But `startsWith` works just fine!

loganmeetsworld:
@brandonlenz The tactic I tried prior on `pull_request` doesn't work because that is targeting the base, not the head. The previous tactic I was using was breaking as well because of some weirdness around curly brackets. GA still definitely has some rough edges vis a vis understandability.

Issue #562, opened 2023-05-01
Connect up nameserver management page with EPP commands


neilmb:
Ensure NameServers are set on save.

NOTE: Testing interactions of our Django app and the results of EPP commands is not something we have figured out before, probably some extra work there.

This should be completed after #493 .
relates to #848

PR #575, opened 2023-05-05
Form to change nameservers for approved domains

### LINKED ISSUE #326: Manage Approved Domain - DNS Name Servers Management Page
**Acceptance Criteria**

- Create DNS page for managing name server setting
- Create fields for adding nameservers
- Update DB as necessary
- Match comp: https://www.figma.com/proto/ninNjo0bTg7XfB4mvTp4hG/GetGov%2FDesigns?page-id=1585%3A14299&node-id=1598%3A17340&viewport=-5317%2C729%2C0.59&scaling=min-zoom&starting-point-node-id=1598%3A16996&show-proto-sidebar=1

Note: Per 12/16 - only should be available for approved domains

neilmb:
# Form to edit a domain's nameservers #

## 🗣 Description ##

This adds a view with a form to edit the nameservers for an approved domain. It is just the web form, IT DOES NOTHING. It calls a `set_nameservers` method on the `Domain` object which is a currently no-op.  Because of this, submitting changes via the form succeeds, but you will get back the exact same form as you started with, the values won't change.

Screen Shot 2023-05-05 at 11 27 28 AM

The "Add another name server" button uses Javascript to put additional forms onto the page. Those forms are processed when the form is submitted, BUT you won't see the additional fields after submitting because THIS DOES NOTHING, so the additional nameservers don't get saved so they don't show up on the form after saving.

Closes #326 (finally). #562 is the ticket for making the `set_nameservers` method actually do the thing with EPP that it should.


kewlguy781 (approved):
Look good to me. We need to have pa11y test these pages (I realize because the url use pk key, theres problem with pa11y running these pages). I tested the domain format, error show as expected when left blank, and received success message when putting in correct information (although it doesn't save at the moment).

(Neil had to add me under domain management for me to access to a registered domain to test. We do need to figure out how to make it work for new incoming dev team.)

src/registrar/templates/domain_nameservers.html
```diff
@@ -0,0 +1,52 @@
...
+          {% input_with_errors form.server %}
+        {% endwith %}
+      {% else %}
+        {% input_with_errors form.server %}
+      {% endif %}
+    {% endwith %}
+    </div>
+    {% endfor %}
+
+    <button type="submit" name="submit_button" class="usa-button usa-button--unstyled display-block" id="add-form">
```
abroddrick, line 39: @neilmb, myself, @brandonlenz and @rachidatecs  were wondering why the type here is submit? Couldn't it be type='button'?

src/registrar/assets/js/get-gov.js
```diff
@@ -230,4 +230,36 @@ function handleValidationClick(e) {
...
+  let addButton = document.querySelector("#add-form")
+  let totalForms = document.querySelector("#id_form-TOTAL_FORMS")
+
+  let formNum = serverForm.length-1
+  addButton.addEventListener('click', addForm)
+
+  function addForm(e){
+      e.preventDefault()
 
+      let newForm = serverForm[0].cloneNode(true)
```
rachidatecs, line 251:
```suggestion
      let newForm = serverForm[2].cloneNode(true)
```

src/registrar/templates/domain_nameservers.html
neilmb: You're right. The `type="submit"` was still present from a previous approach (that we use in e.g.  `registrar/templates/application_dotgov_domain.html`). I'll change it to `button`.

neilmb:
@rachidatecs @abroddrick @brandonlenz Thanks for the review on this!

src/registrar/tests/test_views.py
```diff
@@ -1222,6 +1233,55 @@ def test_domain_invitation_flow(self):
...
+        self.app.set_cookie(settings.SESSION_COOKIE_NAME, session_id)
+        with less_console_noise():  # swallow log warning message
+            result = nameservers_page.form.submit()
+        # form submission was a post, response should be a redirect
+        self.assertEqual(result.status_code, 302)
+        self.assertEqual(
+            result["Location"],
+            reverse("domain-nameservers", kwargs={"pk": self.domain.id}),
+        )
+        self.app.set_cookie(settings.SESSION_COOKIE_NAME, session_id)
```
rachidatecs, line 1261: We (@abroddrick @brandonlenz and me) are wondering why you're setting this cookie again?

src/registrar/tests/test_views.py
```diff
@@ -1222,6 +1233,55 @@ def test_domain_invitation_flow(self):
...
+        Uses self.app WebTest because we need to interact with forms.
+        """
+        nameservers_page = self.app.get(
+            reverse("domain-nameservers", kwargs={"pk": self.domain.id})
+        )
+        session_id = self.app.cookies[settings.SESSION_COOKIE_NAME]
+        self.app.set_cookie(settings.SESSION_COOKIE_NAME, session_id)
+        # first two nameservers are required, so if we empty one out we should
+        # get a form error
+        nameservers_page.form["form-0-server"] = ""
```
rachidatecs, line 1277: We (@abroddrick @brandonlenz and me) are wondering are you getting "form-0-server" from?

src/registrar/assets/js/get-gov.js
```diff
@@ -230,4 +230,36 @@ function handleValidationClick(e) {
...
+  let serverForm = document.querySelectorAll(".server-form")
+  let container = document.querySelector("#form-container")
+  let addButton = document.querySelector("#add-form")
+  let totalForms = document.querySelector("#id_form-TOTAL_FORMS")
+
+  let formNum = serverForm.length-1
+  addButton.addEventListener('click', addForm)
+
+  function addForm(e){
+      e.preventDefault()
```
brandonlenz, line 249: Now that the button is type="button", do we need to preventDefault?

src/registrar/tests/test_views.py
neilmb: This is the ID of the first form input element on the page. Django-webtest (that's what `self.app` is) https://github.com/django-webtest/django-webtest addresses the form fields by their ids https://docs.pylonsproject.org/projects/webtest/en/latest/forms.html#filling-a-form. In this case, I think that I got that ID by inspecting the page in my browser. ;)

src/registrar/assets/js/get-gov.js
neilmb: Likely true. I can test it manually locally. (We don't have a Javascript-capable testing setup yet.)

PR #577, opened 2023-05-05
Fix puppeteer crashing error

abroddrick:
# Fix Puppeteer crashing Node container #

## 🗣 Description ##
Changes the node container to also use the node.dockerfile (which says which browsers to use) just as the pa11y container does. 

On my mac, "docker-compose up" kept resulting in node crashing due to the chromium dependency of puppeteer.  Online it seemed most people having this issue were calling out they had an M1 chip Mac and that the chromium dependency puppeteer uses is missing files that are needed on M1s but not others. 
Original error output that this fixes.
image

See [this slack thread for more detail](https://cisa-corp.slack.com/archives/C03QM0JGSQG/p1683230458692039)

Note to developers: double check that this doesn't just fix it on my machine but cause a bug on yours, I would suggest deleting your docker image completely and make sure it rebuilds and runs just fine with this new change.

### DIFF
--- a/src/docker-compose.yml
+++ b/src/docker-compose.yml
@@ -62,6 +62,9 @@ services:
       - POSTGRES_PASSWORD=feedabee
 
   node:
+    build:
+      context: .
+      dockerfile: node.Dockerfile
     image: node
     volumes:
       - .:/app


SSPJ (approved):
LGTM, @abroddrick ! Nice work. 👍 

I did not have the bug you encountered, so I was only able to test that building from scratch with this branch on my local resulted in a working build.

The Basics
- [x] I read the code
- [x] I ran the code

Documentation

Security

Reliability

PR #578, opened 2023-05-06
Upgrade psycopg2 version to fix error on Mac M1

abroddrick:
#  #

## 🗣 Description ##

upgrades psycopg2 version in the pipfile

Psycopg2 had an error that causes it to crash the app and prevent it from running on M1 macs, but this bug is fixed in the newer version.


neilmb (approved):
Tests pass and it does what we need. 🚢

neilmb:
Thanks for pulling this out into its own PR @abroddrick. Our custom is to let PR authors click the green "Merge" button when the PR is approved.

abroddrick:
> Thanks for pulling this out into its own PR @abroddrick. Our custom is to let PR authors click the green "Merge" button when the PR is approved.

Thanks, will do!

PR #585, opened 2023-05-08
Update README with steps on accessing /admin

abroddrick:
#  #

## 🗣 Description ##

Added the steps needed to give oneself access to admin locally 

Both Rachid and I had questions on how to do this, so it seems like people onboarding in the future may have the same question and that it warrants documenting.

### DIFF
--- a/docs/developer/README.md
+++ b/docs/developer/README.md
@@ -36,7 +36,7 @@ You'll need to create the `.env` file yourself. Get the secrets from Cloud.gov b
 
 ## Adding user to /admin
 
-To be able to view and use /admin locally:
+The endpoint /admin can be used to view and manage site content, including but not limited to user information and the list of current applications in the database. To be able to view and use /admin locally:
 
 1. Login via login.gov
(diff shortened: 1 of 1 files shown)


docs/developer/README.md
```diff
@@ -34,6 +34,26 @@ DJANGO_SECRET_LOGIN_KEY="<...>"
...
+ ADMINS = [
+        {
+            "username": "<UUID here>",
+            "first_name": "",
+            "last_name": "",
+        },
+        ...
+ ]
+```
+
```
kewlguy781, line 56: Do you want to mention once it has been added to this, it will also create "fake domains" along with various statuses as well. This is a good way to verify that its working.

abroddrick: I didn't realize adding yourself to /admins is what triggered that, I assumed that was set elsewhere. Yes I will add that in.

abroddrick: updated let me know if there is anything else that should be changed

kewlguy781 (approved):
👍

rachidatecs (approved):

PR #587, opened 2023-05-09
Add character limit validation on textarea, increase char limit to 1000, increase textarea height

rachidatecs:
## 🗣 Description ##

- Add character limit validation on textarea
- Increase char limit to 1000
- Increase textarea height
(all apply "anything else", "type of work", and "purpose of your domain")

Resolves https://github.com/cisagov/getgov/issues/415


src/registrar/assets/sass/_theme/_uswds-theme-custom-styles.scss
```diff
@@ -422,3 +422,9 @@ abbr[title] {
   border-bottom: none;
   text-decoration: none;
 }
+
+.usa-textarea {
+  @include at-media('tablet') {
+    height: 20rem;
```
igorkorenfeld, line 428: I've been trying to avoid unnamed space and size units throughout the style sheet as best I could. Could we use a USWDS unit here? For 20 rem you could use `units('mobile')`

igorkorenfeld (approved):
This looks good to me, with one note on the styling. 

As @h-m-f-t  mentioned in the slack thread, it would be good to test (or analyze) how much users are inputting for these fields versus how much the analysts need. I wonder if having a max of 1000 characters creates an anchor that folks then try to get closer to. 

One thing I didn't mark off in the checklist is a unit test. I didn't see a test for a character limit error and I am not sure if this is something we would want to add? @SSPJ do you have any thoughts on that? 

The Basics
- [x] I read the code
- [x] I ran the code (tested in the `rjm` sandbox).


Documentation


Security


Reliability
- [ ] functionality is tested with unit or integration tests
- [x] Did dependency updates in Pipfile also get changed in requirements.txt?

SSPJ:
@rachidatecs and @igorkorenfeld , I wouldn't sweat not having it, but adding a few tests to the `TestFormValidation` class in `registrar/tests/test_forms.py` might be good practice for getting familiar with our unit tests.

rachidatecs:
100% on unit tests

SSPJ (approved):
@rachidatecs LGTM 👍 

Consider making the SCSS change @igorkorenfeld requested. Self-merge when you are ready.

src/registrar/assets/sass/_theme/_uswds-theme-custom-styles.scss
rachidatecs: 10-4

PR #588, opened 2023-05-09
Resize CISA logo in identifier

igorkorenfeld:
## 🗣 Description ##

Small PR to resize the CISA logo in the identifier so that it matches the size of the logo on get.gov

image


neilmb (approved):
Looks easy.

PR #604, opened 2023-05-11
Add sleep and increase timeout for pa11y github action

igorkorenfeld:
## 🗣 Description ##

This PR adjusts the pa11y test getting up action to try and fix pa11y failing to connect and run on some pages. 

The PR adds a sleep step and increases the timeout in pa11y's config to try and make sure that the server in the docker container is up and running.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -53,5 +53,6 @@ jobs:
       - name: run pa11y
         working-directory: ./src
         run: |
+          sleep 10;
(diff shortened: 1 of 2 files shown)


neilmb (approved):
This seems like a fine improvement. The longer timeout means a longer wait for this to fail if something is totally broken, but 30 seconds doesn't seem too painful. LGTM

loganmeetsworld (approved):

PR #609, opened 2023-05-12
Domain Interface

SSPJ:
## 🗣 Description ##

Adds properties and methods to the Domain model to allow accessing the registry through it.

tbd.


docs/operations/data_migration.md
```diff
@@ -28,8 +28,7 @@ information to make connections between registry users and the domains that
 they manage. The registrar stores very few fields about a domain except for
 its name, so it could be straightforward to import the exported list of domains
 from Verisign's `escrow_domains.daily.dotgov.GOV.txt`. It doesn't appear that
-that table stores a flag for active or inactive, so every domain in the file
-can be imported into our system as `is_active=True`.
+that table stores a flag for active or inactive.
```
abroddrick, line 31: was there a reason to remove the  "is_active=True" part, such as a new requirement?

SSPJ: It originally existed on `Domain` because `Domain` has been a stub standing in for something we hadn't developed yet.

I removed it in this PR because I think it will be better to ask the registry rather than caching `is_active` in the registrar database. If it turns out to be too much of a performance cost, you could bring it back, I think.

In terms of performance vs stability, I have been favoring stability in my design choices.

abroddrick (commented):
Publishing a couple comments I had while looking at this yesterday, not a full review of the  PR though.

docs/operations/data_migration.md
abroddrick: I definitely agree with favoring stability as well and like that approach. Not marking them as active by default also checks out for me, thanks for explaining!

src/requirements.txt
```diff
@@ -1,17 +1,17 @@
 -i https://pypi.python.org/simple
 asgiref==3.6.0 ; python_version >= '3.7'
-boto3==1.26.69
-botocore==1.29.69 ; python_version >= '3.7'
+boto3==1.26.85
```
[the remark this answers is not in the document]
SSPJ: Yup, super awkward setup to have both a requirements.txt and a Pipfile. Context and documentation for how to update is here: https://github.com/cisagov/getgov/blob/main/docs/operations/runbooks/update_python_dependencies.md

src/epplibwrapper/client.py
```diff
@@ -67,54 +67,47 @@ def __init__(self) -> None:
...
             if response.code >= 2000:
-                raise RegistryError(response.msg)
+                raise RegistryError(response.msg, code=response.code)
             else:
                 return response
 
-    def send(self, command):
+    def send(self, command, *, cleaned=False):
         """Login, send the command, then close the connection. Tries 3 times."""
+        # try to prevent footguns
```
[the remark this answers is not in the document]
SSPJ: People overly familiar with Django expecting the framework to do string sanitization for them. As it stands, epplib does no checking, so we have to be extra vigilant not to send bad strings.

Have updated this comment to be more illuminating.

PR #621, opened 2023-05-17
Implement domain security email page with stubs for EPP

rachidatecs:
## 🗣 Description ##

- Stubs for EPP GET and POST
- Add security email form
- Add security email view
- Add security email template
- Routing
- Testing
- Remove security email from DomainInformation model and run migrations*

Completes[ 516](https://github.com/cisagov/getgov/issues/516)

 *NOTE: Security email will be done via EPP instead of DomainInformation.

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue-default.yml
+++ b/.github/ISSUE_TEMPLATE/issue-default.yml
@@ -1,0 +1,37 @@
+name: Standard Issue Tempalte
+description: Create an Issue
+title: "[Issue]:"
+labels: ["research"]
+assignees:
+  - kitsushadow
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Please fill this out best you can
+  - type: textarea
+    id: why
+    attributes:
+      label: Why or Acceptance Criteria
+      description: What is the reason we determined we need an issue.
+      placeholder: This issue is a product of...
+      value: "While working on X or researching Y it was discovered that Z was an issue that need captured."
+    validations:
+      required: true
+  - type: dropdown
+    id: type
+    attributes:
+      label: Issue Type
+      description: What type of issue is this
+      options:
+        - research (Default)
+        - dev
+        - product
+    validations:
+      required: true
+  - type: textarea
+    id: Dependencies
+    attributes:
+      label: Link dependent issues
+      description: If this ticket is dependent on another issue or blocks a current issue, please link.
+      render: shell
--- a/docs/developer/database-access.md
+++ b/docs/developer/database-access.md
@@ -12,19 +12,34 @@ to get a `psql` shell on the sandbox environment's database.
 
 ## Running Migrations
 
-When new code changes the database schema, we need to apply Django's migrations.
+When new code changes the database schema (ie, you change a model or pull some
+code that has), we need to apply Django's migrations.
+
+### On Local
+
+```shell
+docker-compose exec app bash
+./manage.py makemigrations
+```
+
+Then perform docker-compose down & docker-compose up to run with the new migrations.
+
+### On Cloud.gov
+
 We can run these using CloudFoundry's tasks to run the `manage.py migrate`
 command in the correct environment. For any developer environment, developers
 can manually run the task with
 
 ```shell
-cf run-task getgov-ENVIRONMENT --command 'python manage.py migrate' --name migrate
+cf run-task getgov-ENVIRONMENT --wait --command 'python manage.py migrate' --name migrate
 ```
+
+(The optional 'wait' argument will wait until the environment is stable)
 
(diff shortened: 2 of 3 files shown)


src/registrar/views/domain.py
```diff
@@ -96,6 +96,55 @@ def form_valid(self, formset):
...
+        """The form is valid, call setter in model."""
+
+        # Set the security email from the from
+        try:
+            new_email = form.cleaned_data["security_email"]
+        except KeyError:
+            # no server information in this field, skip it
+            pass
+        domain = self.get_object()
+        domain.set_security_email(new_email)
```
abroddrick, line 139: You used new_email here but if it fails on the try new_email will have never been set.   In the except you can set new_email to none if that is the intention and requirement if a security email doesn't exist. Or just use a else after the except and only set the domain email if there is in fact an email available to use.

src/registrar/views/domain.py
```diff
@@ -96,6 +96,55 @@ def form_valid(self, formset):
...
+        if form.is_valid():
+            # there is a valid email address in the form
+            return self.form_valid(form)
+        else:
+            return self.form_invalid(form)
+
+    def form_valid(self, form):
+        """The form is valid, call setter in model."""
+
+        # Set the security email from the from
```
abroddrick, line 132:
```suggestion
        # Set the security email from the form
```
typo

abroddrick (commented):
Didn't do a full review, just adding a couple comments.

src/registrar/views/domain.py
rachidatecs: You're right, that pass does not make sense. The exception should not happen so maybe that whole try/except block is unneeded and can be replaced by new_email = form.cleaned_data["security_email"]?

kewlguy781:
Look good to me. I have reviewed all the files and codes (thank you for adding the updated migrations file). I would like to see more information under "Context" explaining that security email will be done via EPP instead of DomainInformation, which is why they are seeing these changes. Please add these information, and once you do that, go ahead and merge it.

kewlguy781 (approved):
Approved with the comment entered previously.

src/registrar/views/domain.py
abroddrick: I would say, if there is a chance of a crashing exception on the server side, best practice is to catch it. So I wouldn't remove the try/except; for consistency you could follow this pattern where you set new_email to have a default value on error. 
image
or 
just move the domain.set_security_email portion into an else. The else means "If the exception is not hit", so no email gets added if there is no email (which sounds like that shouldn't happen really anyways).
```
       else:
              domain = self.get_object()
              domain.set_security_email(new_email)
```

That being said I saw another part of the code not doing the try/except block too, so I am fine with it being dealers choice and we can circle back to it if we ever need to.

rachidatecs: Let's see if 3rd time's the charm. I removed the 'KeyError' on the except as it seemed like it's specific to the Neil's use case and initialized new_email to an empty string.

neilmb (approved):
This is great work @rachidatecs, very nice to see that these pages are repeatable. I think the only comment that needs following up on is adding the page to the `test_no_domain_role` unit test. 

This is our "saved reply" review checklist. A check means that I checked that the criteria is met or that I checked that the item does not apply.

## Accessibility
- [X] New pages have been added to .pa11yci file so that they will be tested with our automated accessibility testing
   
  We don't have any of these domain detail pages in that file right now, so let's not figure it out on this PR.

    - [X] All relevant images use an img tag
    - [X] All images have alt attributes
    - [X] Multimedia is tagged. All multimedia has appropriate captioning and audio description
    - [X] Tab order is logical
    - [X] Site doesn’t timeout unexpectedly. Identify elements that may “timeout” and verify that the user can request more time
    - [X] Tables are coded properly. Tables have proper headers and column attributes
    - [X] Frames are named. All frames have a name element
    - [X] Flashing elements are compliant. Elements that flash on screen do so at a rate of less than 3 Hz
    - [X] The language for sections on the page that differ from the site language are set
    - [X] Required plugins are linked on the page
 
## Device Matrix
- [X] firefox
- [X] chrome/chromium/edge
- [X] safari
   
  We have a long-standing bug that Safari doesn't work with Login on localhost, but it appears to work on the sandbox.

    - [X] at 480px
    - [X] at 1024px

src/registrar/tests/test_views.py
```diff
@@ -1063,6 +1063,11 @@ def test_not_logged_in(self):
         )
         self.assertEqual(response.status_code, 302)
 
+        response = self.client.get(
```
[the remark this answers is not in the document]
rachidatecs: I merged too soon. I'll make a follow-up PR

PR #625, opened 2023-05-17
Add public site helper tag for Django site

### LINKED ISSUE #612: Make a shorthand template variable to point to the public site
A/C: There is a simple way in Django to refer to a page on our public site such as {% public_site_page /domains/requirements %}

The Django application has many links out to the public site. Rather than having to specify the base of the public site URL in every single link out, we can provide a configuration setting and a way to use it in our HTML templates so that we can write links simply, and also when the public site changes URL (e.g. from federalist preview to beta.get.gov to get.gov), we will be able to make all those links point to the correct location easily.

neilmb:
# Add a template helper to refer to pages on the public site #

## 🗣 Description ##

The Django application will have lots of links out to the help content on the public site. This PR adds a template tag called `public_site_url` that takes the path part of a URL on the public site and adjoins the base URL of that site as configured in an environment variable called `GETGOV_PUBLIC_SITE_URL`. For example, `{% public_site_url 'domains/requirements/' %}`

This should allow changing all of the links to the public site to point to a different site (e.g. migrating from beta.get.gov to get.gov) at the same time by changing that environment variable in Cloud.gov.

@michelle-rago Do we need to document the use of this tag anywhere? I included in this PR the first link to the requirements that you had started on.

Closes #612.


michelle-rago:
@neilmb The requirements link I'm seeing on the purpose page in your sandbox is https://getgov-nmb.app.cloud.gov/domains/requirements/. Is that what it should be?

michelle-rago:
> @michelle-rago Can you try again on getgov-nmb.app.cloud.gov? I think I set it correctly. I still see the `link TBD` styling, but I think @igorkorenfeld added that for things that start with "federalist"?

That works! This is the link now https://federalist-877ab29f-16f6-4f12-961c-96cf064cf070.sites.pages.cloud.gov/site/cisagov/getgov-home/domains/requirements/. 😺

igorkorenfeld:
I did, if we would rather have the text say something like "Temporary link" Or "Check link"  or federalist links I am happy to make an update (can also style it a little differently to differentiate from the TBD ones).

src/.env-example
```diff
@@ -1,2 +1,3 @@
 DJANGO_SECRET_KEY=""
 DJANGO_SECRET_LOGIN_KEY=""
+GETGOV_PUBLIC_SITE_URL="https://federalist-877ab29f-16f6-4f12-961c-96cf064cf070.sites.pages.cloud.gov/site/cisagov/getgov-home/"
```
loganmeetsworld, line 3: Do we need this in a future `.env`? Or do we just need it in the docker-compose?

ops/scripts/manifest-sandbox-template.yaml
```diff
@@ -20,6 +20,8 @@ applications:
     DJANGO_BASE_URL: https://getgov-ENVIRONMENT.app.cloud.gov
     # Tell Django how much stuff to log
     DJANGO_LOG_LEVEL: INFO
+    # default public site location
+    GETGOV_PUBLIC_SITE_URL: https://beta.get.gov
```
loganmeetsworld, line 24: Does this need to then be put in all our manifests?

src/.env-example
loganmeetsworld: I guess some clarity on docker-compose vs. manifest vs. .env would be good.

loganmeetsworld (approved):

loganmeetsworld (approved):

PR #627, opened 2023-05-17
Document local migrations, mention --wait for cf migrations

rachidatecs:


docs/developer/database-access.md
```diff
@@ -12,19 +12,32 @@ to get a `psql` shell on the sandbox environment's database.
...
 ## Running Migrations
 
+### On Local
+
+```shell
+docker-compose exec app bash
+./manage.py makemigrations
+```
+
+The docker compose down then up to run the new migrations.
```
abroddrick, line 22:
```suggestion
Then perform docker-compose down & docker-compose up to run with the new migrations.
```

docs/developer/database-access.md
```diff
@@ -12,19 +12,32 @@ to get a `psql` shell on the sandbox environment's database.
 
 ## Running Migrations
 
+### On Local
+
```
abroddrick, line 16: maybe add context here, such as why you may want to do it. (ie you changed a model, or have pulled in new code and are seeing the warning when you do docker-compose up).

abroddrick (approved):
Add some comments and suggestions, but overall looks great.

PR #628, opened 2023-05-18
update suggestions from h-m-f-t

kitsushadow:
# Updates to the Issue Template #

## 🗣 Description ##
Suggestions provided by folks to improve the content of the issue as well as its presentation.

This issue template will allow any issue creator a simple guided form to capture data regarding a discovery or addition.

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue-default.yml
+++ b/.github/ISSUE_TEMPLATE/issue-default.yml
@@ -1,32 +1,32 @@
-name: Standard Issue Tempalte
+name: Standard Issue Template
 description: Create an Issue
-title: "[Issue]:"
+title: "[Issue]: "
 labels: ["research"]
-assignees:
-  - kitsushadow
+# assignees:
+#   - kitsushadow
 body:
   - type: markdown
     attributes:
       value: |
-        Please fill this out best you can
+        Describe an issue, idea, task, bug, or story.
   - type: textarea
     id: why
     attributes:
       label: Why or Acceptance Criteria
-      description: What is the reason we determined we need an issue.
+      description: What is the reason it was determined this is an issue.
       placeholder: This issue is a product of...
-      value: "While working on X or researching Y it was discovered that Z was an issue that need captured."
+      value: "While working on X or researching Y it was discovered that Z was an issue that needed captured."
     validations:
       required: true
   - type: dropdown
     id: type
     attributes:
       label: Issue Type
-      description: What type of issue is this
+      description: What kind of role would take action on your issue?
       options:
         - research (Default)
-        - dev
-        - product
+        - development
+        - product team
     validations:
       required: true
   - type: textarea


.github/ISSUE_TEMPLATE/issue-default.yml
```diff
@@ -1,32 +1,32 @@
-name: Standard Issue Tempalte
+name: Standard Issue Template
 description: Create an Issue
-title: "[Issue]:"
+title: "[Issue]: "
```
h-m-f-t, line 3: I'd prefer we categorize in ways other than the title of the issue (we have various flag options, for instance). Let's not prepend the issue title with "[Issue]:".

kitsushadow:
@h-m-f-t  i made some adjustments in line with your thoughts. I think it reads a bit more invitingly now.

.github/ISSUE_TEMPLATE/issue-default.yml
```diff
@@ -1,32 +1,32 @@
...
+        Describe an issue, idea, task, bug, or story.
   - type: textarea
     id: why
     attributes:
       label: Why or Acceptance Criteria
-      description: What is the reason we determined we need an issue.
+      description: What is the reason it was determined this is an issue.
       placeholder: This issue is a product of...
-      value: "While working on X or researching Y it was discovered that Z was an issue that need captured."
+      value: "While working on X or researching Y it was discovered that Z was an issue that needed captured."
```
brandonlenz, line 18: With bug reports and user stories separated into different templates, I recommend doing something like this for the user story template, to prefill the format we might want folks to answer in

```suggestion
      value: |
        As a <!--who-->
        I want <!--what-->
        So that <!--why-->
```

brandonlenz: Not sure if it will use the `<!--comment-->` md syntax, so we can leave that out if not

.github/ISSUE_TEMPLATE/issue-default.yml
```diff
@@ -1,32 +1,32 @@
...
 body:
   - type: markdown
     attributes:
       value: |
-        Please fill this out best you can
+        Describe an issue, idea, task, bug, or story.
   - type: textarea
     id: why
     attributes:
       label: Why or Acceptance Criteria
```
brandonlenz, line 15:
```suggestion
      label: User story
```

.github/ISSUE_TEMPLATE/issue-default.yml
brandonlenz: Then we could add an additional textarea of acceptance criteria, perhaps?

I don't have strong preferences on format, but I've seen folks enjoy the given / when / then format in the past, so I'll add that as a suggestion, e.g:

> ## Example Acceptance Criteria – Given/When/Then
> 
> ### User story:
> 
> As a product manager.
> 
> I want to score potential ideas.
> 
> So that I can decide what to include on my product roadmap.
>
> ### Acceptance criteria for that user story could be:
> 
> Scenario: The product manager adds potential ideas and ranks the best ideas based on benefit versus cost.
> 
> Given that I have added two or more ideas and scored them using the Benefit vs Cost scoring model
> 
> When I click the Rank button
> 
> Then ideas are sorted with the top-scoring ideas at the top.
> 

- [source](https://www.productplan.com/glossary/acceptance-criteria/)

.github/ISSUE_TEMPLATE/issue-default.yml
```diff
@@ -1,32 +1,32 @@
...
     attributes:
       label: Issue Type
-      description: What type of issue is this
+      description: What kind of role would take action on your issue?
       options:
         - research (Default)
-        - dev
-        - product
+        - development
+        - product team
```
brandonlenz, line 29: I could see this being useful. It is a bummer that dropdowns don't seem to integrate with things like labels just yet (https://github.com/orgs/community/discussions/4266 😢)

I might even suggest moving it closer to the top of the issue template so that it appears at a glance, setting the context for how we read the rest.

.github/ISSUE_TEMPLATE/issue-default.yml
```diff
@@ -1,32 +1,32 @@
...
     validations:
       required: true
   - type: textarea
```
brandonlenz, line 32: Curious about this one. I don't think we'll want to use the `render` property for issue linking, as links won't auto-url link with the `#{issue_number}` syntax once the issue is created and the contents of this are placed in a code block (assuming I'm understanding [the docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#attributes-for-textarea) correctly)

.github/ISSUE_TEMPLATE/issue-default.yml
```diff
@@ -1,32 +1,32 @@
-name: Standard Issue Tempalte
-description: Create an Issue
-title: "[Issue]:"
-labels: ["research"]
-assignees:
-  - kitsushadow
+name: Issue
+description: Provide a title for the issue you are describing 
+title: "Please provide a clear title"
```
brandonlenz, line 3: If we go the route of splitting up the templates, I also recommend something like this, to coincide with the corresponding type of issue being opened

```suggestion
title: "[Story]: "
```

brandonlenz: Optional if using labels, but sometimes the visual prefix is nice too, since labels aren't always visible in every project view, or there might be many labels on a given story

brandonlenz (changes requested):
In addition to the suggestion of splitting up into at least 2 issue templates, I think we should do an audit of the [labels](https://github.com/cisagov/getgov/labels)

Not all labels have descriptions, so I'm confused about when to use the design vs research label, for example.

If we want to label stories as `story` and bugs as `bug`, for example, we'll need to add this labels.

.github/ISSUE_TEMPLATE/issue-default.yml
brandonlenz: Ah I see @h-m-f-t's comment about this already. Fair enough, let's use our other tools!

h-m-f-t: I think adding "Bug" can make sense. I think what I'm opposed to is making Random Issue Filer follow our naming conventions to get our attention. We can always edit the title to reflect that it's a bug.

.github/ISSUE_TEMPLATE/issue-default.yml
```diff
@@ -1,32 +1,32 @@
...
+labels: ["review"]
```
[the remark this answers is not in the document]
abroddrick: I strongly agree with splitting up the templates as well; a bug ticket, research task, and feature request are all very different and should have different information on them. A research task can be pretty vague or generic I imagine, but a bug ticket needs to clearly indicate how to replicate the bug. Likewise a feature needs a description of what is being added but with enough detail that an engineer is not confused or having to do refinement in order to work on the ticket. Having guidelines  (aka templates) for each of these really helps tickets maintain consistency no matter who is filling them out.

brandonlenz:
Added a bug issue template example in #641. The idea being that we're able to ask very specifically about what we'd want to know for a bug report, which otherwise wouldn't make sense to fill out for most user stories. It also applies the "bug" label automatically, and a title prefix (if we want to do that)

.github/ISSUE_TEMPLATE/issue-default.yml
brandonlenz: 100% agree. Naming conventions are subject to change, and too easy to mistakenly deviate from that I would not recommend filtering based on name alone. Labels are probably the most effective way to categorize in Github

.github/ISSUE_TEMPLATE/issue-default.yml
```diff
@@ -1,32 +1,32 @@
...
+# assignees:
+#   - kitsushadow
```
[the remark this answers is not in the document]
kitsushadow: I was looking at this the other day. I agree some workflows would add some great QoL

brandonlenz (approved):
I don't think we're at our final iteration yet, but knowing this is a zero-risk change that's hard to preview until merged, I'm good to approve and merge aggressively. 

Let's keep iterating

PR #651, opened 2023-05-24
Add abstract views that enforce permissions

### LINKED ISSUE #544: Prevent unauthorized access by using a base class
The current permission system is opt-in, which allows too much room for mistakes. Update the permission system to be opt-out.

One potential method is to make all views class-based and ensure they inherit from a common base class which has appropriate permission mixins. Other methods may exist as well. Django's own auth system may provide some inspiration.

This work may intersect with #543.

neilmb:
# Enforce permissions using abstract View subclasses #

## 🗣 Description ##

Authorization is the overall process of ensuring that only the appropriate users have access to various resources. This PR introduces some abstract base classes  for views on our various data types that enforce appropriate permissions for that data.  `DomainPermissionView`, `DomainApplicationPermissionView`, and `DomainInvitationPermissionDeleteView` check the appropriate permission description for each of those models.

This also adds additional test cases for each view that it raises the appropriate permission denied error when accessed by the "wrong" user.

Closes #544 and fixes bugs with being able to withdraw someone else's domain applications or domain invitations because those views did not have authorization protection.

### DIFF
--- a/src/registrar/config/urls.py
+++ b/src/registrar/config/urls.py
@@ -59,12 +59,12 @@ urlpatterns = [
     ),
     path(
         "application/<int:pk>/withdraw",
-        views.ApplicationWithdraw.as_view(),
+        views.ApplicationWithdrawConfirmation.as_view(),
         name="application-withdraw-confirmation",
     ),
     path(
         "application/<int:pk>/withdrawconfirmed",
-        views.ApplicationWithdraw.updatestatus,
+        views.ApplicationWithdrawn.as_view(),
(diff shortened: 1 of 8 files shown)


SSPJ (approved):
👍 

@neilmb and I had a discussion that this may be better implemented as middleware to avoid a proliferation of view-specific subclasses. However, I tested that this PR does fix a presently existing bug in our permission system, so I'm approving it. Thank you, Neil, for helping to keep the app secure!

PR #652, opened 2023-05-25
Improve bug issue template

brandonlenz:
## 🗣 Description ##

Iterate on the first pass of the bug issue template:
- Instead of placeholder text for fields where a specific format is desired, place examples in description so that it is easier to copy past (if desired), and so that the example doesn't go away when a user starts typing
  - Note: We did not opt for default value, because deleting content and replacing it is just as annoying/cumbersome
- Mark optional fields more clearly as (optional)

We want issue templates to be as ergonomic as possible, as the time spent now will pay dividends when folks are more easily able to contribute quality issues with relevant details.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -24,25 +24,30 @@ body:
     id: steps-to-reproduce
     attributes:
       label: Steps to Reproduce
-      description: "How can the issue be reliably reproduced? feel free to include screenshots or other supporting artifacts"
-      placeholder: |
+      description: |
+        How can the issue be reliably reproduced? Feel free to include screenshots or other supporting artifacts
+        
         Example:
         1. In the test environment, fill out the application for a new domain
         2. Click the button to trigger a save/submit on the final page and complete the application 
         3. See the error
+      value: |
+        1.
     validations:
       required: true
   - type: textarea
     id: environment
     attributes:
-      label: Environment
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -24,25 +24,30 @@ body:
...
-      placeholder: |
+      description: |
+        How can the issue be reliably reproduced? Feel free to include screenshots or other supporting artifacts
+        
         Example:
         1. In the test environment, fill out the application for a new domain
         2. Click the button to trigger a save/submit on the final page and complete the application 
         3. See the error
+      value: |
+        1.
```
brandonlenz, line 35: In this case, we know only that all steps will start at `1.`, so we can provide that to save some clicks/typing

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -24,25 +24,30 @@ body:
...
     validations:
       required: true
   - type: textarea
     id: environment
     attributes:
-      label: Environment
-      description: "Where is this issue occurring? If related to development environment, list the specific relevant tool versions"
-      placeholder: |
+      label: Environment (optional)
```
brandonlenz, line 41: No default value provided here (even though `-` would be convenient-ish), because the field is optional, so it should be blank to start.

h-m-f-t (approved):

PR #655, opened 2023-05-25
Add example User Story issue template

brandonlenz:
## 🗣 Description ##

Similar to the bug issue template (#652), add an example user story issue template

This is a foundational template with a lot of my own bias. 
This PR gets _something_ out there that we can see and interact with, then decide whether to keep it, tweak it, or get rid of it.

The issue template will looks something like this once merged, but with the hint text between the question labels and their respective inputs, instead of below the input 🤪:

image

### DIFF
--- a/.github/ISSUE_TEMPLATE/user-story.yml
+++ b/.github/ISSUE_TEMPLATE/user-story.yml
@@ -1,0 +1,62 @@
+name: User Story
+description: Capture actionable sprint work
+title: "[Story]: "
+labels: ["story"]
+
+body:
+  - type: markdown
+    id: help
+    attributes:
+      value: |
+        > **Note**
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/user-story.yml
```diff
@@ -0,0 +1,62 @@
+name: User Story
+description: Capture actionable sprint work
```
brandonlenz, line 2: @PaulKuykendall may have preferences for what this should say. 

This is the text that will appear under the "User story" option when selecting the type of issue you want to create on the https://github.com/cisagov/getgov/issues/new/choose page

.github/ISSUE_TEMPLATE/user-story.yml
```diff
@@ -0,0 +1,62 @@
...
+title: "[Story]: "
```
brandonlenz, line 3: Like the bug template, not married to prefixing them, so we'll just be consistent whichever way we go. The prefix is not for search/organization purposes. The default label does that for us.

h-m-f-t (approved):

PR #656, opened 2023-05-26
Create a DraftDomain model for requested domains

SSPJ:
## 🗣 Description ##

This PR places the strings of characters which applicants request into a DraftDomain table, rather than in the Domain table. They are copied from one to the other when the DomainApplication record is approved.

The motivation for this change is twofold: first, the Domain model no longer needs to maintain a complicated uniqueness constraint to account for the fact that approved domains must be unique while requested domains need not be unique; second, the Domain model no longer needs to distinguish approved domains from requested domains when performing actions on model instances.

⚠️ This PR is not targeting main. Update the base branch before merging! ⚠️


neilmb (approved):
This seems like a significant change and I think it's very impressive to see it made and passing the test suite. 

I think that the addition of `DraftDomain` changes the timeline or lifecycle of our models from `docs/architecture/diagrams/model_timeline.md`. Is it appropriate to update that documentation as part of this PR, or to make a follow-up ticket to do so?

SSPJ:
> Is it appropriate to update that documentation as part of this PR, or to make a follow-up ticket to do so?

Yep, done. (Hopefully I didn't make any mistakes; I am new to that diagramming syntax.)

PR #657, opened 2023-05-26
Write a new Domain interface

SSPJ:
## 🗣 Description ##

This had formerly been a WIP on #609 . It heavily refactors the Domain class.

There are numerous changes, the most prominent being a switch to using `@property` rather than `get_/set_`. This will help during caching, as `property` can be subclassed and custom code inserted before/after each property lookup.

Beyond that, this PR attempts to provide methods for all the common actions a caller of the interface may want to take.

Each of the methods has type hinting. Although most of the implementation details are missing on this PR, these type hints (combined with the docstrings) are sufficient to fully describe the shape of an API for interacting with the registry.

⚠️ This PR is not targeting main. Update the base branch before merging! ⚠️


src/registrar/models/domain.py
```diff
@@ -91,221 +93,193 @@ class Status(models.TextChoices):
...
-            if self.domain_application.status != DomainApplication.APPROVED:
-                raise ValueError("Cannot activate. Application must be approved.")
-        if Domain.objects.filter(name=self.name, is_active=True).exists():
-            raise ValueError("Cannot activate. Domain name is already in use.")
-        # TODO: depending on the details of our registry integration
-        # we will either contact the registry and deploy the domain
-        # in this function OR we will verify that it has already been
-        # activated and reject this state transition if it has not
-        pass
+            {"registrant": "jd1234", "admin": "sh8013",...}
```
neilmb, line 128: **comment (non-blocking)**: In the future, when implementing this method, I think that we now have these dictionary keys "registrant", "admin", etc. as class constants under `PublicContact`, maybe we should re-use those comments here.

src/registrar/models/domain.py
```diff
@@ -91,221 +93,193 @@ class Status(models.TextChoices):
...
-        return self.name
+    @expiration_date.setter  # type: ignore
+    def expiration_date(self, ex_date: date):
+        raise NotImplementedError()
 
-    def nameservers(self) -> List[str]:
-        """A list of the nameservers for this domain.
+    @property
+    def password(self) -> str:
+        """Get the `auth_info.pw` element from the registry. Not a real password."""
```
neilmb, line 158: **question**: This comment about not being a real password is surprising enough that I feel like I want more explanation about what it means. Can we add more here, or maybe better, add a section in our documentation on how EPP uses `auth_info.pw` and then link to that section in comments like this?

neilmb (approved):
I guess I'll say it again: It's very exciting to see this interface growing and to imagine how it can be put to use in actually managing domains from registrar to registry.

(Is there a minimal slice of functionality that we could implement in the application first to see this round trip working?)

PR #665, opened 2023-05-31
Remove compliance documentation

loganmeetsworld:
#  #

## 🗣 Description ##

We have not been using trestle here and recent discussions mean that we are likely not going to be using this or OSCAL for our documentation. To remove confusion about what this folder is doing, let us just remove it.


h-m-f-t (approved):
I would love to do compliance documentation like this, but this is the right call for now.

PR #673, opened 2023-06-01
Use login required middleware

### LINKED ISSUE #543: Prevent unauthenticated access by using middleware
It is easy to forget `@login_required` decorators and `LoginRequired` mixins, so a middlware layer should be installed to ensure only ok-listed routes are accessible when a user is not signed in.

A suitable 3rd party middleware package may not exist, so it may be necessary to write a small middleware. Existing examples in the codebase can be used.

Middleware may also help with bugs like #537 and #405. It could also resolve and close #486.

This work may intersect with #544.

neilmb:
# Use login required middleware #

## 🗣 Description ##

We don't want to have to remember to add `LoginRequired` to every one of our views, so we are going to use a `LoginRequiredMiddleware` that enforces `LoginRequired` on every single request. (There is a decorator and mixin for "login not required", see https://pypi.org/project/django-login-required-middleware/ for more details.) We can't require login for the URLs in our OIDC app under `/openid` because they need to be able to handle the initial request and its callbacks.

This also adds a wild unit test that iterates through every URL in our Django configuration and tests that every one of them requires login (except for the few exceptions). I just ~stole~borrowed that test from https://github.com/18F/tock/blob/main/tock/tock/tests/test_url_auth.py

Closes #543


SSPJ (approved):
LGTM 👍 

The Basics
- [x] I read the code
- [x] I ran the code

Documentation
- ❓  changes to “how we do things” are documented in READMEs -- worth checking to see if any dev docs need updating

Security

Reliability
- [x] functionality is tested with unit or integration tests 💯

src/registrar/tests/test_views.py
```diff
@@ -35,10 +35,9 @@ def test_health_check_endpoint(self):
         self.assertContains(response, "OK", status_code=200)
 
     def test_home_page(self):
-        """Home page should be available without a login."""
+        """Home page should NOT be available without a login."""
```
loganmeetsworld, line 38: Thank you! Do we need to remove the non-auth'd view for the homepage so that only the dashboard is on the homepage?

neilmb: Done.

src/Pipfile.lock
```diff
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "6a6f3722d88e1f0059e38f23c104708d7922d41b42be1f17287cd4d84dcf9b05"
+            "sha256": "187278933085d8c5448015e2d28c23934c2098f4aadd82157379174629d8cf6b"
```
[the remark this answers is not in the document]
neilmb: I think these are all merged up now.

neilmb:
OWASP tests finally passing, merging.

PR #683, opened 2023-06-05
Update styles on Domain overview page

igorkorenfeld:
## 🗣 Description ##

This PR follows up on the comments in #631 with a number of style updates. It moves the edit button back up to the right, and reduces the size of the heading for each section. Additionally, because the `summary-item` include is used on both the "Domain overview" page and the "Domain request summary" page, but requires different heading levels, this PR adds a `heading_level` option to the include. Thanks to @neilmb for advice on setting that up!

Updates to the  summary item include on the Domain overview page

Updates to the  summary item include on the Submitted request page

### DIFF
--- a/src/registrar/templates/application_status.html
+++ b/src/registrar/templates/application_status.html
@@ -4,11 +4,11 @@
 {% load static url_helpers %}
 
 {% block content %}
-<main id="main-content" class="grid-container register-form-step">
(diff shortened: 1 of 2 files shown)


neilmb (approved):
Looks like good changes, thanks for cleaning up after us!

PR #689, opened 2023-06-07
Send email to submitter when application status changes to "investigating"

rachidatecs:
## 🗣 Description ##

- Implement 'investigating' email templates
- Trigger email when application status changes to "investigating"
- Componentize completed_application for unit tests
- Componentize application summary in email templates
- Unit test

Resolves [472](https://github.com/cisagov/getgov/issues/472)


src/registrar/admin.py
```diff
@@ -50,13 +53,51 @@ class MyHostAdmin(AuditedAdmin):
...
+                    original_obj.submitter is None
+                    or original_obj.submitter.email is None
+                ):
+                    logger.warning(
+                        "Cannot send status change (in review) email,"
+                        "no submitter email address."
+                    )
+                    return
+                try:
+                    print(
```
neilmb, line 77: **question**: We don't usually leave debugging print statements in. Would this be better as a `logging.info` message that email is being sent to this email address?

src/registrar/templates/emails/status_change_in_review.txt
```diff
@@ -0,0 +1,81 @@
...
+
+
+THANK YOU
+
+.Gov helps the public identify official, trusted information. Thank you for
+requesting a .gov domain.
+
+----------------------------------------------------------------
+
+SUMMARY OF YOUR DOMAIN REQUEST
```
neilmb, line 37: **suggestion**: Since we have access to the power of Django's templates here in our emails, I think that we should make a separate include file for this domain summary so that we can just `{% include emails/includes/application_summary.html %}` here and in `templates/emails/submission_confirmation.txt`. That way, if we fix or improve the summary section we don't have to follow the copy-paste and repeat the fixes.

src/registrar/tests/test_admin.py
```diff
@@ -0,0 +1,131 @@
...
+        args, kwargs = call_args
+
+        # Retrieve the email details from the arguments
+        from_email = kwargs.get("FromEmailAddress")
+        to_email = kwargs["Destination"]["ToAddresses"][0]
+        email_content = kwargs["Content"]
+        email_body = email_content["Simple"]["Body"]["Text"]["Data"]
+
+        # Assert or perform other checks on the email details
+        expected_string = "Your .gov domain request is being reviewed"
```
neilmb, line 122: **question (non-blocking)**: I like this test very much. I noticed there is some error checking code in `save_model` for when various things might be none. Should we be making additional unit tests to verify the correct behavior in those error situations?

src/registrar/admin.py
```diff
@@ -50,13 +53,51 @@ class MyHostAdmin(AuditedAdmin):
     inlines = [HostIPInline]
 
 
+class DomainApplicationAdmin(AuditedAdmin):
+
+    """Customize the applications listing view."""
+
+    # Trigger action when a fieldset is changed
+    def save_model(self, request, obj, form, change):
```
neilmb, line 61: **question**: I can see where for a custom admin view, this is where state-change side effects can be made, but my concern is that it isn't very well integrated with the finite state machine that we use for application status (see https://github.com/cisagov/getgov/blob/main/docs/architecture/decisions/0015-use-django-fsm.md). Ideally, I would like to see this email sending code in an `investigate` transition method inside of `models/domain_application.py` like we do with `submit` (and ideally `withdraw` https://github.com/cisagov/getgov/issues/690). But I am not clear how that can integrate with the way that the admin allows directly editing the application's `status` field. Do we need to inspect the status change that was made in the admin view here and call the transition method from this method? Or is this a place where we can use something like `django-fsm-admin` to make the admin views aware of our transition methods? (https://github.com/gadventures/django-fsm-admin#usage)

neilmb (changes requested):
I am really happy to see you working with all the parts of the code @rachidatecs, thank you. I have some questions here and probably the most substantial one is how to keep these side-effects contained in our finite state machine code once we start using the Django admin.

src/registrar/admin.py
rachidatecs: Fantastic feedback. On it.

src/registrar/admin.py
```diff
@@ -50,13 +53,51 @@ class MyHostAdmin(AuditedAdmin):
...
+        if change:  # Check if the application is being edited
+            # Get the original application from the database
+            original_obj = models.DomainApplication.objects.get(pk=obj.pk)
+
+            if obj.status != original_obj.status and obj.status == "investigating":
```
brandonlenz, line 66: Can this be?:

```suggestion
            if obj.status != original_obj.status and obj.status == DomainApplication.INVESTIGATING:
```

src/registrar/models/domain_application.py
```diff
@@ -493,6 +496,32 @@ def _send_confirmation_email(self):
...
+        nothing.
+        """
+        if self.submitter is None or self.submitter.email is None:
+            logger.warning(
+                "Cannot send status change (in review) email,"
+                "no submitter email address."
+            )
+            return
+        try:
+            logging.info(f"In review email sent to: {self.submitter.email}")
```
neilmb, line 513: **suggestion (non-blocking)**: As above, move the log statement after the send.

src/registrar/models/domain_application.py
```diff
@@ -541,6 +570,14 @@ def approve(self):
             user=self.creator, domain=created_domain, role=UserDomainRole.Roles.ADMIN
         )
 
+    @transition(field="status", source=SUBMITTED, target=INVESTIGATING)
+    def in_review(self, updated_domain_application):
+        """Investigate an application that has been submitted."""
+
+        # When an application is moved to in review, we need to send a
```
neilmb, line 577: **comment**: I love this comment, it's so clear.

src/registrar/models/domain_application.py
```diff
@@ -541,6 +570,14 @@ def approve(self):
             user=self.creator, domain=created_domain, role=UserDomainRole.Roles.ADMIN
         )
 
+    @transition(field="status", source=SUBMITTED, target=INVESTIGATING)
+    def in_review(self, updated_domain_application):
```
neilmb, line 574: **suggestion**: This transition method is different than the other ones because it uses an additional argument and then uses that object to send the email message. Could you explain more in this method's docstring about why it is that way?

neilmb (approved):
LGTM, just a few small suggestions for doc changes and tiny code changes.

src/registrar/models/domain_application.py
rachidatecs: Noted. What's the difference between `logger.info` and `logging.info`? I see `logger = logging.getLogger(__name__)` at the top of the doc but don't really get what it's doing.

And I assume those are app logs, so I should see them in my docker console and we can see them in cloud.gov's logs somewhere?

src/registrar/models/domain_application.py
rachidatecs: Yep! I commented to this in admin.py's `save_model` and was wondering whether I should document it on this side as well.

PR #692, opened 2023-06-07
Get environment variable from a better location

neilmb:
# Fix GETGOV_PUBLIC_SITE_URL #

## 🗣 Description ##

I was getting the value of the `GETGOV_PUBLIC_SITE_URL` environment variable in the wrong way in `registrar/config/settings.py`. This should fix the `{% public_site_url ... %}` helper in our sandbox environments.

Fixes a bug that prevents `public_site_url` from working right.


SSPJ (approved):
LGTM 👍

PR #695, opened 2023-06-08
Add organization name and mailing address page for domains

### LINKED ISSUE #517: Add Organization Contact Information to Accepted Domain Section

igorkorenfeld:
## 🗣 Description ##
This PR adds the organization name and mailing address form page for a registered domain.

Closes #517 

image


src/registrar/tests/test_views.py
```diff
@@ -1315,6 +1317,23 @@ def test_domain_authorizing_official_content(self):
...
+
+    def test_domain_org_name_address_content(self):
+        """Org name and address information appears on the page."""
+        self.domain_information.organization_name = "Town of Igorville"
+        self.domain_information.save()
+        page = self.app.get(
+            reverse("domain-org-name-address", kwargs={"pk": self.domain.id})
+        )
+        self.assertContains(page, "Town of Igorville")
+
```
neilmb, line 1336: **suggestion**: I think that we could use another test for the form here. See `test_domain_user_add_form` for an example. Basically you want to load the page with `self.app`, fill out one of the form fields, click `submit` and then verify that the new value you entered is saved.

src/registrar/views/domain.py
```diff
@@ -41,6 +42,47 @@ class DomainView(DomainPermissionView):
...
+
+    def get_form_kwargs(self, *args, **kwargs):
+        """Add domain_info.organization_name instance to make a bound form."""
+        form_kwargs = super().get_form_kwargs(*args, **kwargs)
+        form_kwargs["instance"] = self.get_object().domain_info
+        return form_kwargs
+
+    def get_success_url(self):
+        """Redirect to the overview page for the domain."""
+        return reverse("domain-org-name-address", kwargs={"pk": self.object.pk})
```
neilmb, line 61: **question (non-blocking)**: I think this is probably right as it is, but I did wonder for a second if we wanted a successful save to land back on the domain overview page?

neilmb (changes requested):
This looks great. I think that the only suggestion that's blocking to me would be an additional unit test for filling out and submitting the form. The possibility of setting `verbose_name` on our database fields takes a bit of consideration either way.

neilmb:
Accessibility PR review:

## Accessibility
- ❌ New pages have been added to .pa11yci file so that they will be tested with our automated accessibility testing
  
  We still haven't figured out how to specify URLs in pa11yci for pages generated from the database. (We probably need to run some fixture code to load up the database before we run pa11y.)

    - [X] All relevant images use an img tag
    - [X] All images have alt attributes
    - [X] Multimedia is tagged. All multimedia has appropriate captioning and audio description
    - [X] Tab order is logical
    - [X] Site doesn’t timeout unexpectedly. Identify elements that may “timeout” and verify that the user can request more time
    - [X] Tables are coded properly. Tables have proper headers and column attributes
    - [X] Frames are named. All frames have a name element
    - [X] Flashing elements are compliant. Elements that flash on screen do so at a rate of less than 3 Hz
    - [X] The language for sections on the page that differ from the site language are set
    - [X] Required plugins are linked on the page
 
## Device Matrix
- [X] firefox
- [X] chrome/chromium/edge
- [X] safari
    - [X] at 480px
    - [X] at 1024px

neilmb (approved):
Changes look great, especially the form unit test. LGTM.

PR #700, opened 2023-06-08
Link and text updates

michelle-rago:
# Updated links from the app to the public site (and a few text updates) #

## 🗣 Description ##

Used the public_site_url method to replace the federalist links.

### DIFF
--- a/src/registrar/templates/401.html
+++ b/src/registrar/templates/401.html
@@ -1,5 +1,6 @@
 {% extends "base.html" %}
 {% load i18n static %}
+{% load url_helpers %}
 
 {% block title %}{% translate "Unauthorized | " %}{% endblock %}
 
@@ -25,7 +26,7 @@
         Would you like to <a href="{% url 'login' %}"> try logging in again?</a>
       </p>
       <p>
-        If you would like help with this error <a href="https://federalist-877ab29f-16f6-4f12-961c-96cf064cf070.sites.pages.cloud.gov/site/cisagov/getgov-home/contact/"> contact us </a>
+     If you'd like help with this error <a href="{% public_site_url 'contact' %}"> contact us </a>.
       </p>
 
       {% if log_identifier %}
--- a/src/registrar/templates/401.html
+++ b/src/registrar/templates/401.html
@@ -26,7 +26,7 @@
         Would you like to <a href="{% url 'login' %}"> try logging in again?</a>
       </p>
       <p>
-     If you'd like help with this error <a href="{% public_site_url 'contact' %}"> contact us </a>.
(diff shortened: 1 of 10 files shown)


michelle-rago:
@neilmb Can you make sure I've loaded url_helpers correctly? Note that some of the files already included "static." I'm not sure when that should be there. 

@igorkorenfeld If you want to remove the "todo" styling in this PR, or if you want me to do it, just let me know.

src/registrar/templates/401.html
```diff
@@ -25,7 +26,7 @@ <h2>
         Would you like to <a href="{% url 'login' %}"> try logging in again?</a>
       </p>
       <p>
-        If you would like help with this error <a href="https://federalist-877ab29f-16f6-4f12-961c-96cf064cf070.sites.pages.cloud.gov/site/cisagov/getgov-home/contact/"> contact us </a>
+     If you'd like help with this error <a href="{% public_site_url 'contact' %}"> contact us </a>.
```
neilmb, line 29: **question (non-blocking)**: I think that this probably works either way, but the new link won't have a trailing slash. Was that intentional?

src/registrar/templates/includes/footer.html
```diff
@@ -1,4 +1,5 @@
 {% load static %}
+{% load field_helpers url_helpers %}
```
neilmb, line 2: **question (non-blocking)**: It's not a big deal, but I don't see where you need the field_helpers here? With what you added, I think that this will work with just
```suggestion
{% load url_helpers %}
```

src/registrar/templates/includes/footer.html
```diff
@@ -87,19 +85,16 @@
             >
           </li>
           <li class="usa-identifier__required-links-item">
-            <a href="https://federalist-877ab29f-16f6-4f12-961c-96cf064cf070.sites.pages.cloud.gov/site/cisagov/getgov-home/privacy-policy/" class="usa-identifier__required-link usa-link"
-              >Privacy policy</a
-            >
+ <a href="{% public_site_url 'privacy-policy' %}" class="usa-identifier__required-link usa-link">Privacy policy</a>
```
neilmb, line 88: **question (non-blocking)**: As above, this has removed the trailing slash. Is that intentional?

neilmb (approved):
This looks great @michelle-rago, I just had very minor questions. And except for one of the files, the `load url_helpers` looks just right.

src/registrar/templates/401.html
michelle-rago: It wasn't intentional. Is it a better practice to include the slash?

PR #702, opened 2023-06-09
Using State Transition for Withdraw

### LINKED ISSUE #690: Domain Application withdraw view should use a FSM transition method for withdrawing
### Ticket Description

When we implemented the action for withdrawing domain applications, we manipulated the application's `status` field directly https://github.com/cisagov/getgov/blob/main/src/registrar/views/application.py#L505. So that we can handle side-effects cleanly (for example sending a "Your application has been withdrawn" email message, we should change this code to use a new `withdraw()` method on the `DomainApplication` model.

There is an example in that same file for `approve()`. We would want applications to only be able to be withdrawn before they are approved.

### Issue Type

development

### Link dependent issues

_No response_

kewlguy781:
# Withdraw State Transition added #

## 🗣 Description ##

* Added withdraw transition
* Edited the withdraw process to use state transition instead of update the field manually.
* Ran Test to ensure nothing else break

This will change the way withdraw process in the backend. Before it was set as update field, but we have seen some places where approved were be able to withdraw. Using state transition in place for withdraw will protect it from making an accidental or mess up when withdrawing an application that is not in "submitted/review" state
Resolves and Closes #690

### DIFF
--- a/src/registrar/models/domain_application.py
+++ b/src/registrar/models/domain_application.py
@@ -541,6 +541,10 @@ class DomainApplication(TimeStampedModel):
             user=self.creator, domain=created_domain, role=UserDomainRole.Roles.ADMIN
         )
 
+    @transition(field="status", source=[SUBMITTED, INVESTIGATING], target=WITHDRAWN)
+    def withdraw(self):
+        """Withdraw an application that has been submitted."""
+
     # ## Form policies ###
     #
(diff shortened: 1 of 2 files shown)


neilmb:
Follow-on ticket to use the new transition method to send email confirmation: #706

brandonlenz (approved):

PR #709, opened 2023-06-12
Add dev sandbox for Katherine

abroddrick:
#  #

## 🗣 Description ##
Adds Katherines (ko) sandbox


loganmeetsworld:
We need to also add to here: https://github.com/cisagov/getgov/blob/main/.github/workflows/deploy-sandbox.yaml#L12-L19

loganmeetsworld (approved):
One comment about spacing but otherwise lgtm!

PR #718, opened 2023-06-13
Send email to submitter when application status changes to "investigating" - fsm_amin vs

rachidatecs:
## 🗣 Description ##

- Implement 'investigating' email templates
- Trigger email when application status changes to "investigating"
- Use package fsm_admin to make admin classes aware of the FSM
- Edit settings to fix fsm_admin and django version conflicts
- Componentize completed_application for unit tests
- Componentize application summary in email templates
- Unit test

Resolves [472](https://github.com/cisagov/getgov/issues/472)


neilmb:
To me this is a major simplification worth having. I haven't pulled the branch and run it yet, so I'd love to see a screenshot of what those status-change buttons look like on the Django admin. My suggestion would be to make this change as a PR after the other one so that it just includes the new libraries and the changes in `admin.py`.

rachidatecs:
Here's an example for an application with status = investigating

Screen Shot 2023-06-13 at 5 32 52 PM

rachidatecs:
Replaced with PR 723: https://github.com/cisagov/getgov/pull/723

PR #719, opened 2023-06-13
Sketch test cases for Domain integration

### LINKED ISSUE #492: Draft test cases for registry integration
Draft a set of test cases for automated testing, perhaps using Cucumber. These can use red, green, refactor, e.g. they can be failing at the time they are committed and make use of `@skip`.

These should illuminate the "shape" of the interface which we'll use to interact with the registry.

- [x] Activating a domain
- [x] Placing and removing client hold (suspending)
- [x] Placing and removing client X prohibited (locking)
- [x] Deleting a domain
- [x] CRUD actions on contacts
- [x] CRUD actions on nameservers
- [ ] CRUD actions on DNSSEC

SSPJ:
## 🗣 Description ##

An example set of test cases to test a broad but non-exhaustive set of registrar-registry functionality.

This PR does not include the implementations, only the outline.

This also includes an illustration of how action buttons could be integrated into the admin portal.

Closes #492


SSPJ:
Gonna tag @rachidatecs on here as well for awareness of the admin experiment-y thing.

src/registrar/admin.py
```diff
@@ -50,13 +50,40 @@ class MyHostAdmin(AuditedAdmin):
...
 
+class DomainAdmin(AuditedAdmin):
+
+    """Custom domain admin class to add extra buttons."""
+
+    change_form_template = "django/admin/domain_change_form.html"
+    readonly_fields = ["state"]
+
+    def response_change(self, request, obj):
+        if "_place_client_hold" in request.POST:
```
neilmb, line 61: **comment (non-blocking)**: Because this is a magic string that has to match with one in a "distant" template file, maybe a comment on where this value comes from?

src/registrar/templates/django/admin/domain_change_form.html
```diff
@@ -0,0 +1,8 @@
+{% extends 'admin/change_form.html' %}
+
+{% block field_sets %}
+    <div class="submit-row">
+        <input type="submit" value="Place hold" name="_place_client_hold">
```
neilmb, line 5: **question (non-blocking)**: This is a huge can of worms because of how styling is done on the Django admin, but I initially expected this to have the `usa-button` class. Does it look right if we start using USWDS components inside of admin?

neilmb (approved):
These test cases look very well-organized and thoughtfully specified. 

One thing I wonder is how @PaulKuykendall and @kitsushadow can use this breakdown to specify tasks and work items on the backlog to make the `Domain` implementation work and pass these tests.

src/registrar/admin.py
abroddrick: +1 to that.

src/registrar/tests/test_models_domain.py
```diff
@@ -1,54 +1,487 @@
...
+        """
+        # implementation note: this requires seeing what happens when these are actually
+        # sent like this, and then implementing appropriate mocks for any errors the
+        # registry normally sends in this case
+        raise
+
+    @skip("not implemented yet")
+    def test_update_is_unsuccessful(self):
+        """
+        Scenario: An update to the security contact is unsuccessful
```
abroddrick, line 316:
```suggestion
        Scenario: An update to the dns data is unsuccessful
```

src/registrar/tests/test_models_domain.py
```diff
@@ -1,54 +1,487 @@
...
+        """
+        Scenario: Analyst tries to lock a domain twice
+            Given `CLIENT_*_PROHIBITED` is already in the domain's statuses
+            When `domain.lock()` is called
+            Then Domain returns normally (without error)
+        """
+        raise
+
+    @skip("not implemented yet")
+    def test_analyst_removes_client_hold(self):
```
abroddrick, line 416:
```suggestion
    def test_analyst_removes_lock(self):
```

src/registrar/tests/test_models_domain.py
```diff
@@ -1,54 +1,487 @@
...
+            Given `CLIENT_*_PROHIBITED` is in the domain's statuses
+            When `domain.unlock()` is called
+            Then `CLIENT_DELETE_PROHIBITED` is no longer in the domain's statuses
+            And `CLIENT_TRANSFER_PROHIBITED` is no longer in the domain's statuses
+            And `CLIENT_UPDATE_PROHIBITED` is no longer in the domain's statuses
+        """
+        raise
+
+    @skip("not implemented yet")
+    def test_analyst_removes_client_hold_idempotent(self):
```
abroddrick, line 428:
```suggestion
    def test_analyst_removes_lock_idempotent(self):
```

src/registrar/tests/test_models_domain.py
```diff
@@ -1,54 +1,487 @@
...
+            Given the analyst is logged in
+            And a domain exists in the registry
+        """
+        pass
+
+    @skip("not implemented yet")
+    def test_analyst_deletes_domain(self):
+        """
+        Scenario: Analyst permanently deletes a domain
+            When `domain.delete()` is called
```
abroddrick, line 462: Just curious why do we just change the status to deleted and not remove it from the user and our tables? I assume it's so it can be brought back or to retain a record, but just curious.

abroddrick (approved):
Most of my comments are just questions and a couple quick fixes. Nothing to block this for

src/registrar/admin.py
SSPJ: Fixed.

src/registrar/templates/django/admin/domain_change_form.html
SSPJ: I'll defer this to the new team. 😄

src/registrar/tests/test_models_domain.py
```diff
@@ -1,54 +1,487 @@
...
 
-class TestDomain(TestCase):
-    def test_empty_create_fails(self):
+class TestDomainCreation(TestCase):
+    """Rule: An approved domain application must result in a domain"""
+
+    def setUp(self):
+        """
+        Background:
+            Given that a valid domain application exists
```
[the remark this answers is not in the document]
SSPJ: That is true. However, testing the FSM machine for domain status is likely outside the scope of this test, so it is more a case of "you probably can't get a valid application to use unless you put it into the submitted state".

src/registrar/tests/test_models_domain.py
```diff
@@ -1,54 +1,487 @@
...
+        """
+        raise
+
+    @skip("not implemented yet")
+    def test_accessing_domain_properties_creates_domain_in_registry(self):
+        """
+        Scenario: A registrant checks the status of a newly approved domain
+            Given that no domain object exists in the registry
+            When `domain.is_active()` is called
+            Then Domain sends `commands.CreateDomain` to the registry
```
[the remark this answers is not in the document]
SSPJ: I can see where this is causing confusion. `domain.is_active()` is merely an example of a property to access: this could be tested with any property.

`domain.is_active()` should really only result in "create domain" when the domain doesn't already exist in the registry (regardless of whether it is active -- although if it doesn't exist, of course it is not active)

src/registrar/tests/test_models_domain.py
```diff
@@ -1,54 +1,487 @@
...
+        Background:
+            Given the registrant is logged in
+            And the registrant is the admin on a domain
+        """
+        pass
+
+    @skip("not implemented yet")
+    def test_no_security_email(self):
+        """
+        Scenario: Registrant declines to add a security contact email
```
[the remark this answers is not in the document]
SSPJ: This is the default state, no user action required. I'll update to make that clearer.

src/registrar/tests/test_models_domain.py
```diff
@@ -1,54 +1,487 @@
...
+            When the DomainApplication transitions to approved
+            Then a Domain exists in the database with the same `name`
+            But a domain object does not exist in the registry
+        """
+        raise
+
+    @skip("not implemented yet")
+    def test_accessing_domain_properties_creates_domain_in_registry(self):
+        """
+        Scenario: A registrant checks the status of a newly approved domain
```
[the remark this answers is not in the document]
abroddrick: This one wasn't yes or no, re-reading I see that the or gets kind of swallowed in my wordy-ness. Is it 1)using is_active or 2)  accessing domain.state that should be the triggering event.

src/registrar/tests/test_models_domain.py
```diff
@@ -1,54 +1,487 @@
...
+                created contact of type 'security'
+        """
+        raise
+
+    @skip("not implemented yet")
+    def test_security_email_is_idempotent(self):
+        """
+        Scenario: Registrant adds a security contact email twice, due to a UI glitch
+            When `commands.CreateContact` and `commands.UpdateDomain` are sent
+                to the registry twice with identical data
```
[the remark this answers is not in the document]
SSPJ: I will refer you to the guide that the registry vendor just posted, but afaik, you can have up to 2 security contacts in the registry.

We just don't support that in the registrar UI.

src/registrar/tests/test_models_domain.py
```diff
@@ -1,54 +1,487 @@
...
+
+    @skip("not implemented yet")
+    def test_user_adds_two_nameservers(self):
+        """
+        Scenario: Registrant adds 2 or more nameservers, thereby activating the domain
+            Given the domain has zero nameservers
+            When `domain.nameservers` is set to an array of length 2
+            Then `commands.CreateHost` and `commands.UpdateDomain` is sent
+                to the registry
+            And `domain.is_active` returns True
```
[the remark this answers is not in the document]
SSPJ: No, it is unrelated. The intent is more to underline the fact that once 2 valid nameservers are set, the domain should be live. This could be tested here, but may also be better split into its own test of `is_active`.

src/registrar/tests/test_models_domain.py
```diff
@@ -1,54 +1,487 @@
...
+        """
+        raise
+
+    @skip("not implemented yet")
+    def test_user_adds_too_many_nameservers(self):
+        """
+        Scenario: Registrant adds 14 or more nameservers
```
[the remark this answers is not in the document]
SSPJ: 13 is the max they are allowed to have, per the registry.

src/registrar/tests/test_models_domain.py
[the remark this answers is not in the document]
abroddrick: This makes so much more sense, just check that any given property being accessed causes the create. Thank you for clarifying.

src/registrar/tests/test_models_domain.py
SSPJ: Basically that. To keep a record for future use and to help CISA meet their document retention policies.

PR #724, opened 2023-06-15
Add domain caching layer and tests

### LINKED ISSUE #636: Create and test Domain caching layer
A/C: information about a domain is fetched from the registry via EPP on first access and retrieved from the Domain object on each subsequent access

abroddrick:
#  #

## 🗣 Description ##

-adds caching layer from Seamus' branch
-adds tests for caching
-implementation of various models/domain function out of scope as they will be completed in one of the subsequent PRs. 

closes #636 
-caching is used so as to avoid having to call epp every time for getters as the registry should not be changing data with out the registrar setting it, thus we can assume the last thing we cached hasn't changed until will send a set command.


src/registrar/models/domain.py
```diff
@@ -300,3 +346,222 @@ def __str__(self) -> str:
...
+                else:
+                    raise e
+
+    def _update_or_create_host(self, host):
+        raise NotImplementedError()
+
+    def _delete_host(self, host):
+        raise NotImplementedError()
+
+    def _fetch_cache(self, fetch_hosts=False, fetch_contacts=False):
```
neilmb, line 458: **comment (non-blocking)**: This is a very large method. Is it possible that we might break it down into separate `_fetch_info`, `_fetch_contacts`, `_fetch_hosts` methods that get called from in here?

src/registrar/tests/test_models_domain.py
```diff
@@ -22,6 +59,70 @@ def test_minimal_create(self):
...
+        }
+
+        # this can be changed when the getter for contacts is implemented
+        domain._get_property("contacts")
+
+        # check domain info is still correct and not overridden
+        self.assertEqual(domain._cache["auth_info"], self.mockDataInfoDomain.auth_info)
+        self.assertEqual(domain._cache["cr_date"], self.mockDataInfoDomain.cr_date)
+
+        # check contacts
```
neilmb, line 118: **comment (non-blocking)**: This seems non-obvious that there are two "contacts" things under different keys. Could we have more explanation in a comment here about why, or refer to a docstring about it?

neilmb (approved):
Lots of comments, none of them too important.  

I can recognize that this is a very deep part of the registry interaction code and I think that the tests are great, but I also recognize that they don't have a way to test interaction effects beyond the mocked epplib calls. 

Is there a way that we can start considering some separate automated integration tests with our code and an EPP test environment?

src/registrar/tests/test_models_domain.py
abroddrick: Good point there for sure be a docstring on it and comments! i will do that

rachidatecs (approved):

PR #728, opened 2023-06-15
Update attribute for alternate domains

igorkorenfeld:
## 🗣 Description ##

There is a bug on the dotgov_domain page in the application form that is causing a server error. This updates the attribute so that the correct data is pulled in.

image

### DIFF
--- a/src/registrar/forms/application_wizard.py
+++ b/src/registrar/forms/application_wizard.py
@@ -498,7 +498,7 @@ class BaseAlternativeDomainFormSet(RegistrarFormSet):
 
     @classmethod
     def on_fetch(cls, query):
-        return [{"alternative_domain": Domain.sld(domain.name)} for domain in query]
+        return [{"alternative_domain": Domain.sld(domain.website)} for domain in query]
 
     @classmethod
     def from_database(cls, obj):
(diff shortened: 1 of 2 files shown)


neilmb (approved):
Good catch. 

I'm fine merging the fix as-is, but is there a simple unit test that we could/should have that could have caught this? If I remember correctly it's just loading that page in the existing application flow? I just wonder why we didn't see it in our existing test suite.

igorkorenfeld:
Looking at it closer, the test suite didn't catch the issue because the current set of tests only creates a new application, but this bug came up when going back to edit an existing application. 

When trying to add a test to go through an existing application I ran into a couple of issues. I used our `completed_application` function to set up an existing application to test. However, adding a `click` on the "Edit" button for this new application seemed to generate a brand new application instead of getting into editing an existing application, so I wasn't able to step through it. 

I tried an alternative approach of using a `get` to go through the existing application. However the `get` results in a 302 redirect, and attempting to add a `follow` option to the `get` results in an error in Django:

> django.core.exceptions.ImproperlyConfigured: CSRF_USE_SESSIONS is enabled, but request.session is not set. SessionMiddleware must appear before CsrfViewMiddleware in MIDDLEWARE.

Adding  `self.app.set_cookie(settings.SESSION_COOKIE_NAME, session_id)`  doesn’t seem to resolve the issue.

@abroddrick I'll leave it up to you to decide if you all would like to merge this PR as is, or add a test as part of it (or close it and start a new one :))

abroddrick:
> @abroddrick I'll leave it up to you to decide if you all would like to merge this PR as is, or add a test as part of it (or close it and start a new one :))
 I'm good with this being merged as is, feel free to do so; otherwise I will merge it at the end of the day

PR #755, opened 2023-06-28
Fix bug in approved domain notification

rachidatecs:
## 🗣 Description ##

Fix typo in argument passed to the email sending method for approved domains. This bug is triggered when an email send is triggered in fixtures loading, not django admin.


abroddrick (approved):
One line code changes are the best.

PR #760, opened 2023-06-30
Django admin MVP implementation: views and permissions

rachidatecs:
## 🗣 Description ##

- Create fixtures for CISA analysts and set permissions to view logs, contacts and domains, edit domain applications
- Add columns to domain applications list
- Add filters to domain applications list
- Enable Search on domain applications, contacts and domains (with search helper text)
- Implement descriptive sub-header on domain applications, domains and contacts

Resolves [680](https://github.com/cisagov/getgov/issues/680), [753](https://github.com/cisagov/getgov/issues/753) and [754](https://github.com/cisagov/getgov/issues/754)


docs/django-admin/roles.md
```diff
@@ -0,0 +1,20 @@
...
+## CISA analyst
+
+### Basic permission level
+
+Staff
+
+### Additional group permissions
+
+auditlog | log entry | can view log entry
+registrar | contact | can view contact
```
abroddrick, line 18: Since they are going to need to see users too, do you want to add that here or will that be the next PR?

src/registrar/admin.py
```diff
@@ -24,6 +24,40 @@ def history_view(self, request, object_id, extra_context=None):
...
+
+    def changelist_view(self, request, extra_context=None):
+        if extra_context is None:
+            extra_context = {}
+        # Get the filtered values
+        filters = self.get_filters(request)
+        # Pass the filtered values to the template context
+        extra_context["filters"] = filters
+        extra_context["search_query"] = request.GET.get(
+            "q", ""
```
abroddrick, line 39: What does the empty "" indicate?

src/registrar/config/settings.py
```diff
@@ -86,6 +92,10 @@
     # the "user" model! *\o/*
     "django.contrib.auth",
     # generic interface for Django models
+    "auditlog",
+    # library to simplify form templating
+    # it needs to be listed before django.contrib.contenttypes
```
abroddrick, line 97: wait this comment confuses me. It sounds like this is referring to django.contrib.contenttypes but then it says it needs to be listed before django.contrib.contenttypes. Maybe this is meant to be a comment for auditlog?

src/registrar/config/settings.py
```diff
@@ -86,6 +92,10 @@
     # the "user" model! *\o/*
     "django.contrib.auth",
     # generic interface for Django models
```
abroddrick, line 94: this comment goes with the contenttypes import

src/registrar/config/settings.py
```diff
@@ -99,15 +109,11 @@
     # application used for integrating with Login.gov
     "djangooidc",
     # audit logging of changes to models
```
abroddrick, line 111: move this to go above the auditlog import

src/registrar/fixtures.py
```diff
@@ -73,7 +99,58 @@ def load(cls):
                 logger.debug("User object created for %s" % admin["first_name"])
             except Exception as e:
                 logger.warning(e)
-        logger.debug("All users loaded.")
+        logger.debug("All superusers loaded.")
```
abroddrick, line 102: I would make this info

src/registrar/fixtures.py
```diff
@@ -73,7 +99,58 @@ def load(cls):
...
+                        + ", ".join(permission_list)
+                        + " added for user "
+                        + staff["first_name"]
+                    )
+
+                user.save()
+                logger.debug("User object created for %s" % staff["first_name"])
+            except Exception as e:
+                logger.warning(e)
+        logger.debug("All CISA analysts (staff) loaded.")
```
abroddrick, line 153: same as above, I would make this info, just given the log that stays it is going to load is also info.

src/registrar/admin.py
rachidatecs: In this context, it means that if there is no search query parameter in the GET parameters of the request, the value of extra_context["search_query"] will be an empty string.

docs/django-admin/roles.md
rachidatecs: I think it's better left for the next PR, alongside the code that implements it?

src/registrar/fixtures.py
```diff
@@ -73,7 +99,58 @@ def load(cls):
...
+                user.is_staff = True
+                user.is_active = True
+
+                for permission in cls.STAFF_PERMISSIONS:
+                    app_label = permission["app_label"]
+                    model_name = permission["model"]
+                    permissions = permission["permissions"]
+
+                    # Retrieve the content type for the app and model
+                    content_type = ContentType.objects.get(
```
abroddrick, line 122: I believe you said accessing the contenttype was throwing an error on fixtures on the first couple loads, and thus when (if) we fix the fixture bug it won't work correctly. Is this the line you were referring to? (I am not seeing the same error in the console log though).

src/registrar/tests/test_admin.py
```diff
@@ -162,3 +178,54 @@ def test_save_model_sends_approved_email(self):
...
         application.delete()
+
+    def test_changelist_view(self):
+        # Have to get creative to get past linter
+        p = "adminpassword"
+        self.client.login(username="admin", password=p)
+
+        # Make the request using the Client class
+        # which handles CSRF
+        # Follow=True handles the redirect
```
abroddrick, line 189: ❤️ to these helpful comments!

src/registrar/fixtures.py
```diff
@@ -56,9 +59,32 @@ class UserFixture:
         },
     ]
 
+    STAFF = [
+        {
+            "username": "319c490d-453b-43d9-bc4d-7d6cd8ff6844",
+            "first_name": "Rachid-Analyst",
+            "last_name": "Mrad-Analyst",
+        },
```
abroddrick, line 67:
```suggestion
        },
        {
            "username": "b6a15987-5c88-4e26-8de2-ca71a0bdb2cd",
            "first_name": "Alysia-Analyst",
            "last_name": "Alysia-Analyst",
        },
```
If you dont mind adding this to your PR, otherwise I have the change saved and can make a PR to add myself

abroddrick (approved):
See some minor comments and suggestions. It looks great!

src/registrar/fixtures.py
rachidatecs: Happy to add to my PR :)

src/registrar/admin.py
```diff
@@ -24,6 +24,40 @@ def history_view(self, request, object_id, extra_context=None):
         )
 
 
+class ListHeaderAdmin(AuditedAdmin):
```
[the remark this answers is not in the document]
rachidatecs: This is just at the point where it's non-trivial enough to warrant a follow-up ticket (and no, this has not been captured anywhere yet)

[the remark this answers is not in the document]
rachidatecs: I may be on to a solution for this

src/registrar/fixtures.py
rachidatecs: Looks like it to me

Screen Shot 2023-07-06 at 10 39 28 AM

abroddrick (approved):
image

Changes look good. I did note that a similar translation change probably should be applied to the organization type, BUT since the org type is really long I think we already discussed maybe shortening it for ease of reading. So just noting this here, but probably best to implement a change after asking design what to shorten it to.

PR #782, opened 2023-07-07
Expose users in  CISA Analysts' django admin while hiding uuids

rachidatecs:
## 🗣 Description ##

- Expose users table for staff role
- Hide uuids in list and detail views
- Unit tests to make sure uuids remain hidden
- Clean up admin unit tests and move some common methods to common.py

Resolves [578](https://github.com/cisagov/getgov/issues/758)

### DIFF
--- a/src/registrar/admin.py
+++ b/src/registrar/admin.py
@@ -103,9 +103,9 @@ class MyUserAdmin(BaseUserAdmin):
         if not request.user.is_superuser:
             # Customize the list display for staff users
             return ("email", "first_name", "last_name", "is_staff", "is_superuser")
-        else:
-            # Use the default list display for non-staff users
-            return super().get_list_display(request)
+
+        # Use the default list display for non-staff users
+        return super().get_list_display(request)
 
     def get_fieldsets(self, request, obj=None):
         if not request.user.is_superuser:
--- a/src/registrar/admin.py
+++ b/src/registrar/admin.py
@@ -1,6 +1,6 @@
 import logging
 from django.contrib import admin, messages
-from django.contrib.auth.admin import UserAdmin
+from django.contrib.auth.admin import UserAdmin as BaseUserAdmin
 from django.contrib.contenttypes.models import ContentType
 from django.http.response import HttpResponseRedirect
 from django.urls import reverse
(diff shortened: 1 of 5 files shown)


src/registrar/admin.py
```diff
@@ -93,12 +93,29 @@ class UserContactInline(admin.StackedInline):
...
 
     """Custom user admin class to use our inlines."""
 
     inlines = [UserContactInline]
 
+    def get_list_display(self, request):
+        if not request.user.is_superuser:
+            # Customize the list display for staff users
+            return ("email", "first_name", "last_name", "is_staff", "is_superuser")
+        else:
```
abroddrick, line 106: technically this else isn't needed because the if being true results in the return being called. You could leave it out, but that can also be a code style thing

src/registrar/tests/common.py
```diff
@@ -88,6 +88,37 @@ def send_email(self, *args, **kwargs):
...
+    user_kwargs = dict(
+        id=4,
+        first_name="Rachid",
+        last_name="Mrad",
+    )
+    mock_user, _ = User.objects.get_or_create(**user_kwargs)
+    return mock_user
+
+
+def create_superuser(self):
```
abroddrick, line 102: self is never used here or below in create_user, I think it can be removed unless there is some other reason it was added. Also normally if you have a utility function, I wouldn't pass "self" because "self" is like "this" in js and usually is meant to convey a special meaning (ie pass the class this function resides in). Since this isn't a class, I would have my argument describe what it's for.

src/registrar/tests/test_admin.py
```diff
@@ -240,7 +238,44 @@ def test_get_filters(self):
...
+
+
+class MyUserAdminTest(TestCase):
+    def setUp(self):
+        admin_site = AdminSite()
+        self.admin = MyUserAdmin(model=get_user_model(), admin_site=admin_site)
+
+    def test_list_display_without_username(self):
+        request = self.client.request().wsgi_request
+        request.user = create_user(self)
```
abroddrick, line 252: see comment above, not sure if self is needed here

src/registrar/tests/test_admin.py
```diff
@@ -240,7 +238,44 @@ def test_get_filters(self):
...
+            "is_staff",
+            "is_superuser",
+        )
+
+        self.assertEqual(list_display, expected_list_display)
+        self.assertNotIn("username", list_display)
+
+    def test_get_fieldsets_superuser(self):
+        request = self.client.request().wsgi_request
+        request.user = create_superuser(self)
```
abroddrick, line 268: self may not be needed here see comment on definition.

src/registrar/tests/test_admin.py
```diff
@@ -240,7 +238,44 @@ def test_get_filters(self):
...
+        fieldsets = self.admin.get_fieldsets(request)
+        expected_fieldsets = super(MyUserAdmin, self.admin).get_fieldsets(request)
+        self.assertEqual(fieldsets, expected_fieldsets)
+
+    def test_get_fieldsets_non_superuser(self):
+        request = self.client.request().wsgi_request
+        request.user = create_user(self)
```
abroddrick, line 275: self may not be needed here see comment on definition.

abroddrick (approved):
Looks good. Love the inline comments and test coverage seems perfect for this. A couple minor comments but non-blocking

src/registrar/admin.py
```diff
@@ -1,6 +1,6 @@
 import logging
 from django.contrib import admin, messages
-from django.contrib.auth.admin import UserAdmin
+from django.contrib.auth.admin import UserAdmin as BaseUserAdmin
```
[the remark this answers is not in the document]
rachidatecs: I feel that with foundational class like UserAdmin, the alias clearly distinguishes the base and customized classes and leaves access to the original implementation if needed.

PR #819, opened 2023-07-28
Session bug fix

### LINKED ISSUE #394: Bug: Starting a new application takes user back to existing application
When I try to start a 2nd new application the button on the dashboard takes me into an existing application.

To replicate
1. Click "Start a new domain request" on the dashboard
2. Fill out the first few pages of the form to make it distinct 
3. Navigate back to home 
4. Click "Start a new domain request" on the dashboard
5. Check the first few page you filled out before to confirm this is the same application

Additionally, other bugs associated with this request:

AC:

- Update functionality to allow for multiple applications in the same session

See: https://gsa-tts.slack.com/archives/C03QM0JGSQG/p1675441631322649

A note for anyone picking this up to fix it: this is more challenging than it first appears. The current application is being saved in state to allow a user to navigate between the pages of the application without losing progress. We don't have any ability to know when they have "exited" the application. (We should avoid any JavaScript hacks to try to "guess" since all of these approaches are buggy and rely on the user having JavaScript enabled/working.) We could *very easily* reset state each time they visit the "front page" of an application (as they do when clicking "apply"), however, we would then need to ensure they had some way of getting back to previously saved applications (since the "front page" will now always show them a fresh blank application).

The code to fix this bug might be pretty simple, but there are plenty of ramifications to think through. 🙂

abroddrick:
# Resolves #394  #

## 🗣 Description ##
-after editing an application, a new application can now be created
-the previous application's data should not appear on any pages

When a user had already been editing a previous application and returned to the home page, clicked start new application, then a bug showed where a new application was not actually created and instead the previous application was shown instead.

Note to PR reviewer, testing this was getting rather complicated with a unit test, so the test was skipped for time saving reasons. I think this one is fine for just manual testing and we can come back to upping unit test coverage later. Feel free to disagree and Request Changes

### DIFF
--- a/src/registrar/tests/test_views.py
+++ b/src/registrar/tests/test_views.py
@@ -150,7 +150,6 @@ class DomainApplicationTests(TestWithUser, WebTest):
         type_result = type_page.form.submit()
         # should see results in db
         application = DomainApplication.objects.get()  # there's only one
-
         self.assertEqual(application.organization_type, "federal")
         # the post request should return a redirect to the next form in
         # the application
@@ -168,7 +167,6 @@ class DomainApplicationTests(TestWithUser, WebTest):
 
         # test next button
         self.app.set_cookie(settings.SESSION_COOKIE_NAME, session_id)
-
         federal_result = federal_form.submit()
         # validate that data from this step are being saved
         application = DomainApplication.objects.get()  # there's only one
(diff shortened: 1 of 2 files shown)


src/registrar/tests/test_views.py
```diff
@@ -485,6 +490,7 @@ def test_application_form_started_allsteps(self):
         # type_page = home_page.click("Edit")
 
         session_id = self.app.cookies[settings.SESSION_COOKIE_NAME]
+        session_wizard = self.app.session["wizard_application"]
```
rachidatecs, line 493: This guy seems to be causing linter to fail, since it's been added and is not used.

abroddrick: Yep already fixed that, was using the sesison wizard before to fix the testing issue then found it wasn't needed

abroddrick:
> after we implement the ideal user-flow that design has put together, would you agree?

yes, this is meant to clear only when a user is going to a brand new application and should allow users to pull up and switch between multiple in progress applications without a problem. I see no conflict with the user flow that design has suggested.

rachidatecs (approved):

PR #831, opened 2023-08-03
823 - Explore extending django components with uswds markup

### LINKED ISSUE #823: Investigate overriding markup for django-admin components
### Issue Description

As we are overriding django-admin styles anyway, it makes sense to look at the components at the same time. Thus an investigation should be done to determine how much effort it would be to change the most common components we use so that they match USWDS (such as checkbox). 

AC:
-decision is made on whether or not override components in django admin for mvp to use USWDS 
-if the decision is yes, more tickets should be made for which components will be included in mvp
-if the decision is no, the decision needs to be documented in this ticket as a comment and a ticket with why and any related test branch that was made. If applicable make a new tickets that are marked for post mvp.

_No response_

### Issue Links

Relates to: #795

rachidatecs:
## 🗣 Description ##

- Test using uswds chackbox markup
- Test using uswds table markup

Resolves #823 
This is an exercise to scope-out overriding django admin's markup to use USWDS markup and CSS classes.
 
## Findings

- When possible: this should be done through extending the admin templates instead of overwriting them as that would come with a hefty maintenance cost.
- Implementing uswds tables worked out fine. It's an obvious improvement over djando's styles
- Implementing uswds checkboxes was problematic. The uswds generated markup stands out for being too large in the django admin context, and the sr-only classes on the labels needed removing since uswds generated markup pretty much == the labels (a workaround is to add sr-only span child elements). 
- I recommend considering merging the tables implementation if it looks good to designers after merging the CSS theming with this work, shelving other components and especially form components for later consideration.

Screen Shot 2023-08-03 at 12 52 41 PM

Screen Shot 2023-08-03 at 12 58 09 PM

## Decision

Use usa-table usa-table--borderless usa-table--striped on change_list_results

### DIFF
--- a/src/registrar/templates/admin/app_list.html
+++ b/src/registrar/templates/admin/app_list.html
@@ -3,7 +3,7 @@
 {% if app_list %}
   {% for app in app_list %}
     <div class="app-{{ app.app_label }} module{% if app.app_url in request.path|urlencode %} current-app{% endif %}">
-      <table>
+      <table class="usa-table">
         <caption>
           <a href="{{ app.app_url }}" class="section" title="{% blocktranslate with name=app.name %}Models in the {{ name }} application{% endblocktranslate %}">{{ app.name }}</a>
         </caption>
--- a/src/registrar/templates/admin/change_list_results.html
+++ b/src/registrar/templates/admin/change_list_results.html
@@ -20,10 +20,10 @@ Load our custom filters to extract info from the django generated markup.
 {# .gov - hardcode the select all checkbox #}
 <th scope="col" class="action-checkbox-column" title="Toggle all">
     <div class="text">
-        <span>
-            <input type="checkbox" name="_selected_action" id="action-toggle">
-            <label for="action-toggle" class="usa-sr-only">Toggle all</label>
-        </span>
+        <div class="usa-checkbox">
+            <input type="checkbox" name="_selected_action" id="action-toggle" class="usa-checkbox__input">
+            <label for="action-toggle" class="usa-checkbox__label"><span class="usa-sr-only">Toggle all</span></label>
+        </div>
     </div>
     <div class="clear"></div>
 </th>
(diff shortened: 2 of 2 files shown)


abroddrick:
To confirm, are you suggested a ticket be made for us to revisit updating these other components? Any idea if it were to be implemented the time it would take?

rachidatecs:
My findings were too nuanced. I recommend abandoning this and revisiting in the future if we know django admin will survive beyond MVP. Even then, it may not be worth it.

gabydisarli:
The tables implementation looked good if that's still on the table. If it isn't then yes, let's abandon this and revisit in the future.

zandercymatics:
@rachidatecs I currently do not have access to /admin, so I'll make some notes on what I can access for now... I've noted some minor visual things, but disregard the below if we decide not to go with this.

### Checkboxes

- The 'select all' checkbox is not inline with the rest of the checkboxes. This may be resolved with removing 'toggle all' (see below). I think that would make it appear more visually consistent with the rest of the elements.  
- If its not caused by title="Toggle all", it could be caused by subtle styling differences between the (children of) <thead> and <body> elements - such as margin/padding or different display types. From experience, its usually not noticeable until you insert an element of a specific size (such as a large checkbox). In that case, the best solution would be to target that specific cell and make alterations as needed. See below:
- image

### Toggle All

- The 'toggle all' text may be better represented as a [tooltip](https://designsystem.digital.gov/components/tooltip/), or removed. 
- The labels for the select boxes seem redundant, as that information is on the preceding column. The length of those labels appear to be causing some odd spacing issues, causing the information on that column to not line up with the text on other columns. I would recommend removing (or hiding) the label. If it is necessary to have it, a tooltip would be a better option, or I'd potentially recommend reducing the font size and background color to match the other cols.  
- image

### Other
Code-wise, it looks good. The element slicing bit finicky but it seems to potentially be the best option for customization in this scenario. I agree that maintainability would not be fun if we went forward with this and needed to replicate this exact solution across many pages. In that scenario, I would say this only makes sense to do if we can find a streamlined way of doing it.

You may be aware of this given you did a lot of digging on this issue, but if we do decide to go forward with this styling elsewhere in the app, it looks like we have some options for custom table implementations (with proxy models):
https://hakibenita.com/how-to-turn-django-admin-into-a-lightweight-dashboard
https://docs.djangoproject.com/en/1.10/topics/db/models/#proxy-models

rachidatecs:
Great comments @zandercymatics, those links you shared are gold. @gabydisarli, tables are the one markup override that we can proceed with since it's a 'minimally invasive' implementation and does not come with USWDS margin and padding side effects that'll look out of place. Want to collab for 10 minutes after I get PR 826 approved and merged?

abroddrick:
just pulled and recompiled with npx gulp compile and I get this 
image. Boxes are all wonky

abroddrick (dismissed):
Looks good to me, but please don't merge or move forward until @gabydisarli or @Katherine-Osos can approve the design side of things. 

EDIT: ignore this I meant to add this to Zander's PR

gabydisarli:
@rachidatecs sorry I missed your earlier comment. Did you still want to sync? I'm happy to approve and don't want to be a blocker, just trying to see where your at.

abroddrick:
This PR is being marked as closed and not merged as it was an investigation into the difficulty level of configuring the django admin components with USWDS markup. The result is that it will be a large lift and take too much time to implement at the moment. See findings above for more information.

abroddrick:
This PR is being marked as closed and not merged as it was an investigation into the difficulty level of configuring the django admin components with USWDS markup. The result is that it will be a large lift and take too much time to implement at the moment. See findings above for more information.

PR #833, opened 2023-08-03
Updated README.md to include adding an analyst

zandercymatics:
#  #

## 🗣 Description ##
The current README.md does not detail the steps necessary to create an 'Analyst' user, which are largely similar to creating an admin user. I've detailed this process, and also made a small alteration to 'ADMINS' so that it stands out a bit better on the markdown

To better document internal processes

### DIFF
--- a/docs/developer/README.md
+++ b/docs/developer/README.md
@@ -66,7 +66,7 @@ The endpoint /admin can be used to view and manage site content, including but n
 1. Login via login.gov
 2. Go to the home page and make sure you can see the part where you can submit an application
 3. Go to /admin and it will tell you that UUID is not authorized, copy that UUID for use in 4
-4. in src/registrar/fixtures.py add to the ADMINS list in that file by adding your UUID as your username along with your first and last name. See below:
+4. in src/registrar/fixtures.py add to the `ADMINS` list in that file by adding your UUID as your username along with your first and last name. See below:
 
 ```
  ADMINS = [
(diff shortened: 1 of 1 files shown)


docs/developer/README.md
```diff
@@ -78,9 +78,30 @@ The endpoint /admin can be used to view and manage site content, including but n
         ...
  ]
 ```
+5. In the browser, navigate to /admin. To verify that all is working correctly, under "domain applications" you should see fake domains with various fake statuses.
```
abroddrick, line 81: add a new line before this line

docs/developer/README.md
```diff
@@ -78,9 +78,30 @@ The endpoint /admin can be used to view and manage site content, including but n
...
+            "first_name": "",
+            "last_name": "",
+        },
+        ...
+ ]
+```
+
+5. In the browser, navigate to /admin. To verify that all is working correctly, verify that you can only see a sub-section of the modules and some are set to view-only.
+
+Do note that if you wish to have both an analyst and admin account, append `-Analyst` to your first and last name. Example: `Bob-Analyst`
```
abroddrick, line 104:
```suggestion
Do note that if you wish to have both an analyst and admin account, append `-Analyst` to your first and last name, or use a completely different first/last name to avoid confusion. Example: `Bob-Analyst`
```

abroddrick (approved):
See two small comments, please change those if you agree, before merging. But I won't make you get another review after you make those changes.

zandercymatics:
@abroddrick Totally agree with those changes, good catch!

PR #835, opened 2023-08-04
New dev sandbox za

abroddrick:
#  #

## 🗣 Description ##

Adds Zander's sandbox infrastructure


zandercymatics (approved):
I wonder how much of this could be automated down the line, most of these changes are code duplicates with the text 'za' appended in certain locations.

I don't think its worth the time investment because its not like in the hundreds of users we need to add though. That said, I think these look good to me (aka consistent with the others). I would also wait on Rachid since he is more familiar with this process 
https://stackoverflow.com/questions/23914370/update-yaml-file-programmatically

Issue #822, opened 2023-08-01
Update the home page to match small design changes


abroddrick:
### Issue Description

Follow [ this figma](https://www.figma.com/proto/aJbuDLJNNikqQObTuCNKQa/GetGov_Designs_ECS-Truss?page-id=4702%3A38556&type=design&node-id=4829-51217&viewport=-1054%2C72%2C0.29&scaling=min-zoom&starting-point-node-id=4829%3A51217&show-proto-sidebar=1) in order to remove the status column from the Registered domain section on the home page and add change the Archived Domains title to Archive. 
AC:
- remove the status column 
- change the title for the archived domains
- edit existing tests if applicable
- ensure that everything in the figma now matches the updated page.

### Additional Context (optional)

Design determined the status wasn't needed for a registered domain, because all of them should be approved. See #697 for more background on design discussions if desired. 

### Issue Links

Created from #697 
Renders #818  a won't do

abroddrick:
@gabydisarli  does this successfully capture the ticket what is discussed in #697

abroddrick:
@zandercymatics  for the most part you will just have to edit src\registrar\templates\home.html for this one. See the h2 tags  for Registered domains and Active domain request. under the Registered domains h2 tag you will see a tr for table row and th for table header, remove the th line that has the term status. Then in the section below you will see <tbody> remove the status line from being added. 

The archived domains label change should be pretty straight forward there.

see the readme on running tests locally and double check no tests are failing when these are changed. You may have to update tests that are looking for those status or archived domain labels.

zandercymatics:
@abroddrick Since we are removing Status, will the DB object also need to be changed to remove this inclusion, or is it used elsewhere in the app? Once I am able to get the app running + able to test, I'll also remove it from index.py (as no data is being passed) unless that has some downstream affects.

With regards to testing/collectstatic, I am roadblocked on that end with the manage.py issue (usr/bin/env: 'python\r' no such file or directory)... Currently troubleshooting that but I've identified that the issue is with the line endings

abroddrick:
@zandercymatics good question! No it would need to stay in the db object as it is going to be used elsewhere. That makes sense to remove it from index.py I don't imagine that having any negative effects. Glad you identified the issue, do you need additional help with troubleshooting it?

zandercymatics:
[...]

Issue #697, opened 2023-06-08
Proposed UI for approved domain requests


igorkorenfeld:
### Ticket Description

We had a [discussion in slack](https://gsa-tts.slack.com/archives/C03QM0JGSQG/p1686150566878219) about what should be displayed to users once a domain request has been approved. I am proposing this [updated UI](https://www.figma.com/proto/ninNjo0bTg7XfB4mvTp4hG/GetGov%2FDesigns?page-id=4702%3A38556&type=design&node-id=4702-38557&viewport=607%2C209%2C0.2&scaling=min-zoom&starting-point-node-id=4702%3A38557&show-proto-sidebar=1) for displaying approved requests and viewing past (archived) requests.

A few notes with questions for feedback:
- I am suggesting that under "Registered domains" we leave off the status column. If it only displays "approved" then I am not sure how much value it would provide. However, I may be missing other important statuses that may need to be displayed here? It also helps differentiate a bit from the approved domain request that will now appear in the "Archive" section with an "Approved" status. 

- Similarly under "Archive" I am suggesting we rename this section slightly from "Archived domains" to just "Archive." The idea is that it could contain domains that have expired in addition to old requests. Do we need to accommodate expired domains (i.e. will those continue to exist and do we want to show them to registrants)? Should it instead be "Archived requests"? 

- Is there additional language we want to include on the detailed view of the approved request? I put in some text to start, but any feedback or additional thoughts are welcome here. 

### Issue Type

design review

gabydisarli:
> I am suggesting that under "Registered domains" we leave off the status column. If it only displays "approved" then I am not sure how much value it would provide. However, I may be missing other important statuses that may need to be displayed here? It also helps differentiate a bit from the approved domain request that will now appear in the "Archive" section with an "Approved" status.


I agree with this. It helps differentiate domain from domain request which strengthens the purpose behind the Archive section.

> Similarly under "Archive" I am suggesting we rename this section slightly from "Archived domains" to just "Archive." The idea is that it could contain domains that have expired in addition to old requests. Do we need to accommodate expired domains (i.e. will those continue to exist and do we want to show them to registrants)? Should it instead be "Archived requests"?


[...]

PR #837, opened 2023-08-04
#822 - Za/update home page

zandercymatics:
#  #

## 🗣 Description ##
As per #822, I've removed the 'status' column under the Registered Domains table and renamed 'Archived Domains' to 'Archive' 
image
image
These changes are ready, though I did notice some small differences between figma and the page that was designed: https://github.com/cisagov/getgov/issues/822#issuecomment-1666078778

To satisfy #822

### DIFF
--- a/src/registrar/templates/home.html
+++ b/src/registrar/templates/home.html
@@ -32,6 +32,8 @@
       </thead>
       <tbody>
         {% for domain in domains %}
+        {% comment %} ticket 796
+        {% if domain.application_status == "approved" or (domain.application does not exist)  %} {% endcomment %}
         <tr>
           <th th scope="row" role="rowheader" data-label="Domain name">
             {{ domain.name }}
--- a/src/registrar/templates/home.html
+++ b/src/registrar/templates/home.html
@@ -17,7 +17,7 @@
       Start a new domain request
     </a>
   </p>
-
+  {# Removed status per #822 #}
   <section class="section--outlined tablet:grid-col-11 desktop:grid-col-10">
     <h2>Registered domains</h2>
     {% if domains %}
--- a/src/registrar/templates/home.html
+++ b/src/registrar/templates/home.html
@@ -27,7 +27,6 @@
         <tr>
           <th data-sortable scope="col" role="columnheader">Domain name</th>
           <th data-sortable scope="col" role="columnheader">Date created</th>
-          <th data-sortable scope="col" role="columnheader">Status</th>
           <th scope="col" role="columnheader"><span class="usa-sr-only">Action</span></th>
         </tr>
       </thead>
@@ -38,7 +37,6 @@
             {{ domain.name }}
           </th>
           <td data-sort-value="{{ domain.created_time|date:"U" }}" data-label="Date created">{{ domain.created_time|date }}</td>
-          <td data-label="Status">{{ domain.application_status|title }}</td>
           <td>
             <a href="{% url "domain" pk=domain.pk %}">
               <svg 
@@ -75,7 +73,6 @@
         <tr>
           <th data-sortable scope="col" role="columnheader">Domain name</th>
           <th data-sortable scope="col" role="columnheader">Date created</th>
-          <th data-sortable scope="col" role="columnheader">Status</th>
           <th scope="col" role="columnheader"><span class="usa-sr-only">Action</span></th>
         </tr>
       </thead>
@@ -86,7 +83,6 @@
             {{ application.requested_domain.name|default:"New domain request" }}
           </th>
           <td data-sort-value="{{ application.created_at|date:"U" }}" data-label="Date created">{{ application.created_at|date }}</td>
-          <td data-label="Status">{{ application.status|title }}</td>
           <td>
              {% if application.status == "started" or application.status == "action needed" or application.status == "withdrawn"  %}
               <a href="{% url 'edit-application' application.pk %}">
@@ -119,7 +115,7 @@
   </section>
 
   <section class="section--outlined tablet:grid-col-11 desktop:grid-col-10">
-    <h2>Archived domains</h2>
+    <h2>Archive</h2>
     <p>You don't have any archived domains</p> 
   </section>
 
--- a/src/registrar/templates/home.html
+++ b/src/registrar/templates/home.html
@@ -17,7 +17,7 @@
       Start a new domain request
     </a>
   </p>
-  {# Removed status per #822 #}
+
   <section class="section--outlined tablet:grid-col-11 desktop:grid-col-10">
     <h2>Registered domains</h2>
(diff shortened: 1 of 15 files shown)


zandercymatics:
It is worth noting that I did not need to edit any test cases for this particular change

zandercymatics:
I was able to push these changes with the cf push command (they are live), though something is failing on the automatic build and deploy. I am not yet sure as to the solution, though it seems like it could potentially be related to a null value

zandercymatics:
@abroddrick I went ahead and made those changes. Its worth noting that for the 'Export Domains' section, I did not follow it word for word (though I kept it as close as possible) as it had a few grammatical issues that I'd imagine we would not want to replicate. Attached is the figma
image

abroddrick:
@gabydisarli I added you to this one so you can double check this matches the intended designs

src/registrar/templates/home.html
```diff
@@ -17,7 +17,7 @@ <h1>Manage your domains</h2>
       Start a new domain request
     </a>
   </p>
-
+  {# Removed status per #822 #}
```
abroddrick, line 20: Delete this comment, please

abroddrick (approved):
Looks good to me. Don't move it forward without @gabydisarli or @Katherine-Osos approving it. Also please delete the one comment as that is basically documented via commit history.

src/registrar/templates/home.html
```diff
@@ -121,13 +119,13 @@ <h2>Active domain requests</h2>
...
-    <h2>Archived domains</h2>
-    <p>You don't have any archived domains</p> 
+    <h2>Archive</h2>
+    <p>You don't have anything in the archive yet.</p> 
   </section>
 
   <section class="tablet:grid-col-11 desktop:grid-col-10">
     <h2 class="padding-top-1 mobile-lg:padding-top-3"> Export domains</h2>
-    <p>Download a list of your domains and their statuses as a csv file.</p>
+    <p>If you would like to analyze your list of domains further, you can download your list of domains and their statuses as a csv file.</p>
```
gabydisarli, line 128: @Katherine-Osos Can you take a look at this? I never looked at the original copy, but it had some grammatical and spelling errors. Zander has passed through it and cleaned it up, but something about the sentence structure feels off to me. Maybe it's the use of "your" so much?

gabydisarli (commented):
Nice work! One thing I noticed is that the Start a new domain request button stays in the Active domain requests table when it should disappear once that table has active domain requests in it. It should only be visible when that table is empty.

There's something else that might be a larger issue that has less to do with your update, but it's integral to how this page is designed. @h-m-f-t Reading through our prior discussions on #697 my understanding is that this bottom Archive table would contain
1. Expired (past due) domains
2. Old Domain requests that were never submitted
3. Withdrawn requests
Can you either confirm or update my understanding? Now that the record of the domain request that was approved will live in a different area (post-mvp) we need to confirm how things move into this archive area and how the mechanics work. 

cc: @Katherine-Osos + @brooksr4

src/registrar/templates/home.html
zandercymatics: @gabydisarli Personally, I found the original text here to be a bit more concise: "Download your list of domains and their statuses as a csv file.", but if we're keeping with the spirit of the Figma doc maybe we can reword it as so? 
`For further analysis, you can download your list of domains and their statuses as a csv file.`
(Maybe we could also say 'for your convenience' instead of 'for further analysis' here)

Katherine-Osos: Was this the original verbiage?: "Download a list of your domains and their statuses as a csv file."

If so, I like that better. 

Also, it looks like this content is going to be temporarily removed based on [this slack thread ](https://cisa-corp.slack.com/archives/C05BDEA3C11/p1691455212627079)and the new ticket created: #860

gabydisarli: Actually, since this is getting removed in https://github.com/cisagov/getgov/issues/860 let's shelve this conversation.

gabydisarli:
> Nice work! One thing I noticed is that the Start a new domain request button stays in the Active domain requests table when it should disappear once that table has active domain requests in it. It should only be visible when that table is empty.
> 
> There's something else that might be a larger issue that has less to do with your update, but it's integral to how this page is designed. @h-m-f-t Reading through our prior discussions on #697 my understanding is that this bottom Archive table would contain
> 
> 1. Expired (past due) domains
> 2. Old Domain requests that were never submitted
> 3. Withdrawn requests
>    Can you either confirm or update my understanding? Now that the record of the domain request that was approved will live in a different area (post-mvp) we need to confirm how things move into this archive area and how the mechanics work.
> 
> cc: @Katherine-Osos + @brooksr4

@zandercymatics + everyone tagged. I think this warrants a larger discussion and so I am going to note it as an agenda topic for tomorrow's product huddle. Zander you can ignore my comments on the archive section.

h-m-f-t:
As we discussed in the product huddle, I think it makes sense to drop "Archived Domains" for MVP because we're not clearly representing domains and _domain requests_. There are more design and engineering challenges here than we have time for, but it's still a useful problem to solve later.

abroddrick:
@zandercymatics I met with @gabydisarli  and turns out they may want to look into using that field to show the status of a domain if it's approved. But that may take some investigation and talking amongst the designers. The request is to hold off on merging the status removal until more conversations can be done. As for the other changes, such as removing the archive section, @PaulKuykendall is making a ticket for that and that can be a separate ticket and branch.

zandercymatics:
@abroddrick That makes a lot of sense, especially given that they're thinking of removing the archive. I'll leave this open for now until we make a decision on where to go next

neilmb (commented):
Looks good to me.

abroddrick:
@zandercymatics  I believe @PaulKuykendall is making a ticket for removing the archive section and there is a ticket up now for removing the export section. I'm thinking we can also just close the PR with a comment that removing the status is now being debated and that the other changes are unnecessary until archive and export are added back in.

>t appears as though the described behavior existed before this ticket was worked (screenshot is from Nikis sandbox which is based off main), for tracking purposes would it be best if this got splintered off into a separate issue, or should this be contained here?

This is definitely a bug where that button should just go away when a domain application is started. That should be a separate ticket and separate PR for sure. In general, any unrelated finding in a PR review should be made in a separate PR and not block the existing one.

gabydisarli:
Adding the link to the new ticket for the updated work for reference https://github.com/cisagov/getgov/issues/861

abroddrick:
@zandercymatics I think given the new redesigns, we can just close this with a comment for now. Would you like to do the honors?

zandercymatics:
Closing due to redesigns and intended functionality changes that have drifted away from the original purpose of this ticket

PR #858, opened 2023-08-08
update copy on election question resolving 774

### LINKED ISSUE #774: Content Update: Provide more guidance on election offices
### Story

As a Registrant applying for a domain, I want to understand how to answer the question "Is your organization an election office?" so that my application is correctly categorized when submitted.

electionsoffice

### Acceptance Criteria
- [x] Copy is drafted
- [x] Copy is approved by CISA
- [x] Question is updated in registrar 

### Additional Context
Provide clarity that the question is for elections board specifically, and not for organizations that hold elections.

_No response_

### Issue Links

See [.Gov for election offices](https://federalist-877ab29f-16f6-4f12-961c-96cf064cf070.sites.pages.cloud.gov/site/cisagov/getgov-home/domains/election-offices/) for context

gabydisarli:
## 🗣 Description ##

Updated copy on the election office question so applicants have more context to make an informed choice.
#774 

resolves #774

### DIFF
--- a/src/registrar/templates/application_org_election.html
+++ b/src/registrar/templates/application_org_election.html
@@ -6,7 +6,11 @@
     Is your organization an election office?  <abbr class="usa-hint usa-hint--required" title="required">*</abbr>
   </h2>
 
-  <p>Answer “yes” if the primary purpose of your organization is to manage elections.</p>
+  <p>An election office is a government entity whose <i>primary</i> responsibility is overseeing elections and/or conducting voter registration.
+  <br>
+  <br>
+  Answer “yes” only if the <i>main purpose</i> of your organization is to serve as an election office.</p>
+
 {% endblock %}
 
 {% block form_required_fields_help_text %}


abroddrick (approved):
Changes look good to me!

However, on a side note I tried to run voice over and ANDI on this page to double check the paragraph still read correctly with the use of "" and / in the text. I then realized I couldn't get the paragraph to be focused on with either accessibility tool! I admit I am not the best at using either of them, have any of the designers checked that we can use accessibility tools on these text fields? If so maybe someone can take a few mins to show me how I can check it correctly.

gabydisarli:
VO always throws me for a loop but I was able to focus on the paragraph and have it read correctly, I think. It does read out the slash like: "and slash or" but I'm not sure how else it would read it. I think that is correct? The thing that seems wrong to me is when navigating from the first line "An election office..." to the second line, "Answer 'yes' only if..." I get an empty group readout. That seems wrong and I wonder if it's because I have two line breaks in there <br>. 

cc: @Katherine-Osos can you give it a whirl since you are completing the TT program?

Katherine-Osos:
Visually, this looks good to me. 

I tested with VO and hear the same double "empty group" readout before the second sentence.

src/registrar/templates/application_org_election.html
```diff
@@ -6,7 +6,11 @@ <h2 class="margin-bottom-05">
     Is your organization an election office?  <abbr class="usa-hint usa-hint--required" title="required">*</abbr>
   </h2>
 
-  <p>Answer “yes” if the primary purpose of your organization is to manage elections.</p>
+  <p>An election office is a government entity whose <i>primary</i> responsibility is overseeing elections and/or conducting voter registration.
```
neilmb, line 9: *comment (non-blocking)*: I think that the semantic `<em>...</em>` is preferred for emphasis markup in the WCAG standards https://www.w3.org/WAI/WCAG21/Techniques/html/H49

src/registrar/templates/application_org_election.html
```diff
@@ -6,7 +6,11 @@ <h2 class="margin-bottom-05">
...
+  <br>
```
neilmb, line 10: *comment (non-blocking)*: I think that a more semantic approach would be to use two paragraphs `</p><p>` for keeping these two statements separate rather than the purely visual markup with `<br/>` tags.

src/registrar/templates/application_org_election.html
```diff
@@ -6,7 +6,11 @@ <h2 class="margin-bottom-05">
...
+  <br>
+  Answer “yes” only if the <i>main purpose</i> of your organization is to serve as an election office.</p>
```
neilmb, line 12: `<em>` as above?

neilmb (commented):
Just some comments around maybe preferring semantic markup for accessibility reasons.

gabydisarli:
@neilmb I've gone ahead and added your suggestions, thank you! It's reading proper to me now on VO, no more empty group. I've also used <em> instead of <i>. Visually it still looks right to me, but would like to get your thoughts.

gabydisarli:
I think we're good to go, so I'm going to merge.

PR #859, opened 2023-08-08
Add new developer sandbox 'nl' infrastructure

abroddrick:
#  #

## 🗣 Description ##

Adds developer sandbox infrastructure for Nicole

### DIFF
--- a/.github/workflows/deploy-sandbox.yaml
+++ b/.github/workflows/deploy-sandbox.yaml
@@ -16,6 +16,7 @@ jobs:
         || startsWith(github.head_ref, 'ko/')
         || startsWith(github.head_ref, 'gd/')
         || startsWith(github.head_ref, 'za/')
+        || startsWith(github.head_ref, 'nl/')
     outputs:
       environment: ${{ steps.var.outputs.environment}}
     runs-on: "ubuntu-latest"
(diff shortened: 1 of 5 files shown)


zandercymatics (approved):
LGTM, I don't see any issues here. The OWASP scan failed initially but I just reran it (as the code here is the same as before) and it worked fine

abroddrick:
@zandercymatics This is actually something that is good for you and @therealslimhsiehdy  to know. Th OWASP security test does fail occasionally on PRs, even if all you changed is readmes. As a rule we always re-run it, and if you see someone else's owasp failed, it doesn't hurt to click the re-run on it (like you did Zander). From what I recall it has to do with a race condition where it tries to run a security check on an endpoint that isn't available yet. Since it's just an informational check we don't worry a ton about it unless it fails after re-running as well. The previous team did look into fixing this issue and found it was almost impossible to troubleshoot and would have been too much of a time sync. 

^^Just an fyi as I'm sure you two will see this at some point too.

therealslimhsiehdy (approved):

abroddrick:
> @abroddrick Great info on that, thank you! I've heard OWASP is particularly prone to false-positives, is that accurate or is it generally pretty reliable with that stuff?

That is accurate from what we looked at too! Better to have false positives than false negatives?

PR #866, opened 2023-08-10
864 - Fix text too light on certain pages in dark mode (user facing pages)

### LINKED ISSUE #864: CSS changes causing light text to show when in dark mode
### Current Behavior

Going to certain pages displays some subtexts as too pale in color. 
image
image

### Expected Behavior

1. Text should match figma designs

### Steps to Reproduce
Be in dark mode and: 
1.login and click manage on an approved domain
2. see the color of the text below DNS Nameservers

or 
1.login and click on a started application 
2. Click through and see the subtexts are all pale

### Environment

Sandbox rjm & local
browser: chrome and firefox

*note ran npx gulp compile with disabled cache

machine is set to dark mode

### Additional Context

See linked branch below

### Issue Links

relates to #795 (see connected branch)

rachidatecs:
## 🗣 Description ##

- Remove broad 'body' CSS selector when setting color in dark mode (initially, to fix text too dark on change_list_results tables)
- Tested locally in django admin and on user facing pages, using system and django admin dark/light mode toggles.

Closes #864

### DIFF
--- a/src/registrar/assets/sass/_theme/_admin.scss
+++ b/src/registrar/assets/sass/_theme/_admin.scss
@@ -106,16 +106,22 @@ html[data-theme="light"] {
 
     // Dark mode django (bug due to scss cascade) and USWDS tables
     .change-list .usa-table,
-    .change-list .usa-table--striped tbody tr:nth-child(odd) td {
-        color: var(--body-fg)!important;
+    .change-list .usa-table--striped tbody tr:nth-child(odd) td,
+    .change-list .usa-table--borderless thead th,
+    .change-list .usa-table thead td, 
+    .change-list .usa-table thead th {
+        color: var(--body-fg);
     }
 }
 
 // Firefox needs this to be specifically set
 html[data-theme="dark"] {
     .change-list .usa-table,
-    .change-list .usa-table--striped tbody tr:nth-child(odd) td {
-        color: var(--body-fg)!important;
+    .change-list .usa-table--striped tbody tr:nth-child(odd) td,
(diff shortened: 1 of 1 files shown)


abroddrick (changes requested):
image
"Domain Information" now fails for contrast when system is in dark mode to begin with. it also defaulted to light at first and I clicked the toggle. not sure if that is related

rachidatecs:
Fixed. The defaulting to light at first is interesting and unexpected. I assume you double checked that your system setting is still dark mode?

zandercymatics (changes requested):
Built locally from your branch (not on sandbox as I am not added to fixtures on rjm), but this fix doesn't appear to be working for me on Chrome when the system style is set to dark:
image

That said, when I use the style button in admin (and change to dark), it works as expected

rachidatecs:
@zandercymatics, if you checkout the branch while docker was up, you might need to run `npx gulp compile` and clear cache/refresh. Lemme know.

zandercymatics:
@rachidatecs Looks like that fixed it! Everything else seems to be working as expected, the only exception is the 'None Available' text which seems to be colored black in dark mode. This seems to be agnostic when it comes to whether its system or the button... Let me know if this is not reproducible  
image

rachidatecs:
Good catch. On it.

rachidatecs:
Fixed

zandercymatics:
@rachidatecs Looks good! Doesn't seem to be broken anymore

zandercymatics (approved):

PR #867, opened 2023-08-10
839 - Fix missing column in django admin for Staff users

### LINKED ISSUE #839: Column not showing on django admin for Staff users
### Current Behavior

The first column on django admin model tables is not showing. Example: The domain names on domain applications table do not show, and a user is unable to click through the detail edit view for these rows.

### Expected Behavior

All table columns should show for Staff in django admin as long as they match set permissions.

### Steps to Reproduce

1. Log in a Staff (CISA analyst)
2. Navigate to Domain Applications in Admin

### Environment

_No response_

### Additional Context

_No response_

### Issue Links

_No response_

rachidatecs:
## 🗣 Description ##

In our change_list_results template override, we slice the first item in the results list returned for each table row and replace it with a label+checkbox markup block that meets accessibility requirements. This woks for a user with admin permissions because that user has delete perms on the model, and so there's a delete action on these tables and checkboxes get generated for each table row to allow selection for the aforementioned action.

For a user with staff permissions, she can access these models in django admin but does not have any perms to delete (or otherwise populate the 'actions' dropdown), therefore the first item in the returned results set for each row in NOT a checkbox. The bug is a result of us slicing the first data element in the results set and replacing it with a useless checkbox.
 
The bug is fixed by checking to see if first item in the result set for a change_list_results table row is a form before performing the checkbox slice/markup switch. This should be resilient enough as the first item in a row is either a checkbox or data, and we'd want to replace any checkbox with accessible markup but leave data alone.

Closes #839

### DIFF
--- a/src/registrar/templates/admin/change_list_results.html
+++ b/src/registrar/templates/admin/change_list_results.html
@@ -13,57 +13,49 @@ Load our custom filters to extract info from the django generated markup.
 {% endif %}
 {% if results %}
 <div class="results override-change_list_results">
-<table id="result_list" class="usa-table usa-table--borderless usa-table--striped">
+<table id="result_list">
 <thead>
 <tr>
 
-{# .gov - hardcode the select all checkbox #}
-<th scope="col" class="action-checkbox-column" title="Toggle all">
-    <div class="text">
-        <span>
-            <input type="checkbox" name="_selected_action" id="action-toggle">
-            <label for="action-toggle" class="usa-sr-only">Toggle all</label>
-        </span>
-    </div>
-    <div class="clear"></div>
-</th>
-{# .gov - don't let django generate the select all checkbox #}
-{% for header in result_headers|slice:"1:" %}
+{% if results.0.form %}
+    {# .gov - hardcode the select all checkbox #}
+    <th scope="col" class="action-checkbox-column" title="Toggle all">
+        <div class="text">
+            <span>
+                <input type="checkbox" name="_selected_action" id="action-toggle">
+                <label for="action-toggle" class="usa-sr-only">Toggle all</label>
+            </span>
+        </div>
+        <div class="clear"></div>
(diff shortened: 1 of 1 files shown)


src/registrar/templates/admin/change_list_results.html
```diff
@@ -13,57 +13,49 @@
 {% endif %}
 {% if results %}
 <div class="results override-change_list_results">
-<table id="result_list" class="usa-table usa-table--borderless usa-table--striped">
+<table id="result_list">
```
neilmb, line 16: **question (non-blocking)**: I don't understand why we would have to remove the USWDS table styling classes from this element in order to address this bug?

src/registrar/templates/admin/change_list_results.html
```diff
@@ -82,7 +74,37 @@
             {{ item }}
         {% endfor %}
     </tr>
-{% endfor %}
+    {% endfor %}
+
+{% else %}
```
neilmb, line 79: *suggestion*: This else is pretty far away from it's if. Could we use a short `{# ... #}` comment to explain what case this corresponds to?
```suggestion
{% else %} {# results doesn't have a form as its first element #}
```

neilmb (commented):
Looks great, thanks for breaking this out of the other work.

src/registrar/templates/admin/change_list_results.html
rachidatecs: Thank you for catching that, I'll revert (I missed this while breaking away code)

src/registrar/templates/admin/change_list_results.html
rachidatecs: Will do

abroddrick (approved):
checked djano admin tables logged in as analyst and admin. Looks fine no issue. Also not worth a unit test

PR #869, opened 2023-08-10
818 Fix the status None for approved domains in the domains table

### LINKED ISSUE #818: Domains approved in admin show up with status "None" in Registered Table
### Current Behavior

If a domain is changed in django admin to be approved, it shows in the registered table with a status of "None" even after refreshing the page. 

image

### Expected Behavior

When a domain is approved by an analyst in django admin, then it should show the status "approved" in the registered table.

### Steps to Reproduce

1. In django admin, locate a domain application that is in the "in review" state
2. Approve the domain application
3. navigate back to the home page of the application and look at the Active Domains table

### AC

When a domain application is approved, the domain application status should show as approved

### Environment

Stable and local 
Chrome & Safari

### Additional Context

Note, while the Registered domain section doesn't show the proper status, it does show the correct "approved" status in the Active Domain Request section. This may help with debugging, however, this ticket does not cover fixing the domain application from appearing in both locations that is covered in #796.  

### Issue Links

Relates to: #796 
Relates to: #701

rachidatecs:
## 🗣 Description ##

Setting the related approved_domain in the approve() in domain applications is not persisting, so when we look at the Domains table on the landing page we see None for status because the related application for that domain is not linked.

I tried calling save() in approve(), also tried setting a post_transition signal (https://stackoverflow.com/questions/47133896/does-django-fsm-call-save-method-after-state-changed). No luck on either, so I separated out the setting of approved_domain into its own non-annotated method (outside of fsm) and called it in parallel to approve() in django admin. Documented the change.

Unit test looks for application.requested_domain.name =  application.approved_domain.name and catches the bug that went unnoticed in code.

Note - the signal post_transition looked promising, if it could be made to work:

```
from django_fsm.signals import post_transition
...
@receiver(post_transition, sender=DomainApplication)
def save_new_workflow_state(sender, instance, name, source, target, **kwargs):
    """Save the new workflow state following successful django_fsm transition."""
    logger.debug('POST TRANSITION')
    if source != target:
        logger.debug('SOURCE IS DIFFERENT THAN TARGET')
        instance.save() 
```

Closes #818


src/registrar/models/domain_application.py
```diff
@@ -501,6 +501,18 @@ def _send_status_update_email(
         except EmailSendingError:
             logger.warning("Failed to send confirmation email", exc_info=True)
 
+    def set_approved_domain(self):
+        """Always called in parallel to the approve method"""
```
abroddrick, line 505:
```suggestion
        """Always called in parallel to the approve method and is used to link the domain as a foreign key in the application"""
```
added a but more detail on the purpose here

src/registrar/models/domain_application.py
```diff
@@ -501,6 +501,18 @@ def _send_status_update_email(
...
+
+        # The domain should be ready for us, let's find it
+        Domain = apps.get_model("registrar.Domain")
+        if not Domain.objects.filter(name=self.requested_domain.name).exists():
+            raise ValueError("Something went wrong and the domain did not get created.")
+        created_domain = Domain.objects.get(name=self.requested_domain.name)
```
abroddrick, line 511: created_domain implies it's being made, and get just retrieves. maybe linked_domain here and below (or something else)

src/registrar/tests/test_admin.py
```diff
@@ -153,6 +157,33 @@ def test_save_model_sends_approved_email(self):
         # Perform assertions on the mock call itself
         mock_client_instance.send_email.assert_called_once()
 
+    def test_save_model_sets_approved_domain(self):
```
abroddrick, line 160: Add a test for the negative case. That if the domain status transition fails the domain is not linked to it. That part is critical and would break a lot of things. so good to ensure it is always working now that it is separated out.

abroddrick (commented):
Really think you should add a test for the negative case but not going to block for it. tested and works.

rachidatecs:
Better diagnosed and solved in https://github.com/cisagov/getgov/pull/878

Issue #813, opened 2023-07-25
Admin | Sort all drop-downs alphabetically


PaulKuykendall:
### Story

As a CISA Analyst
I want to see all options in dropdowns sorted alphabetically in Django Admin
so that I can more easily locate my selection.

APPLIES to AO (Authorizing official), requested domain, submitter, creator.

### Acceptance Criteria

Dropdown box selection options are presented in alphabetical order.

zandercymatics:
@PaulKuykendall Hey Paul, quick question about the scope of this ticket... The ticket mentions that this only applies to four fields: AO (Authorizing official), requested domain, submitter, creator. That said, there are a few other fields on the /admin site which present a list of users out of alphabetical order, such as investigator or 'user'. Should this also apply to those, or just to the listed fields
![image](https://github.com/cisagov/getgov/assets/141044360/f9cab064-504a-4ece-91ca-00a17c436801)

zandercymatics:
Note: I've spoken to Paul about this ticket and he mentioned that this applies to most dropdowns in /admin. The exception would be 'special' fields where order does matter: i.e. organization type

PR #875, opened 2023-08-14
#813 - Za/813 sort dropdowns alphabetically

zandercymatics:
## 🗣 Description ##
Django Admin had multiple dropdown fields ordered by id, rather than by their text content (alphabetically). The original scope of ticket #813 had increased from its original to include all drop-downs where it made sense to do so. 

This PR currently sorts these fields (can add more if wanted): 
- ["submitter", "authorizing_official", "investigator", "creator", "user"] by ['first_name', 'last_name']
- ["domain", "requested_domain"] by ['name']

And includes
- A generalized class, AdminFormOrderHelper, to facilitate this task (thanks @abroddrick for the inspiration!)
- A dictionary / small function in admin.py to facilitate the above
- Test case related things
- Additions to common.py to make creating lists of Domain mock data easier

To fulfill #813


zandercymatics:
These changes should be live on my sandbox despite the back and forth with the linter

zandercymatics:
Not on my sandbox for now while #883 is being looked over, as it is of a higher priority. Will update once its back

src/registrar/models/utility/admin_form_order_helper.py
```diff
@@ -0,0 +1,44 @@
+import logging
+from typing import Dict
+from django.forms import ModelChoiceField
+
+logger = logging.getLogger(__name__)
+
+
+class SortingDictInterface:
```
rachidatecs, line 8: Inspired by Typescript/Angular patterns?

src/registrar/models/utility/admin_form_order_helper.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def __init__(self, model_list, sort_list):
+        self.sorting_dict = {
+            "dropDownSelected": model_list,
+            "sortBy": sort_list
+        }
+
+
+class AdminFormOrderHelper():
+    """A helper class to order a dropdown field in Django Admin, takes the fields you want to order by as an array""" # noqa
```
rachidatecs, line 21: Add a comment to explain the reason for noqa

src/registrar/tests/test_admin.py
```diff
@@ -367,3 +371,83 @@ def test_get_fieldsets_non_superuser(self):
...
+                        # the 'field shorthand' is used for type coercion.
+                        # It is only used here to reduce boilerplate,
+                        # but keep in mind that it is not needed outside this class.
+                        (DomainApplication.authorizing_official.field, 'ao'),
+                        (DomainApplication.submitter.field, 'you'),
+                        (DomainApplication.investigator.field, 'inv'),
+                        (DomainApplication.creator.field, 'cre')
+                        ]
+
+        # Create a sample application - review status does not matter
```
rachidatecs, line 411: Comment: Create a number of sample applications (?)

src/registrar/tests/test_admin.py
```diff
@@ -367,3 +371,83 @@ def test_get_fieldsets_non_superuser(self):
...
+    # but I think for this specific use case
+    # its not necessary since it'll only be used here.
+    def coerced_fk_field_helper(self, first_name, last_name, field_name, queryset_shorthand): # noqa
+        if(first_name and first_name.split(queryset_shorthand)[1] == field_name): # noqa
+            return (first_name, last_name)
+        else:
+            return None
+
+    def tearDown(self):
+        DomainInformation.objects.all().delete()
```
rachidatecs, line 563: Why is it necessary to clean up DomainINformation?

src/registrar/admin.py
```diff
@@ -162,12 +185,16 @@ def response_change(self, request, obj):
 
 
 class ContactAdmin(ListHeaderAdmin):
```
rachidatecs, line 187: Doesn't ContactAdmin inherit from ListHeaderAdmin which inherits from AuditedAdmin? Therefore no need to repeat def formfield_for_foreignkey?

src/registrar/admin.py
```diff
@@ -4,13 +4,32 @@
...
+                        audited_admin_orderby_names
+                        ).sorting_dict,
+    SortingDictInterface(
+                        contact_admin_item_names,
+                        contact_admin_orderby_names
+                        ).sorting_dict
+]
 
 
 class AuditedAdmin(admin.ModelAdmin):
```
rachidatecs, line 32: Suggest keeping AuditedAdmin a one job class, making a new SortFields class and have one inherit from the other.

zandercymatics:
I think my code is in need of some cleanup, but as for the concept generally: is the AuditedAdminMockData class useful? And, is the class admin_form_order_helper over complicating the issue, or is it useful here? Note that there is some repetition code wise for the three tests in AuditedAdminTest... That can totally be generalized into a function but I'm not sure if its worth devoting time to that.

With the shorthand stuff: this was a workaround for differing queryset objects having different structures, but I left it in incase we want to use something like that for a self-consistency test, or something of that nature

src/registrar/models/utility/admin_form_order_helper.py
zandercymatics: Yes! I can remove the 'interface' portion if necessary. I left it in as it was more descriptive as to its purpose, but I don't believe that name is entirely accurate

src/registrar/admin.py
zandercymatics: Good catch

zandercymatics:
Apologies for all the linting emails

src/registrar/models/utility/admin_form_order_helper.py
```diff
@@ -0,0 +1,68 @@
...
+        _order_by_list = []
+
+        for item in self._sorting_list:
+            item_dict = item.get_dict()
+            drop_down_selected = item_dict.get("dropDownSelected")
+            sort_by = item_dict.get("sortBy")
+
+            if db_field.name in drop_down_selected:
+                _order_by_list = sort_by
+                break
```
abroddrick, line 62: I usually try to avoid using breaks to make code more readable, but it does seem easy here. Maybe a comment "exit loop when order list is found"

src/registrar/models/utility/admin_form_order_helper.py
```diff
@@ -0,0 +1,68 @@
...
+
+        # Only order if we choose to do so
+        if _order_by_list is not None:
```
abroddrick, line 65:
```suggestion
        if _order_by_list is not None and _order_by_list !=[]:
```
you set order list to default to [], and not point in running it on an empty list if it is an empty list

src/registrar/tests/common.py
```diff
@@ -2,13 +2,20 @@
...
+import random
+from string import ascii_uppercase
 from unittest.mock import Mock
 from typing import List, Dict
 
 from django.conf import settings
 from django.contrib.auth import get_user_model, login
 
-from registrar.models import Contact, DraftDomain, Website, DomainApplication, User
+from registrar.models import Contact, DraftDomain, Website, DomainApplication
```
abroddrick, line 13:
```suggestion
from registrar.models import Contact, DraftDomain, Website, DomainApplication, DomainInvitation, User, DomainInformation, Domain
```

remove repeated import and just add to the existing one

src/registrar/tests/common.py
```diff
@@ -88,6 +95,204 @@ def send_email(self, *args, **kwargs):
...
+    def dummy_kwarg_boilerplate(
+        self,
+        domain_type,
+        item_name,
+        status,
+        org_type="federal",
+        federal_type="executive",
+        purpose="Purpose of the site",
+    ):
+        """Returns kwargs for different domain object types"""
```
abroddrick, line 191: I would expand this comment a little. Doesn't seem to convey what it is doing clearly

src/registrar/tests/common.py
```diff
@@ -88,6 +95,204 @@ def send_email(self, *args, **kwargs):
...
+            zipcode="10002",
+            type_of_work="e-Government",
+            anything_else="There is more",
+            authorizing_official=self.dummy_contact(item_name, "authorizing_official"),
+            submitter=self.dummy_contact(item_name, "submitter"),
+            creator=self.dummy_user(item_name, "creator"),
+        )
+        return common_args
+
+    # This can be boiled down more, though for our purposes this is OK
```
abroddrick, line 181: Maybe remove this? if it's something to come back to it would be a TODO and I don't think this warrants a todo

src/registrar/tests/common.py
```diff
@@ -88,6 +95,204 @@ def send_email(self, *args, **kwargs):
...
 
+class AuditedAdminMockData:
+    """Creates simple data mocks for AuditedAdminTest"""
+
+    # Constants for different domain object types
+    INFORMATION = "information"
+    APPLICATION = "application"
+    INVITATION = "invitation"
+
+    # These all can likely be generalized more if necessary,
```
abroddrick, line 106: Without looking at the code this comment doesn't make much sense to me. Maybe start with What problem are you solving? How are you solving it?  And if this is for the entire class not just this function consider putting this near or in the the class description

src/registrar/tests/test_admin.py
```diff
@@ -1,13 +1,24 @@
 from django.test import TestCase, RequestFactory, Client
 from django.contrib.admin.sites import AdminSite
-from registrar.admin import DomainApplicationAdmin, ListHeaderAdmin, MyUserAdmin
+from registrar.admin import DomainApplicationAdmin, ListHeaderAdmin
```
abroddrick, line 3:
```suggestion
from registrar.admin import DomainApplicationAdmin, ListHeaderAdmin, MyUserAdmin, AuditedAdmin
```

remove the duplicate import below

src/registrar/tests/test_admin.py
```diff
@@ -1,13 +1,24 @@
...
+
+# Need to split these up due to the linter
+from registrar.admin import MyUserAdmin, AuditedAdmin
 from registrar.models import DomainApplication, DomainInformation, User
+from registrar.models.contact import Contact
+from registrar.models.domain_invitation import DomainInvitation
 from .common import completed_application, mock_user, create_superuser, create_user
+
+# Need to split these up due to the linter
+from .common import multiple_unalphabetical_domain_objects
```
abroddrick, line 13: repeat same here

src/registrar/tests/test_admin.py
```diff
@@ -367,3 +378,209 @@ def test_get_fieldsets_non_superuser(self):
...
+        # Not really a fan of how this looks, but as the linter demands...
+        ordered_list = list(
+            obj_to_sort.get_queryset(request)
+            .order_by(*formatted_sort_fields)
+            .values_list(*formatted_sort_fields)
+        )
+
+        return ordered_list
+
+    # Q: These three tests can be generalized into an object,
```
abroddrick, line 405: For a test,  no. Keeping tests simple makes sense

src/registrar/tests/test_admin.py
```diff
@@ -367,3 +378,209 @@ def test_get_fieldsets_non_superuser(self):
...
+        self.client = Client(HTTP_HOST="localhost:8080")
+
+    def order_by_desired_field_helper(
+        self, obj_to_sort: AuditedAdmin, request, field_name, *obj_names
+    ):
+        formatted_sort_fields = []
+        for obj in obj_names:
+            formatted_sort_fields.append("{}__{}".format(field_name, obj))
+
+        # Not really a fan of how this looks, but as the linter demands...
```
abroddrick, line 396: What was linter complaining about?

src/registrar/tests/test_admin.py
```diff
@@ -367,3 +378,209 @@ def test_get_fieldsets_non_superuser(self):
...
+        request = self.factory.post(
+            "/admin/registrar/domaininformation/{}/change/".format(applications[0].pk)
+        )
+
+        model_admin = AuditedAdmin(DomainInformation, self.site)
+
+        sorted_fields = []
+        # Typically we wouldn't want two nested for fields,
+        # but both fields are of a fixed length.
+        # For test case purposes, this should be performant.
```
abroddrick, line 478: This is good information for a reviewer, but could just be in your PR desciption to keep an eye out for. I would remove from here before merging

src/registrar/tests/test_admin.py
abroddrick: Domains are deleted here? Haven't looked to deep in the code, but if there is an application object and an info object I would assume a domain object exists as well

abroddrick (commented):
Just did a quick pass but didn't have a chance to dive deep into the code or run it to test.  Publishing my comments before I go on lunch break

src/registrar/admin.py
```diff
@@ -4,13 +4,28 @@
 from django.contrib.contenttypes.models import ContentType
 from django.http.response import HttpResponseRedirect
 from django.urls import reverse
+from registrar.models.utility.admin_form_order_helper import AdminFormOrderHelper
```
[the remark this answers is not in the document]
zandercymatics: I had it that way originally but the linter was not happy about the length, would it make sense to add a # noqa tag here?

src/registrar/models/utility/admin_form_order_helper.py
```diff
@@ -0,0 +1,68 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class SortingDict:
+    """Stores a sorting dictionary object"""
+
+    _sorting_dict: Dict[type, type] = {}
+
+    # model_list can be will be called multiple times.
```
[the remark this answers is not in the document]
zandercymatics: Just realized the typo on line 13 lol, not fun. I shortened down this comment, good points

src/registrar/tests/common.py
zandercymatics: Good catch, not sure how that snuck in

src/registrar/admin.py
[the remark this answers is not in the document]
abroddrick: Linter complained about the length, that is no fun. I lean to the #noqa tag vs separating it into two imports. However, because it is to get around  lint if you want to keep it the same way you had it that is fine, I would just update your comment so it is clearit's sepateing due to the line length if combined. Otherwise, I could totally see an engineer coming around later and trying to consolidate them.

src/registrar/tests/test_admin.py
zandercymatics: Eh just length again, probably good to remove that comment there

src/registrar/models/utility/admin_form_order_helper.py
```diff
@@ -0,0 +1,68 @@
...
+    # Not super necessary, but it'd be nice
+    # to have the perf advantage of a dictionary,
+    # while minimizing typing when
+    # adding a new SortingDict (input as a list)
+    def convert_list_to_dict(self, value_list):
+        """Used internally to convert model_list to a dictionary"""
+        dictionary: Dict[type, type] = {}
+        for item in value_list:
```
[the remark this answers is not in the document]
zandercymatics: {item: item for item in value_list} is a neat trick. I didn't know python had that shorthand

zandercymatics:
@rachidatecs + @abroddrick Your requested changes are pushed now. Linter will fail on some of the comments so I'm fixing that now, but otherwise its all there.

src/registrar/tests/test_admin.py
zandercymatics: The FK fields that are sorted span across three different domain 'types': info, application, invitation. I am testing on all three

abroddrick:
image
did docker-compose down then up and not seeing this sorting the drop down on submitter. First one tested

abroddrick:
Just triple checked with git status, and I am on your branch but I do not see any drop downs being sorted. Ran npx gulp compile and cache was already disabled. This is when running it locally

zandercymatics:
> image did docker-compose down then up and not seeing this sorting the drop down on submitter. First one tested

Hmm. Seems to be working on my sandbox and locally:
image
Any console errors?

zandercymatics:
Made another commit to satisfy black. It does some weird things to the imports. I can go black to splitting it if it looks off, otherwise that check doesn't want to pass

zandercymatics:
> Just triple checked with git status, and I am on your branch but I do not see any drop downs being sorted. Ran npx gulp compile and cache was already disabled. This is when running it locally. This is a bit confusing because I see it working fine on yours. Did you do anything different on your sandbox?

No, everything should be up to date with my commit history... My guess might be an uncaught exception? Or something of that nature?

therealslimhsiehdy:
The only unfiltered dropdown that I had chatted with Zander about earlier is the `Domain Application`  > `Domain Application dropdown` that wasn't sorted, but all other dropdowns are sorted for me too? 

Screenshot 2023-08-17 at 3 01 34 PM

zandercymatics:
@PaulKuykendall Does everything look good on your end?

zandercymatics:
@therealslimhsiehdy Got those changes in and expanded the test case for it

src/registrar/admin.py
abroddrick: Yes I agree with this

zandercymatics: @abroddrick @rachidatecs Got that changed for you guys

src/registrar/tests/common.py
```diff
@@ -88,6 +101,226 @@ def send_email(self, *args, **kwargs):
...
+            last_name="{} last_name:{}".format(item_name, short_hand),
+            title="{} title:{}".format(item_name, short_hand),
+            email="{}testy@town.com".format(item_name),
+            phone="(555) 555 5555",
+        )[0]
+        return contact
+
+    def dummy_draft_domain(self, item_name):
+        """Creates a dummy draft domain object"""
+        return DraftDomain.objects.get_or_create(name="city{}.gov".format(item_name))[0]
```
abroddrick, line 157: If you are modifying the provided name maybe that should be documented in the doc string. For tests this seems like it would hide the name you are making, and I could easily see someone trying to make dummy_domain or whatever with "test.gov" and expect the object to have name "test" only to then see it is "citytest.gov.gov". Seems worth documenting here.

src/registrar/tests/common.py
```diff
@@ -88,6 +101,226 @@ def send_email(self, *args, **kwargs):
...
+
+        *domain_type* uses constants. Used to identify what kind of 'Domain' object you'd like to make.
+
+        In more detail: domain_type specifies what kind of domain object you'd like to create, i.e.
+        domain_application (APPLICATION), or domain_information (INFORMATION).
+        """  # noqa
+        common_args = self.get_common_domain_arg_dictionary(
+            item_name, org_type, federal_type, purpose
+        )
+        full_arg_list = None
```
abroddrick, line 223: why full arg "list" if it is type dictioanry?

src/registrar/tests/common.py
```diff
@@ -88,6 +101,226 @@ def send_email(self, *args, **kwargs):
...
+        item_name,
+        status,
+        org_type="federal",
+        federal_type="executive",
+        purpose="Purpose of the site",
+    ):
+        """
+        A helper function that returns premade kwargs for easily creating different domain object types.
+        There is a decent amount of boilerplate associated with
+        creating new domain objects (such as domain_application, or domain_information),
```
abroddrick, line 211: this repeats a lot of what is up above. We don't need why it is being made. Just what does it do and what are the parameters and return values. 
With a comment this long I expect a return statement at the end of the docstring to indicate  "returns <type>" or an arugment breakdown. I know you mention it returns kwargs, which is just keyward args, but what should someone actually expect? It looks like it is a dictionary so maybe having a break down of what the return looks like an the values. A function that is going to return wildly different values like this ideally should indicate what a user should expect, because otherwise someone has to go line by line here to see what the format is for return.  I always like this example for docstrings although it is much simpler than your case:
def multiply (a,b)
   """
    Multiplies two numbers and returns the result.
 
    Args:
        a (int): The first number.
        b (int): The second number.
 
    Returns:
        int: The product of a and b.
    """

src/registrar/tests/common.py
zandercymatics: That is a good point, I think it may be best if I just remove that entirely and replace it with name=item_name. My original intention was to provide a template to reduce typing but it doesn't really make sense because its a difference of like typing 8 characters vs 1

abroddrick (approved):
Looks good to me, thank you for incorporating all the feedback that was received!

zandercymatics:
Temporarily merged #918 into my branch so that @gabydisarli can see my changes and review them. Will make sure to revert any changes and/or differences once this gets merged up into main, none of that content will be included there unless it already exists on main by then

gabydisarli (approved):

PR #878, opened 2023-08-14
818 - Fix the approve() sets status to None bug

### LINKED ISSUE #818: Domains approved in admin show up with status "None" in Registered Table

rachidatecs:
#  #

## 🗣 Description ##

Fix the approve() sets status to None bug by calling the transition methods on the current obj (as opposed to the old obj) in django admin. To manage the fsm_transition status changes, we roll back the status to what it was before the admin user changed.

- add test_save_model_sets_approved_domain unit test
- simplify transition methods by removing the passed original_obj that is no longer needed

Resolves #818

### DIFF
--- a/src/registrar/admin.py
+++ b/src/registrar/admin.py
@@ -275,7 +275,7 @@ class DomainApplicationAdmin(ListHeaderAdmin):
                 elif obj.status == models.DomainApplication.SUBMITTED:
                     # This is an fsm in model which will throw an error if the
                     # transition condition is violated, so we roll back the
-                    # status to what it was before the admn user changed it and
+                    # status to what it was before the admin user changed it and
                     # let the fsm method set it. Same comment applies to
                     # transition method calls below.
                     obj.status = original_obj.status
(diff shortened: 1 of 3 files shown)


neilmb (approved):
What a pleasant simplification. LGTM

abroddrick (approved):
I love this change so much more! This works for me and can't seem to break it!

rachidatecs:
Same, thanks @neilmb for architecting this diagnosis and solution. Merging.

PR #882, opened 2023-08-14
Edit user fixtures to add optional emails

rachidatecs:
## 🗣 Description ##

- Added Rachid's email to Rachid's analyst user fixture

- This should help streamline notification testing

### DIFF
--- a/docs/developer/README.md
+++ b/docs/developer/README.md
@@ -84,6 +84,7 @@ The endpoint /admin can be used to view and manage site content, including but n
 ```
 
 5. In the browser, navigate to /admin. To verify that all is working correctly, under "domain applications" you should see fake domains with various fake statuses.
+6. Add an optional email key/value pair
 
 ### Adding an Analyst to /admin
 Analysts are a variant of the admin role with limited permissions. The process for adding an Analyst is much the same as adding an admin:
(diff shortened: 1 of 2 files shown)


neilmb (commented):
Looks very reasonable.

src/registrar/fixtures.py
```diff
@@ -79,6 +79,7 @@ class UserFixture:
             "username": "319c490d-453b-43d9-bc4d-7d6cd8ff6844",
             "first_name": "Rachid-Analyst",
             "last_name": "Mrad-Analyst",
+            "email": "rachid.mrad@gmail.com",
```
abroddrick, line 82: any reason you just added an email to staff not admin as well? I would think with this setup you could do both in case you log in as a user via your admin level account

abroddrick (approved):

src/registrar/fixtures.py
therealslimhsiehdy: Just to confirm w/ this change, if we add our own email(s) for Staff (and possibly Analyst?), it will dynamically assist us with emails, correct?

therealslimhsiehdy (approved):

docs/developer/README.md
```diff
@@ -84,6 +84,7 @@ The endpoint /admin can be used to view and manage site content, including but n
 ```
 
 5. In the browser, navigate to /admin. To verify that all is working correctly, under "domain applications" you should see fake domains with various fake statuses.
+6. Add an optional email key/value pair
```
zandercymatics, line 87: My only suggestion here would be to include an example somewhere, eg: "email" = "<Optional Email Here>"

zandercymatics (approved):
Nice changes, this will be especially useful for the SES stuff

src/registrar/fixtures.py
rachidatecs: Correct @therealslimhsiehdy. No reason @abroddrick other than I've gotten in the habit of testing notifications on my Analyst user, triggered from my Superuser user.

docs/developer/README.md
rachidatecs: I think the context is strong enough (a few lines before the diff, there's a code block with example firnatname, lastanme, etc.)

Issue #794, opened 2023-07-13
Django admin "log out" doesn't log a user out


abroddrick:
### Current Behavior

On the admin page, when you click  "log out"  in the upper right corner, nothing happens

### Expected Behavior

When clicking log out, you should be logged out of the application

### Steps to Reproduce

1.got to \admin
2. click  "log out"  in the upper right corner
3. notice the sound of crickets as nothing happens

### Environment

environrment: stable or local
Browser: Chrome  114.0.5735.198

zandercymatics:
Reopened as I am unsure as to if the originator should close this

PR #883, opened 2023-08-14
#794 - Za/794 fix admin logout issue

zandercymatics:
## 🗣 Description ##
***Changes:**
- Made an edit to the userlinks block to replace it with an <a> tag, rather than a form
- Changed the order of the 'admin/logout' redirect to persist higher up the list (due to a bug)
- Removed the redirect for 'admin/login' as it was not functional and not intended to be there

The logout button on /admin was not functional as it was using a form submission rather than a simple redirect to do so. This change makes alterations to block 'userlinks' in base_site.html, and modifies urls.py to accommodate this change (redirect view).

image

To satisfy #794

### DIFF
--- a/src/registrar/config/settings.py
+++ b/src/registrar/config/settings.py
@@ -285,7 +285,7 @@ SERVER_EMAIL = "root@get.gov"
 
 # Content-Security-Policy configuration
 # this can be restrictive because we have few external scripts
-allowed_sources = ("'self'",)
+allowed_sources = ("'self'", "https://idp.int.identitysandbox.gov", "https://idp.int.identitysandbox.gov/openid_connect/logout")
 CSP_DEFAULT_SRC = allowed_sources
 # Most things fall back to default-src, but these two do not and should be
 # explicitly set
--- a/src/registrar/config/settings.py
+++ b/src/registrar/config/settings.py
@@ -290,7 +290,7 @@ CSP_DEFAULT_SRC = allowed_sources
 # Most things fall back to default-src, but these two do not and should be
 # explicitly set
 CSP_FRAME_ANCESTORS = allowed_sources
-CSP_FORM_ACTION = ("'self'", "https://idp.int.identitysandbox.gov/openid_connect/logout")
+CSP_FORM_ACTION = allowed_sources
 
 
 # Content-Length header is set by django.middleware.common.CommonMiddleware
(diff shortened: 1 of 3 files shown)


zandercymatics:
Since the new PR template is still in progress, I'll keep to this old one for now, and update later once that gets put in place

src/registrar/config/settings.py
```diff
@@ -285,7 +285,7 @@
 
 # Content-Security-Policy configuration
 # this can be restrictive because we have few external scripts
-allowed_sources = ("'self'",)
+allowed_sources = ("'self'")
```
neilmb, line 288: **suggestion**: The trailing comma is important here to make `allowed_sources` into a one-element tuple. As the [`django-csp` docs](https://django-csp.readthedocs.io/en/latest/configuration.html) say "Note when a setting requires a tuple or list. Since Python strings are iterable, you may get very strange policies and errors."

neilmb:
```suggestion
allowed_sources = ("'self'",)
```

src/registrar/config/urls.py
```diff
@@ -114,18 +118,15 @@
     ),
 ]
 
-
+# What is the purpose of this?
+# This behaviour gets overwritten, so this doesn't do anything...
+# Login in particular
```
neilmb, line 123: **comment (non-blocking)**: I think you're right that we could remove the conditional and make these redirects apply in all situations.

src/registrar/templates/admin/base_site.html
```diff
@@ -13,5 +13,20 @@ <h1 id="site-name"><a href="{% url 'admin:index' %}">.gov admin</a></h1>
   {% include "admin/color_theme_toggle.html" %}
 {% endif %}
 {% endblock %}
-
+{% block userlinks %}
```
neilmb, line 16: **comment (non-blocking)**: It looks like this was probably copied out of the default Django admin templates so that it could be modified. If so, it might be nice to include a link back to the source in case Django modifies the structure of the base template in the future and we need to change this block to match.

neilmb (commented):
The template change looks great, but the CSP configuration needs to be a tuple, not a single string.

src/registrar/config/settings.py
zandercymatics: Good catch, thanks!

gabydisarli (approved):
I don't have the expertise to know if everything you're doing under the hood is correct, but I'm able to click the log out button and it works, so, approved.

PR #885, opened 2023-08-14
818 - Tie the domain permissions to the related domain application's status

rachidatecs:
## 🗣 Description ##

- Tie the domain permissions to the related domain application's status

Solves #818


rachidatecs:
Draft PR, no longer needed

PR #918, opened 2023-08-21
New dev sandbox dk

rachidatecs:
## Ticket

Part of #900 

## Changes

- Create manifest for dk
- Add dk to workflows
- Update Org name in ops and workflow files 
- Add dk to ALLOWED_HOSTS

## Context for reviewers

Not sure whether the Org name update is needed, not needed bu benign, or inadvisable at this point.

### DIFF
--- a/.github/workflows/reset-db.yaml
+++ b/.github/workflows/reset-db.yaml
@@ -25,6 +25,7 @@ on:
           - ab
           - bl
           - rjm
+          - dk
 
 jobs:
   reset-db:
@@ -38,7 +39,7 @@ jobs:
         with:
           cf_username: ${{ secrets[env.CF_USERNAME] }}
           cf_password: ${{ secrets[env.CF_PASSWORD] }}
-          cf_org: cisa-getgov-prototyping
+          cf_org: cisa-dotgov
           cf_space: ${{ github.event.inputs.environment }}
           full_command: "cf run-task getgov-${{ github.event.inputs.environment }} --command 'python manage.py flush --no-input' --name flush"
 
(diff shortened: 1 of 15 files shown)


src/registrar/tests/test_views.py
```diff
@@ -1500,3 +1500,19 @@ def test_application_status_no_permissions(self):
                         reverse(url_name, kwargs={"pk": application.pk})
                     )
                     self.assertEqual(page.status_code, 403)
+
+
+def test_approved_application_not_in_active_requests(self):
```
neilmb, line 1505: **comment**: It looks like this probably crept in from a separate workstream. Is it easy enough leave this change out of this PR?

neilmb (commented):
Thanks for setting this up Rachid. This PR will also make `cisa-dotgov` our default organization for sandboxes, etc. which is awesome!

.github/workflows/reset-db.yaml
```diff
@@ -25,6 +25,7 @@ on:
           - ab
           - bl
           - rjm
+          - dk
```
abroddrick, line 28: did you manually add this? the script should auto add this to the top under staging

abroddrick (approved):
Thanks!

.github/workflows/reset-db.yaml
rachidatecs: Yes @abroddrick , I manually added most of these. The scripts had a stroke at the build step, due to either me having docker composed up already or being in the wrong directory. I had to run the rest of the steps manually and then edit the files by hand.

therealslimhsiehdy (approved):
Thank you!! 👯

PR #920, opened 2023-08-22
Migrate .gov team's cloud.gov organization

neilmb:
## Changes

This adds scripts that can be used to recreate our environments in the new CISA Cloud.gov organization. In the process the app manifests are updated to their latest templates where the GETGOV public site is beta.get.gov. All of the Github Actions changes were already merged in a recent PR.

## Context for reviewers

CISA needed to move everything from the `cisa-getgov-prototyping` org to the new `cisa-dotgov` org. These scripts are vestigial now that the migration is complete, but I think that they are worth keeping in the repo as a record of how these sort of tasks can be handled.

## Setup

Unfortunately, it isn't reasonable to run this code to test it.

As above, it isn't possible to run this code just for test purposes, so the existence of the new environments has to serve as its own documentation.

### DIFF
--- a/ops/scripts/create_environment_migrate.sh
+++ b/ops/scripts/create_environment_migrate.sh
@@ -111,7 +111,7 @@ then
     exit 1
 fi
 
-cf service-key github-cd-account github-cd-key | sed 1,2d  | jq -r '[.credentials.username, .credentials.password]|@tsv' | 
+cf service-key github-cd-account github-cd-key | sed 1,2d  | jq -r '[.credentials.username, .credentials.password]|@tsv' |
 while read -r username password; do
     gh secret --repo cisagov/getgov set CF_${upcase_name}_USERNAME --body $username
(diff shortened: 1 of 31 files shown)


ops/scripts/create_environment_migrate.sh
```diff
@@ -0,0 +1,127 @@
+# This script sets up a completely new Cloud.gov CF Space with all the corresponding
```
rachidatecs, line 1: The one thing that tripped me up when creating Dave's sandbox (and I should've updated the create_sandbox script's comment in hindsight): I ran the script while on a feature branch and docker was composed up, so the build command failed and the created branch did not originate from main. Worth adding to the comments: "Run from main with docker composed down"?

h-m-f-t (approved):
I did not run the scripts, but I did read through them. I think it's great that we'll share this ops documentation with others.

PR #961, opened 2023-08-30
898 - Django admin semantic markup for ADD and HISTORY links

### LINKED ISSUE #898: Non-semantic markup for some action buttons in Django admin
### Issue Description

Our initial accessibility compliance screening result highlighted non-critical but semantically incorrect markup on the 'Add' and 'History' buttons in Django admin.  

These buttons are marked up as anchors inside of unordered lists. The unordered list is not semantic in this case since it's a single link.

_No response_

### Issue Links

Relates to #788

rachidatecs:
## Ticket

Resolves #898

## Changes

- Change in templates ul and li to p for Django admin object-tools links (ADD on change_list and HISTORY on change_form)

## Context for reviewers

This is the last bit of accessibility fix that we needed on django admin, according to Aaron.

This is a minimum viable fix (MVF?) that will satisfy requirement but needs some careful reviewing to make sure it's not dangerous. Form the Django templates, it seems like the initial implementation into an ordered list here was sloppy but I may be missing a repeater or iterator somewhere that'll cause the fix to have some unintended consequences under certain conditions.

Note: The rejected fixture is really tripping up Pa11y, so I commented it out. Maybe that bit belongs into its own separate PR?

Django original code:

https://github.com/django/django/blob/68a8996bdfce2d191decd7b1c1a2b9fdea8e4b2f/django/contrib/admin/templates/admin/change_form.html#L27-L34

https://github.com/django/django/blob/68a8996bdfce2d191decd7b1c1a2b9fdea8e4b2f/django/contrib/admin/templates/admin/change_list.html#L43C7-L49

https://github.com/django/django/blob/68a8996bdfce2d191decd7b1c1a2b9fdea8e4b2f/django/contrib/admin/templates/admin/change_form_object_tools.html

https://github.com/django/django/blob/68a8996bdfce2d191decd7b1c1a2b9fdea8e4b2f/django/contrib/admin/templates/admin/change_list_object_tools.html 

## Setup

Log into admin as a superuser and/or staff, navigate to list views and form views (applications listing then an individual application), inspect the markup on the grey buttons at the to right.

- [x] Meets all designs and user flows provided by design/product

  - [x] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

## Screenshots

Screen Shot 2023-09-01 at 3 38 29 PM

Screen Shot 2023-09-01 at 3 38 16 PM

### DIFF
--- a/src/registrar/fixtures.py
+++ b/src/registrar/fixtures.py
@@ -268,10 +268,11 @@ class DomainApplicationFixture:
             "status": "action needed",
             "organization_name": "Example - Action Needed",
         },
-        {
-            "status": "rejected",
-            "organization_name": "Example - Rejected",
-        },
+        # Pa11y does not like this fixture
+        # {
+        #     "status": "rejected",
+        #     "organization_name": "Example - Rejected",
+        # },
     ]
 
     @classmethod
--- a/src/registrar/fixtures.py
+++ b/src/registrar/fixtures.py
@@ -268,11 +268,10 @@ class DomainApplicationFixture:
             "status": "action needed",
             "organization_name": "Example - Action Needed",
         },
-        # Pa11y does not like this fixture
-        # {
-        #     "status": "rejected",
-        #     "organization_name": "Example - Rejected",
-        # },
+        {
+            "status": "rejected",
+            "organization_name": "Example - Rejected",
+        },
     ]
 
     @classmethod
--- a/src/registrar/templates/admin/change_form.html
+++ b/src/registrar/templates/admin/change_form.html
@@ -1,0 +1,11 @@
+{% extends "admin/change_form.html" %}
+
+{% block object-tools %}
+{% if change and not is_popup %}
+    <p class="object-tools">
+        {% block object-tools-items %}
+            {{ block.super }}
+        {% endblock %}
+    </p>
+{% endif %}
+{% endblock %}
--- a/src/registrar/templates/admin/change_form.html
+++ b/src/registrar/templates/admin/change_form.html
@@ -1,6 +1,6 @@
 {% extends "admin/change_form.html" %}
 
-{% comment %} Replace the Django ul markup with a div. We'll edit the child markup accordingly in form_list_object_tools {% endcomment %}
+{% comment %} Replace the Django ul markup with a div. We'll edit the child markup accordingly in change_form_object_tools {% endcomment %}
 {% block object-tools %}
 {% if change and not is_popup %}
     <div class="object-tools">
--- a/src/registrar/templates/admin/change_form.html
+++ b/src/registrar/templates/admin/change_form.html
@@ -1,11 +1,12 @@
 {% extends "admin/change_form.html" %}
 
+{% comment %} Replace the Django ul markup with a div. We'll edit the child markup accordingly in form_list_object_tools {% endcomment %}
 {% block object-tools %}
 {% if change and not is_popup %}
-    <p class="object-tools">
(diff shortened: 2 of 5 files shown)


src/registrar/fixtures.py
```diff
@@ -268,10 +268,11 @@ class DomainApplicationFixture:
             "status": "action needed",
             "organization_name": "Example - Action Needed",
         },
-        {
-            "status": "rejected",
-            "organization_name": "Example - Rejected",
-        },
+        # Pa11y does not like this fixture
```
zandercymatics, line 271: I wonder if storing this in a variable, or parts of it, would make a difference. Its weird that it doesn't like this when its text content follows a similar pattern to everything else. The only other thing I can think of might be that it doesn't like the length of this dictionary?

rachidatecs: Nope on length. I tried removing Action Needed and keeping Rejected, still freaks Pa11y out.

zandercymatics:
Tested using ANDI and screen reader for chrome: https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn, worked well for both. Was not able to find a consistent screen reader for Firefox (so I won't mark it as tested) but it seemed to work on Perciles and Read Aloud - though they both rip text content directly so I cannot speak to that

src/registrar/templates/admin/change_form.html
```diff
@@ -0,0 +1,11 @@
+{% extends "admin/change_form.html" %}
+
+{% block object-tools %}
+{% if change and not is_popup %}
```
zandercymatics, line 4: Is this grabbed directly from the template? If so, did you see any other condition here, for instance: if change and is_popup?

src/registrar/templates/admin/change_list.html
```diff
@@ -24,3 +24,11 @@ <h2>
         {% endif %}
     </h2>    
 {% endblock %}
+
+{% block object-tools %}
+    <p class="object-tools">
```
zandercymatics, line 29: I think this approach works fine, but I'd make a note somewhere in the description or comments before closing as to the specifics of why we chose to do this. For our purposes I highly doubt we'd need to iterate through a list but it would be good to document this in the event we do

zandercymatics (approved):
LGTM code and functionality wise... I don't see any outstanding issues. Your code was pretty straightforward. The main thing I'd recommend doing here is just leaving a code comment on the edited template in the event we need to change this back for some reason, but I really do doubt we will need too

therealslimhsiehdy (approved):
LGTM as well -- no major qualms as the code change seems pretty straightforward, although do agree with Zander (up to your discretion, it's a small NIT) on leaving a code comment that it was changed from <li> to this bc of accessibility reasons

src/registrar/templates/admin/change_form.html
```diff
@@ -0,0 +1,12 @@
+{% extends "admin/change_form.html" %}
+
+{% comment %} Replace the Django ul markup with a div. We'll edit the child markup accordingly in form_list_object_tools {% endcomment %}
```
neilmb, line 3: Nit: the edit here is in change_form_object_tools

neilmb (commented):
LGTM. Thanks for the screenshot to make sure we didn't wreck the layout!

PR #974, opened 2023-09-01
973 - concatenate the value + extracted text for both label for attribute a…

### LINKED ISSUE #973: Duplicate labels/ids for Django admin checkboxes
### Current Behavior

The ids and labels for the Django admin checkboxes are being pulled from the text of the anchor element in the next column, which is the model identifier. For Domain Applications when a Started DA has no requested domain yet, that value is "-". 

Consequently, if there are 2 or more started DAs with no requested domains, checkboxes with duplicate ids and labels are detected on the page.

The fix is to concatenate the value + extracted text for both label for attribute and checkbox id attribute.

### Expected Behavior

No duplicate ids or labels on any Django admin page

### Steps to Reproduce

1. Log into django admin
2. Create 2 new DAs
3. Run ANDI on DA list page

### Environment

_No response_

### Additional Context

_No response_

### Issue Links

_No response_

rachidatecs:
…nd checkbox id attribute

## Ticket

Resolves #973 

## Changes

- Concatenate the value + extracted text for both label for attribute and checkbox id attribute

## Context for reviewers

The ids and labels for the Django admin checkboxes are being pulled from the text of the anchor element in the next column, which is the model identifier. For Domain Applications when a Started DA has no requested domain yet, that value is "-".

Consequently, if there are 2 or more started DAs with no requested domains, checkboxes with duplicate ids and labels are detected on the page.

The fix is to concatenate the value + extracted text for both label for attribute and checkbox id attribute.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

## Screenshots

Screen Shot 2023-09-01 at 5 43 23 PM

### DIFF
--- a/src/registrar/templates/admin/change_list_results.html
+++ b/src/registrar/templates/admin/change_list_results.html
@@ -60,12 +60,11 @@ Load our custom filters to extract info from the django generated markup.
         <tr><td colspan="{{ result|length }}">{{ result.form.non_field_errors }}</td></tr>
     {% endif %}
     <tr>
-
         {% with result_value=result.0|extract_value %}
             {% with result_label=result.1|extract_a_text %}
                 <td>
-                    <input type="checkbox" name="_selected_action" value="{{ result_value|default:'value' }}" id="{{ result_label|default:result_value }}" class="action-select">
-                    <label class="usa-sr-only" for="{{ result_label|default:result_value }}">{{ result_label|default:'label' }}</label>
+                    <input type="checkbox" name="_selected_action" value="{{ result_value|default:'value' }}" id="{{ result_value|default:'value' }} {{ result_label|default:'label' }}" class="action-select">
+                    <label class="usa-sr-only" for="{{ result_value|default:'value' }} {{ result_label|default:'label' }}">{{ result_label|default:'label' }}</label>
                 </td>
                 {% endwith %}
         {% endwith %}
--- a/src/registrar/templates/admin/change_list_results.html
+++ b/src/registrar/templates/admin/change_list_results.html
@@ -63,8 +63,8 @@ Load our custom filters to extract info from the django generated markup.
         {% with result_value=result.0|extract_value %}
             {% with result_label=result.1|extract_a_text %}
                 <td>
-                    <input type="checkbox" name="_selected_action" value="{{ result_value|default:'value' }}" id="{{ result_value|default:'value' }} {{ result_label|default:'label' }}" class="action-select">
-                    <label class="usa-sr-only" for="{{ result_value|default:'value' }} {{ result_label|default:'label' }}">{{ result_label|default:'label' }}</label>
+                    <input type="checkbox" name="_selected_action" value="{{ result_value|default:'value' }}" id="{{ result_value|default:'value' }}-{{ result_label|default:'label' }}" class="action-select">
+                    <label class="usa-sr-only" for="{{ result_value|default:'value' }}-{{ result_label|default:'label' }}">{{ result_label|default:'label' }}</label>
                 </td>
                 {% endwith %}
         {% endwith %}


therealslimhsiehdy:
Just to clarify, if a user has two started Domain Applications the screen reader will read the same IDs and labels?

(The code LGTM but just wanted to confirm context)

src/registrar/templates/admin/change_list_results.html
```diff
@@ -60,12 +60,11 @@
...
     {% endif %}
     <tr>
-
         {% with result_value=result.0|extract_value %}
             {% with result_label=result.1|extract_a_text %}
                 <td>
-                    <input type="checkbox" name="_selected_action" value="{{ result_value|default:'value' }}" id="{{ result_label|default:result_value }}" class="action-select">
-                    <label class="usa-sr-only" for="{{ result_label|default:result_value }}">{{ result_label|default:'label' }}</label>
+                    <input type="checkbox" name="_selected_action" value="{{ result_value|default:'value' }}" id="{{ result_value|default:'value' }} {{ result_label|default:'label' }}" class="action-select">
+                    <label class="usa-sr-only" for="{{ result_value|default:'value' }} {{ result_label|default:'label' }}">{{ result_label|default:'label' }}</label>
```
neilmb, line 67: **suggestion**: "An id's value must not contain whitespace" https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

```suggestion
                    <input type="checkbox" name="_selected_action" value="{{ result_value|default:'value' }}" id="{{ result_value|default:'value' }}-{{ result_label|default:'label' }}" class="action-select">
                    <label class="usa-sr-only" for="{{ result_value|default:'value' }}-{{ result_label|default:'label' }}">{{ result_label|default:'label' }}</label>
```

neilmb (commented):
Just a tweak to get rid of whitespace in the `id` attribute.

src/registrar/templates/admin/change_list_results.html
rachidatecs: Good catch!

therealslimhsiehdy:
> > Just to clarify, if a user has two started Domain Applications the screen reader will read the same IDs and labels?
> > (The code LGTM but just wanted to confirm context)
> 
> The id and label for each started application will be "-", and therefore duplicates.

Thank you for clarifying! I added two screenshots just for visual reference (feel free to update the PR with them if you'd like!):

On my local without your changes, you can see in the red circles that the IDs are duplicated as "-":
Screenshot 2023-09-05 at 3 31 28 PM

On staging with your newest wonderful changes, you can see in the red circles now that the IDs are different with "81--" and "82--":
Screenshot 2023-09-05 at 3 33 19 PM

Huge impact!

therealslimhsiehdy (approved):

rachidatecs:
You just added the screenshots to the PR, @therealslimhsiehdy  :)
Thanks for the thorough testing.

Do we care that in case of a non-existing result_label string (the string we're pulling from the anchor text, ie the requested domain), Django defaults to '-'. So the concatenated uniques string is now id + '-' + '-' = id--

It still meets the goal of providing accessible and unique  ids and labels, just ain't super pretty. Maybe follow-up with a small ticket to replace '-' with 'new-application', resulting in strings that look like {id}-new-application?

zandercymatics:
@rachidatecs Not Rebecca but I would personally say that what you said is worth addressing only if in the foreseeable future we need to use those ids for something. Django seems to handle those id changes fine (the 'for' label), so that additional dash doesn't seem to break anything. 

Fixing it would benefit the following: 
a) future development using these ids (would make it slightly easier to figure out / deal with)
- If this is a concern, personally I would say documenting this as a comment would be sufficient. (Ex: will result in an id of 'x--') 

b) html/code cleanliness 
- I don't think this negatively detracts from this as its a small thing. In general, I would argue that code cleanliness should be a priority where many eyes will lie and work collaboratively. If its an area that we (very likely) will not touch again, then imo I don't personally consider it necessary.

**TLDR:** I don't think there's any harm in creating a follow-up ticket for this, but I personally don't think this is something we _need_ to address. It falls into the "nice to have" category.

rachidatecs:
I'm thinking this could be an improvement for screen readers. The 'for' attribute gets read out-loud, no?

I could ask Aaron...

zandercymatics:
@rachidatecs Ohh... Yes that is an excellent point. I did not even think of that. If it does get read aloud then 100% we should get this fixed at a later date

neilmb:
`for` does not get read out loud. It is the link between the label's `id` and the input element that corresponds to the label so that the screenreader can announce the correct label when the input element is selected. IDs ought never be exposed to the user, so as long as they are unique, preventing collisions for situations like this, they can have as many dashes as they want.

Issue #854, opened 2023-08-07
Approved state switching requires non-active state


kitsushadow:
### Issue Description

In order to switch from approved check if a domain is active #852

### Additional Context (optional)

If isActive is true, don't allow

abroddrick:
@zandercymatics did you start on this one actually?

PR #986, opened 2023-09-06
811 - Manage domain and domainInfo on rejected/ineligible transitions

### LINKED ISSUE #811: Add ability for approved domains to switch to rejected or ineligible and back to approved later
### Story

As a analyst
I want to be able to mark a domain in a valid non-approved* state as approved even if it was approved previously
so that action needed and ineligible domains are not stuck in that state and can move directly to approved

Valid states to switch to approved are: in review, action needed, ineligible [see the status doc to check for other states not listed ehre](https://docs.google.com/document/d/1no29E2c9QWIHf_2wv3GCy6AotZ2_PVbIBiavKW-opxE/edit#heading=h.qsh70q)

### Acceptance Criteria

-An approved domain should be able to be rejected and follow the correct user flow until it is marked as approved again without any errors (via django admin as an analyst user)
-An approved domain should be able to be marked as ineligible then marked as approved without any errors (via django admin as an analyst user)

Confirm for refinement that this is the best approach and valid additionalAC:
-domain should be removed when switched from approved to another state
-in this domain function there should be a stubbed call to the domain object to first delete the domain from the registry

### Additional Context

Currently, we add the domain object when a domain is approved;  however, when we mark it as rejected or ineligible the domain object isn't removed. Thus when we go to reapprove the domain an error occurs

### Issue Links

_No response_

rachidatecs:
## Ticket

Resolves #811 

## Changes

- User friendly alert in admin if attempt to transition to ineligible or rejected while domain is_active
- Add rule in model to prevent transition to ineligible or rejected while domain is_active
- Add side effects in model to delete domain and domain_info if transitioning from an active DA
- Clean up duplicate isActive in domain model
- Stub out EPP delete_request in domain model
- Unit test models and admin, not allowed transitions and side effects
- Add search for DomainInfo admin list view (Out of scope)
- Add username (uuid) for user list view (Out of scope)

## Setup Instructions

To test cases where the domain is active, pull and hardcode `True`  on the return in is_active in the domain model. Designers, screenshot below of the notification in django admin when the transition is not allowed.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

## Screenshots

Screen Shot 2023-09-06 at 4 35 30 PM


Katherine-Osos (approved):

src/registrar/admin.py
```diff
@@ -328,29 +384,57 @@ def save_model(self, request, obj, form, change):
...
-
-            super().save_model(request, obj, form, change)
+                if (
+                    obj
+                    and original_obj.status == models.DomainApplication.APPROVED
+                    and (
+                        obj.status == models.DomainApplication.REJECTED
+                        or obj.status == models.DomainApplication.INELIGIBLE
+                    )
+                    and not obj.domain_is_not_active()
```
therealslimhsiehdy, line 394: Apologies if I'm missing context (and this may be a NIT), but it may be preferred to have a `domain_is_active` and then do a `not` against that instead of currently having `not` against another 'not active` which can be confusing.

However I do see below that you're referencing `domain_is_not_active` a lot in `src/registrar/models/domain_application.py` ...just some food for thought 😅

(Let me know if this doesn't make sense, I can reword or hop on call to explain!)

abroddrick: Agreed I am curious why is active was changed here. It is also a requirement for ticket 848 so changing it here changes that ticket.

src/registrar/models/domain_application.py
```diff
@@ -477,6 +477,11 @@ def __str__(self):
         except Exception:
             return ""
 
+    def domain_is_not_active(self):
+        if self.approved_domain:
+            return not self.approved_domain.is_active()
```
abroddrick, line 482: Wait please check me, but I believe this actually satisfies #854 

if you agree, please add your name to that ticket, remove zanders name, and update your PR  with this as being resolved as well

src/registrar/models/domain.py
```diff
@@ -338,9 +344,6 @@ def __str__(self) -> str:
         help_text="Very basic info about the lifecycle of this domain object",
     )
 
-    def isActive(self):
```
abroddrick, line 341: Copy this logic instead into is_active. There should not have been two, but this is the accurate implementation. I am actually wonder, when this is added in will your tests just fail? While we said to stub it out initially, I wonder if this a bad idea on my part. And instead, during your tests it should just use a patch for testing. In fact with it stubbed for false only one scenario is tested and we should also check that if it is active it doesn't move forward. What are your thoughts? (Beside "grr alysia why are you making me make changes so late")

src/registrar/models/domain_application.py
abroddrick: EDIT: ignore the above, it doesn't fully meet 854. I added a comment onto your is active function as I wonder if there is a conflict here.

src/registrar/models/domain.py
rachidatecs: I'm testing invalid transitions in the model tests, using patching and active domains.

src/registrar/models/domain_application.py
rachidatecs: Look at the tests, I think I've got this covered?

src/registrar/admin.py
rachidatecs: Yep, conditions=[not domain_is_active] returns this error.

Screen Shot 2023-09-11 at 2 29 21 PM

abroddrick (approved):
Rachid as alwasy love your comments. Tested several times and couldn't get it to break. Seems solid

src/registrar/migrations/0031_alter_domainapplication_approved_domain_and_more.py
```diff
@@ -0,0 +1,37 @@
+# Generated by Django 4.2.1 on 2023-09-06 20:30
+
+from django.db import migrations, models
```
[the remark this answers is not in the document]
therealslimhsiehdy: +1 on this, we've both battled this so many times in the past few days and very happy to help as well!

therealslimhsiehdy (commented):
Everything looks good to me overall -- just not approving yet as you have merge conflicts + you will need to rerun the migration, but will approve when those are updated!!

(Also super sorry, I know those merge + migration conflict is from me 😭)

PR #1002, opened 2023-09-08
add Nicolle LeClair to admin and staff

CocoByte:
## Ticket

Resolves (onboarding - fixtures)

## Changes
Updated fixtures.py with info for Nicolle LeClair
- added info to admin
- added info to staff

### DIFF
--- a/src/registrar/fixtures.py
+++ b/src/registrar/fixtures.py
@@ -77,6 +77,11 @@ class UserFixture:
             "first_name": "David",
             "last_name": "Kennedy",
         },
+        {
+            "username": "f14433d8-f0e9-41bf-9c72-b99b110e665d",
+            "first_name": "Nicolle",
+            "last_name": "LeClair",
+        },
     ]
 
     STAFF = [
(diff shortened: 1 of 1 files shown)


therealslimhsiehdy:
For reformatting locally, the command is `docker-compose exec app black [file path here]` so in this case would be `docker-compose exec app black registrar/fixtures.py`!

src/registrar/fixtures.py
```diff
@@ -123,6 +128,12 @@ class UserFixture:
             "last_name": "DiSarli-Analyst",
             "email": "gaby@truss.works",
         },
+        {
+            "username": "cfe7c2fc-e24a-480e-8b78-28645a1459b3",
+            "first_name": "Nicolle",
+            "last_name": "LeClair",
```
zandercymatics, line 134:
```suggestion
            "first_name": "Nicolle-Analyst",
            "last_name": "LeClair-Analyst",
```
Because of how users are displayed, if your staff user has the same name as your Admin user, it will basically look like there is a duplicate record (though there isn't). For clarity, we add '-Analyst' at the end of first/last names for staff users so we can easily differentiate them when doing fixtures, testing, etc.

zandercymatics (approved):
Not going to block your merge because there's just a tiny change needed here before commit, looks good otherwise!

Issue #935, opened 2023-08-25
UX: Review of Django Admin


PaulKuykendall:
### Issue Description

Design team, please review Django Admin, through all current functionality, proposing guidance and refinements for usability. (Heuristics evaluation). 

Starting HERE: https://getgov-staging.app.cloud.gov/admin/

Scope: Use site from the analyst perspective, executing test cases and MVP functionality, looking for usability, consistency, and logical workflows. This is in preparation for putting this in the Analysts for feedback and further iteration.

Relates to: #887 (test cases)

AC:

- [x] Walked through all applicable test cases for DjA tool.
- [x] Updated Test plan based on Discoveries during this exercise.
- [x] Set of proposed UX/UI and workflow Changes (tickets) to be prioritized, including relative value/impact of the proposed changes (like categorizing a bug, for example)

### Additional Context

[DjAdmin UX Review document](https://docs.google.com/document/d/1HmF_jCOcYUEj_gBd5NMuVuSQi6a85TLTl0Mk7uFtZGs/edit#heading=h.68g1pvvfauc) - created by Katherine and Rachid

PR #1003, opened 2023-09-08
1000 admin ux audit implementation of quick fixes

### LINKED ISSUE #1000: First dev passthrough on the admin UX audit
### Issue Description

1. Implement quick fixes from the [UX audit document](https://docs.google.com/document/d/1HmF_jCOcYUEj_gBd5NMuVuSQi6a85TLTl0Mk7uFtZGs/edit#heading=h.68g1pvvfauc)
2. Quick LOE on all other items

Items implemented:
- Add a favicon
- Fix button on “Are You Sure” pages
- Fix button on “view-only” pages
- Remove table caption links
- Log Entries: 
  - Remove “Resource” links
  - Add search helper text
  - Remove “Save” buttons
- Add email address to contact table
- Domain Information
  - Add search field
  - Add additional columns
  - Add filters
- Domain Invitation
  - Add additional columns
  - Add search
- Domains
  - Add additional columns
  - Add filters
- User Domains Roles
  - Add additional columns
  - Add a search field
- Add a search field to websites 

_No response_

### Issue Links

Relates to #935 

UX Audit: [document](https://docs.google.com/document/d/1HmF_jCOcYUEj_gBd5NMuVuSQi6a85TLTl0Mk7uFtZGs/edit#heading=h.68g1pvvfauc)

rachidatecs:
## Ticket

Resolves #1000

Related to #935 

UX Audit: [document](https://docs.google.com/document/d/1HmF_jCOcYUEj_gBd5NMuVuSQi6a85TLTl0Mk7uFtZGs/edit#heading=h.68g1pvvfauc)

## Changes

- Add a favicon
- Fix button on “Are You Sure” pages
- Fix button on “view-only” pages
- Remove table caption links
- Log Entries: 
  - Remove “Resource” links
  - Add search helper text
  - Remove “Save” buttons
- Add email address to contact table
- Domain Information
  - Add search field
  - Add additional columns
  - Add filters
- Domain Invitation
  - Add additional columns
  - Add search
- Domains
  - Add additional columns
  - Add filters
- User Domains Roles
  - Add additional columns
  - Add a search field
- Add a search field to websites 

## Context for reviewers

Only non-straightforward part is the LogEntry override, see [repo link](https://github.com/jazzband/django-auditlog/blob/master/auditlog/admin.py)

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### DIFF
--- a/src/registrar/admin.py
+++ b/src/registrar/admin.py
@@ -6,8 +6,19 @@ from django.http.response import HttpResponseRedirect
 from django.urls import reverse
 from registrar.models.utility.admin_sort_fields import AdminSortFields
 from . import models
+from auditlog.models import LogEntry # type: ignore
+from auditlog.admin import LogEntryAdmin # type: ignore
+
 
 logger = logging.getLogger(__name__)
+
+class CustomLogEntryAdmin(LogEntryAdmin):
+    # Overwrite the generated LogEntry admin class
+    
+    search_help_text = "Search by resource, changed field, or user."
+    
+    change_form_template = 'admin/change_form_no_submit.html'
+    add_form_template = 'admin/change_form_no_submit.html'
 
 
 class AuditedAdmin(admin.ModelAdmin, AdminSortFields):
@@ -400,6 +411,8 @@ class DomainApplicationAdmin(ListHeaderAdmin):
         return super().change_view(request, object_id, form_url, extra_context)
 
 
+admin.site.unregister(LogEntry)  # Unregister the default registration
+admin.site.register(LogEntry, CustomLogEntryAdmin)
 admin.site.register(models.User, MyUserAdmin)
 admin.site.register(models.UserDomainRole, AuditedAdmin)
 admin.site.register(models.Contact, ContactAdmin)
--- a/src/registrar/admin.py
+++ b/src/registrar/admin.py
@@ -8,14 +8,33 @@ from registrar.models.utility.admin_sort_fields import AdminSortFields
 from . import models
 from auditlog.models import LogEntry # type: ignore
 from auditlog.admin import LogEntryAdmin # type: ignore
-
+from django.contrib.auth import get_user_model
 
 logger = logging.getLogger(__name__)
 
 class CustomLogEntryAdmin(LogEntryAdmin):
     # Overwrite the generated LogEntry admin class
+    # list_select_related = ["content_type", "actor"]
+    list_display = [
+        "created",
+        "custom_resource_url",
+        "action",
+        "msg_short",
+        "user_url",
+    ]
+    # search_fields = [
+    #     "content_type",
+    #     "object_repr",
+    #     "changes",
+    # ]
     
-    search_help_text = "Search by resource, changed field, or user."
+    def custom_resource_url(self, obj):
+        # Replace 'my_field' with the name of your field
+        return f"{obj.content_type} - {obj.object_repr}"  # Return the field value without a link
+    
+    custom_resource_url.short_description = 'Resource'
+    
+    search_help_text = "Search by resource, changes, or user."
     
     change_form_template = 'admin/change_form_no_submit.html'
     add_form_template = 'admin/change_form_no_submit.html'
--- a/src/registrar/admin.py
+++ b/src/registrar/admin.py
@@ -6,38 +6,34 @@ from django.http.response import HttpResponseRedirect
 from django.urls import reverse
 from registrar.models.utility.admin_sort_fields import AdminSortFields
 from . import models
-from auditlog.models import LogEntry # type: ignore
-from auditlog.admin import LogEntryAdmin # type: ignore
-from django.contrib.auth import get_user_model
+from auditlog.models import LogEntry  # type: ignore
+from auditlog.admin import LogEntryAdmin  # type: ignore
 
 logger = logging.getLogger(__name__)
 
+
 class CustomLogEntryAdmin(LogEntryAdmin):
-    # Overwrite the generated LogEntry admin class
-    # list_select_related = ["content_type", "actor"]
+    """Overwrite the generated LogEntry admin class"""
+
     list_display = [
         "created",
-        "custom_resource_url",
+        "resource",
         "action",
         "msg_short",
         "user_url",
     ]
-    # search_fields = [
-    #     "content_type",
-    #     "object_repr",
-    #     "changes",
-    # ]
-    
-    def custom_resource_url(self, obj):
-        # Replace 'my_field' with the name of your field
-        return f"{obj.content_type} - {obj.object_repr}"  # Return the field value without a link
-    
-    custom_resource_url.short_description = 'Resource'
-    
+
+    # We name the custom prop 'resource' because linter
+    # is not allowing a short_description attr on it
+    # This gets around the linter limitation, for now.
+    def resource(self, obj):
+        # Return the field value without a link
+        return f"{obj.content_type} - {obj.object_repr}"
+
     search_help_text = "Search by resource, changes, or user."
-    
-    change_form_template = 'admin/change_form_no_submit.html'
-    add_form_template = 'admin/change_form_no_submit.html'
+
+    change_form_template = "admin/change_form_no_submit.html"
+    add_form_template = "admin/change_form_no_submit.html"
 
 
 class AuditedAdmin(admin.ModelAdmin, AdminSortFields):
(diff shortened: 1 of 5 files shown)


src/registrar/admin.py
```diff
@@ -222,11 +256,97 @@ class ContactAdmin(ListHeaderAdmin):
...
+    # This gets around the linter limitation, for now.
+    def contact(self, obj):
+        if obj.first_name or obj.last_name:
+            return obj.get_formatted_name()
+        elif obj.email:
+            return obj.email
+        elif obj.pk:
+            return str(obj.pk)
+        else:
+            return ""
```
zandercymatics, line 275:
```suggestion
            return None
```
(Optional) Error handling/functionality wise, there is a possibility for things to get a bit buggy when you return empty strings as opposed to None. I'd say this is fine to keep if you need this down the line (i.e. using it in an HTML file), but returning None is a bit more predictable imo if we're using that

src/registrar/admin.py
```diff
@@ -222,11 +256,97 @@ class ContactAdmin(ListHeaderAdmin):
...
+
+    # We name the custom prop 'contact' because linter
+    # is not allowing a short_description attr on it
+    # This gets around the linter limitation, for now.
+    def contact(self, obj):
+        if obj.first_name or obj.last_name:
+            return obj.get_formatted_name()
+        elif obj.email:
+            return obj.email
+        elif obj.pk:
```
zandercymatics, line 272: Note that this will not execute if obj.pk == 0. For integers, a truthy/falsy shorthand will return false if that integer is equal to 0. Here is a good resource which explains this in depth:
https://www.freecodecamp.org/news/truthy-and-falsy-values-in-python/

src/registrar/admin.py
```diff
@@ -222,11 +256,97 @@ class ContactAdmin(ListHeaderAdmin):
...
+    list_display = [
+        "contact",
+        "email",
+    ]
+
+    # We name the custom prop 'contact' because linter
+    # is not allowing a short_description attr on it
+    # This gets around the linter limitation, for now.
+    def contact(self, obj):
+        if obj.first_name or obj.last_name:
```
zandercymatics, line 268: Truthy/Falsy values are fine to use but its generally recommended that you explicitly define the type you want to handle (for example, you may want to add a type on obj). These shorthands will return false in the following conditions:
image
Basically, if for some reason a string gets passed in as a number, python may still inevitably parse that as a string once its rendered but you could run into buggy behavior. For instance, if obj.first_name was parsed as an integer of '0', then this field would never execute as while the first name is technically "0", python may assume otherwise

src/registrar/assets/sass/_theme/_admin.scss
```diff
@@ -158,8 +158,10 @@ table > caption > a {
     padding-top: 20px;
 }
 
-// 'Delete button' layout bug
-.submit-row a.deletelink {
+// Fix django admin button height bugs
+.submit-row a.deletelink,
+.delete-confirmation form .cancel-link,
```
zandercymatics, line 163:
```suggestion
.delete-confirmation form.cancel-link,
```
Did you mean to add a space here? This will apply to all `<form>`(s) otherwise

src/registrar/templates/admin/change_form_no_submit.html
```diff
@@ -0,0 +1,20 @@
+{% extends "admin/change_form.html" %}
+
+{% comment %} Replace the Django ul markup with a div. We'll edit the child markup accordingly in change_form_object_tools {% endcomment %}
+{% block object-tools %}
+{% if change and not is_popup %}
```
zandercymatics, line 5: Do we want to display anything here when if not change or is_popup? I don't think we do from a past conversation (grabbed from a template?) but if I am misremembering then that may need to be handled.

zandercymatics (commented):
Great changes so far! Below are a couple of code things I noticed while going through here. I'll leave another comment after running your code locally

src/registrar/admin.py
rachidatecs: I want to return something here otherwise the list table may break. I'll test what None does. If the table does not break, I'll accept your change. If the table breaks, I'd rather stick with an empty string.

src/registrar/admin.py
rachidatecs: 2 notes:
- The DB API uses a 'bool(pk_val) == false' check to determine if a model instance is a new item needs to be INSERTed, or an existing entry that needs to be UPDATEd.
Most databases use PK's starting at 1, but if you try to manually use a pk value of 0, the object cannot be saved - it can only be recreated.

- This code is a duplicate of the _str_ method in the contact model

src/registrar/templates/admin/change_form_no_submit.html
rachidatecs: I wouldn't mess with it (at least not in this PR). This is how the [original template](https://github.com/django/django/blob/a7c73b944f51d6c92ec876fd7e0a171e7c01657d/django/contrib/admin/templates/admin/change_form.html#L27) is written in django.

src/registrar/admin.py
rachidatecs: obj is an instance of Contact, and I'm comparing against its props (first_name and last_name are models.TextField, etc.) Don't all these types trickle down into obj in the admin class when you link admin.site.register(models.Contact, ContactAdmin)?

src/registrar/admin.py
zandercymatics: @rachidatecs If the table expects an empty string I'd say it makes the most sense to keep that - agreed. Not worth it if it breaks things

src/registrar/admin.py
zandercymatics: @rachidatecs Ah, I was not aware that this is not a possible state for Django. Lack of knowledge on my part, I think this is a non issue then.

src/registrar/templates/admin/change_form_no_submit.html
zandercymatics: Yeah lets conform with the template for now

src/registrar/admin.py
zandercymatics: @rachidatecs That is correct, I think I misunderstood the purpose of the function. I'd assume your using this as a workaround as it doesn't allow you to call and/or invoke .__str__ directly. My original assumption was moreso that this could be used by a developer and they (may) pass in the wrong type but given the function is literally called 'contact' I think that would be implied (or may try to use it on an epp Contact object, for instance)... 

With that context in mind, what your doing here is fine. A nice to have might be to add a type hint on obj (obj: Contact) if it doesn't break anything just as a precaution, but really not necessary especially in this context.

zandercymatics:
### Fix button on “Are You Sure” pages
- (Could be separated in a different PR) There is a font size difference between Yes, I'm sure and No, take me back
image
### Remove “Save” buttons
- No issues with this implementation. One thing to note on this would be that, technically, there would be nothing stopping a malicious admin user from sending a mocked post request and editing anyway, but we are assuming that: A) they will be a small pool of trusted users and B) they ought to have this access anyway because of what their role entails
### Domain Information
- Add filters: This isn't exclusive to here (_different PR_) but its a little hard to visually differentiate between each of the filter items due to the length of the text. I think there are two options here: 1. Add some kind of 'on hover' effect (like hovering over an <a> tag) or 2. as an admin/analyst will likely already know, or at least be aware of, these distinctions so I personally would recommend to keeping these filters to just their names rather than name + description (Federal, Interstate, State) 
image

zandercymatics:
Something to note as well would be that, while using ANDI, I was able to find that some of the added columns are not sortable while others are (note the ones that are highlighted vs not)

image

zandercymatics (changes requested):
Really throughout, great work! The only issue that needs to be addressed is that the "organization" type field under Domain should be sortable (as it currently isn't). You are good to merge after you get that resolved

rachidatecs:
I'll fix the font size issue on the buttons and the sortable column. I recommend tackling the proposed hove effect on filter options in a follow-up quick fix, as it may hold the PR

Katherine-Osos:
Regarding the filters for "Domain Information that @zandercymatics mentioned above. There's already a ticket to shorten the filter names here: https://github.com/cisagov/getgov/issues/787
It just needs to get prioritized.

zandercymatics (approved):

zandercymatics:
LGTM! I see the hover color issue you mentioned, good inclusion

Katherine-Osos:
@rachidatecs This looks great. Just one update: On the "Domains" page, can you also add a filter for "State"? That will be useful once the states are implemented.

Katherine-Osos (approved):

rachidatecs:
New ticket to handle the extra requests: #1017

PR #1022, opened 2023-09-13
Draft (IGNORE) - Za/850 epp contact get

zandercymatics:
### !!DRAFT PR FOR TESTING PURPOSES!!
Ignore this draft. Will be deleted
cf push getgov-za -f ops/manifests/manifest-za.yaml
cf ssh getgov-za
/tmp/lifecycle/shell
./manage.py shell

domain, _ = Domain.objects.get_or_create(name="mydomain.gov")

print("security")
print(domain.security_contact)
print("admin")
print(domain.administrative_contact)
print("tech")
print(domain.technical_contact)
print("registrant")
print(domain.registrant_contact)


abroddrick:
@zandercymatics should the other branch be closed now? seems like this one is more up to date but figured I would ask

src/registrar/models/domain.py
```diff
@@ -668,6 +654,210 @@ def __str__(self) -> str:
...
+        Returns None if nothing is found.
+        check_type compares contact["type"] == check_type.
+
+        For example, check_type = 'security'
+        """
+        for contact in contacts:
+            print(f"grab_contact_in_keys -> contact item {contact.__dict__}")
+            if (
+                isinstance(contact, PublicContact)
+                and contact.registry_id is not None
```
abroddrick, line 852: why this? if registry_id is none should an error be thrown instead?

src/registrar/models/domain.py
```diff
@@ -954,19 +1139,26 @@ def _fetch_cache(self, fetch_hosts=False, fetch_contacts=False):
...
 
             # statuses can just be a list no need to keep the epp object
             if "statuses" in cleaned.keys():
                 cleaned["statuses"] = [status.state for status in cleaned["statuses"]]
-            # get contact info, if there are any
+
+            # Registrant should be of type PublicContact
+            if "registrant" in cleaned.keys():
+                # For linter...
+                _ = cleaned["registrant"]
```
abroddrick, line 1152: This is interesting, what error did linter throw? Can you instead do self._registrant_to_public_contact(cleaned["registrant"])

src/registrar/models/domain.py
```diff
@@ -1012,33 +1188,66 @@ def _fetch_cache(self, fetch_hosts=False, fetch_contacts=False):
...
-                    }
-                    cleaned["hosts"].append(
-                        {k: v for k, v in host.items() if v is not ...}
-                    )
-
+                    # For reviewers - slight refactor here
+                    # as we may need this for future hosts tickets
+                    # (potential host mapper?).
+                    # Can remove if unnecessary
+                    cleaned["hosts"].append(self._get_host_as_dict(name))
```
abroddrick, line 1195: why was hosts updated here? Did a bug occur? Just asking as this will affect @therealslimhsiehdy's ticket and may need to be factored in.

abroddrick (commented):
Sorry for snooping, just adding a couple comments, mostly curious on the host change.

src/registrar/models/domain.py
zandercymatics: Just a line length error, I think I may just rename the function to something shorter... Its an item for code cleanup

src/registrar/models/domain.py
zandercymatics: I think I'll remove this particular change, I was initially thinking this could've been impacted by the EPPLib stuff (like contacts where you have to deal with weird objects) so I was just getting some stuff moved over for that, but that didn't turn out to be an issue. Code leftover pretty much

src/registrar/models/domain.py
zandercymatics: That is a good point. Throwing an error would be better here

Issue #1034, opened 2023-09-18
Modify email content for migration domain invitation script


kitsushadow:
### Issue Description

The email content for domain invitations requires modification to be tailored for the migration script domain invitations. 

Considerations:
- For users with more than one domain, only one email will be sent but it will include the links to all their domains.
- This script is a one time design for use with the migration. Emails should not be expected to be resent as a part of the migration process.

_**NOTE:** Wait for feedback from user testing before starting this ticket_

**Acceptance Criteria**

- [x] update the current [domain invitation email](https://docs.google.com/document/d/1HnMdniAB9INphrLwKjiACyNLrenyXy0QFf0-yz4EStI/edit#heading=h.gjdgxs) based on feedback from user testing
- [x] then, create an alternate version of that domain invitation email to be used for the migration. Add new verbiage to communicate that we're updating to new registrar (i.e. explain why domain managers are receiving this email when they are already a manager)
- [x] ensure engineering tickets are created to make these updates

### Issue Links

Relates to: #927
Relates to: #482
Blocks: #1047

gabydisarli:
Copy is being drafted here https://docs.google.com/document/d/1bSy6lNkoImlO3rG7kzgZQy1nVubXmEDmmMHyhuOuH6g/edit#heading=h.gjdgxs

gabydisarli:
Approved!

PR #1038, opened 2023-09-18
add scripts, templates and settings for sending one-time domain invitations during data migration

### LINKED ISSUE #927: Data Migration: Domain Invitation Email send
### Issue Description

We will be sending emails separate of data ingestions. We want to avoid inviting already invited users while being able to update data from the to-end registry.
- For users with more than one domain, only send one email but include the links to all their domains not just the single domain.
- This script is a one time design for use with the migration. 
- This script should be able to be run multiple times with new sets of data, if the domain invitation has already been sent then it should NOT re-send for that email address 
  - The email is sent when transition_domain.email_sent=False. It should update this field in the table as it sends the emails. 
  - Ensure this is true even in the case where one user has multiple domains (all those domain rows should now have the same flag switched
- This should be VERY well tested. Include tests in the script or in a supporting script. Do not use it in production or with the actual emails from our host data until we are ready to execute the user Migration into our system. Test with fake emails or emails from team members.

NOTE: Use the current domain invitation verbiage, run it by design to ensure it will be good given that it is for the migration vs normal invitations. 
https://docs.google.com/document/d/1UkQZdAmcGprXWdHX_PuT155CCNPbf_Gf/edit?pli=1#heading=h.iqv2v8lecjpn

retest this after #923 is done with all the data from verising
### Issue Links

blocked by: #923

dave-kennedy-ecs:
## Ticket

Resolves #927 

## Changes

- script generate_test_transition_domains.py added to generate test transition_domain data
- script send_domain_invitations.py added to send emails during migration
- added setting to settings.py to reference application's BASE_URL from scripts

## Context for reviewers

Separate ticket #1034 created to modify content of invitation emails.  So, email content does not need further review on this ticket.

generate_test_transition_domains.py is for testing purposes only; it is not used during migration; it is used to generate test transition_domains when they don't exist in order to test the script; this script takes -e option to add email addresses which will be assigned to the transition_domains
(ex. docker-compose exec app python manage.py generate_test_transition_domains -e email1,email2,email3)

send_domain_invitations.py is used to send domain invitations to the appropriate users, and mark the emails as sent in the transition_domain objects with email_sent=True; it can be run without actually sending emails and updating database by supplying no runtime argumens; to send emails and update database use the -s option
(ex. docker-compose exec app python manage.py send_domain_invitations -s)

## Setup

In your environment, run the following command to generate test transition_domains:
docker-compose exec app python manage.py generate_test_transition_domains -e your_email1,your_email2

You can then verify the transition domains in the admin interface in django.  Note that email_sent is unchecked for all transition_domains

Then run the script to send emails:
docker-compose exec app python manage.py send_domain_invitations -s

Check your email to verify that emails were received.
Check the admin interface in django and verify that email_send is checked for all transition_domains

Script can be repeated.  You can uncheck email_sent in admin interface and run again to resend emails.
You can also set the domain name in transition_domain to an invalid value to test error conditions when running script.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

## Screenshots

(added 9/20/23)

Updated example below for single domain:

image

Updated example below for multiple domains:

image


src/registrar/management/commands/generate_test_transition_domains.py
```diff
@@ -0,0 +1,60 @@
...
+            required=True,
+            dest="emails",
+            help="Comma-delimited list of email addresses to be used for testing",
+        )
+
+    def handle(self, **options):
+        """Delete existing TransitionDomains.  Generate test ones."""
+
+        # split options[emails] into an array of test emails
+        test_emails = options["emails"].split(",")
```
zandercymatics, line 35: (Nice to have but optional) I'd add some error handling before this if options["emails"] is None, so we can throw a custom ValueError or something of that nature - "Parameter emails must be supplied" for example

src/registrar/management/commands/generate_test_transition_domains.py
```diff
@@ -0,0 +1,60 @@
...
+    # Transition domains are created with email addresses provided as command line
+    # argument. Email addresses for testing are passed as comma delimited list of
+    # email addresses, and are required to be provided.  Email addresses from the list
+    # are assigned to transition domains at time of creation.
+
+    def add_arguments(self, parser):
+        """Add command line arguments."""
+        parser.add_argument(
+            "-e",
+            "--emails",
```
zandercymatics, line 25: I like this approach

src/registrar/management/commands/generate_test_transition_domains.py
```diff
@@ -0,0 +1,60 @@
...
+        # Need to get actual domain names from the database for this test
+        real_domains = Domain.objects.all()
+        for real_domain in real_domains:
+            TransitionDomain.objects.create(
+                username=test_emails[test_emails_counter % len(test_emails)],
+                domain_name=real_domain.name,
+                status="created",
+                email_sent=False,
+            )
+            test_emails_counter += 1
```
zandercymatics, line 55: I'd add some error handling in this function for test_emails (values of None for instance).

src/registrar/management/commands/generate_test_transition_domains.py
```diff
@@ -0,0 +1,60 @@
...
+        """Delete existing TransitionDomains.  Generate test ones."""
+
+        # split options[emails] into an array of test emails
+        test_emails = options["emails"].split(",")
+
+        # setting up test data
+        self.delete_test_transition_domains()
+        self.load_test_transition_domains(test_emails)
+
+    def load_test_transition_domains(self, test_emails):
```
zandercymatics, line 41:
```suggestion
    def load_test_transition_domains(self, test_emails: list):
```
This doesn't have to be 'list' specifically if it breaks your function (just assuming the type), but type hinting here would be a good addition

src/registrar/management/commands/send_domain_invitations.py
```diff
@@ -0,0 +1,148 @@
...
+        if options["send_emails"]:
+            logger.debug("about to send emails")
+            self.send_emails()
+            logger.debug("done sending emails")
+
+            self.update_domains_as_sent()
+
+            logger.debug("done sending emails and updating transition_domains")
+        else:
+            logger.debug("not sending emails")
```
zandercymatics, line 62: Given the context of this, I'd recommend changing these to be logger.info or (for the else) logger.error. I think this information makes sense to display to the end user

src/registrar/management/commands/send_domain_invitations.py
```diff
@@ -0,0 +1,148 @@
...
+            # domains_with_errors
+            try:
+                domain = Domain.objects.get(name=transition_domain.domain_name)
+                # if prior username does not match current username
+                if (
+                    not email_context["email"]
+                    or email_context["email"] != transition_domain.username
+                ):
+                    # if not first in list of transition_domains
+                    if email_context["email"]:
```
zandercymatics, line 86: This check seems redundant as you are checking for `not email_context["email"]` above... Your comment mentions checking for (not) the first item in the email array, did you mean to do that check here instead?

src/registrar/management/commands/send_domain_invitations.py
```diff
@@ -0,0 +1,148 @@
...
+            except Exception as err:
+                # error condition if domain not in database
+                self.domains_with_errors.append(
+                    copy.deepcopy(transition_domain.domain_name)
+                )
+                logger.error(
+                    f"error retrieving domain {transition_domain.domain_name}: {err}"
+                )
+        # if there are at least one more transition domains than errors,
+        # then send one more
```
zandercymatics, line 107:
```suggestion
        # then append one more item
```
This seemed a bit unclear to me, I would recommend something like above

src/registrar/management/commands/send_domain_invitations.py
```diff
@@ -0,0 +1,148 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = "Send domain invitations once to existing customers."
+
+    # this array is used to store and process the transition_domains
+    transition_domains: List[str] = []
+    # this array is used to store domains with errors, which are not
```
zandercymatics, line 21: This is a neat approach, I like this

src/registrar/management/commands/send_domain_invitations.py
```diff
@@ -0,0 +1,148 @@
...
+                    if email_context["email"]:
+                        # append the email context to the emails_to_send array
+                        self.emails_to_send.append(copy.deepcopy(email_context))
+                    email_context["domains"] = []
+                email_context["email"] = transition_domain.username
+                email_context["domains"].append(
+                    {
+                        "name": transition_domain.domain_name,
+                        "url": settings.BASE_URL
+                        + reverse("domain", kwargs={"pk": domain.id}),
```
zandercymatics, line 95: (Tangent/Unrelated) Not applicable here but when researching if we can grab the entire url within reverse, I found this which may be useful to you (hopefully) in future for django views if you do work on those: https://docs.djangoproject.com/en/2.0/ref/request-response/#django.http.HttpRequest.build_absolute_uri

src/registrar/management/commands/send_domain_invitations.py
```diff
@@ -0,0 +1,148 @@
...
+                logger.error(
+                    f"error retrieving domain {transition_domain.domain_name}: {err}"
+                )
+        # if there are at least one more transition domains than errors,
+        # then send one more
+        if len(self.transition_domains) > len(self.domains_with_errors):
+            self.emails_to_send.append(email_context)
+
+    def send_emails(self):
+        for email_data in self.emails_to_send:
```
zandercymatics, line 112: (Nice to have / Optional) I'd add an if check here for when emails_to_send is empty, something like `if len(emails_to_send) == 0: logger.info("No emails to send")`

src/registrar/management/commands/send_domain_invitations.py
```diff
@@ -0,0 +1,148 @@
...
+        try:
+            send_templated_email(
+                "emails/transition_domain_invitation.txt",
+                "emails/transition_domain_invitation_subject.txt",
+                to_address=email_data["email"],
+                context={
+                    "domains": email_data["domains"],
+                },
+            )
+            # if log level set to debug, success message is logged
```
zandercymatics, line 125: Ah, I see why you may be doing logger.debug in your handle function. Will this script always be restricted to debug only, or would you like to only restrict information but allow it to run otherwise?

src/registrar/templates/emails/transition_domain_invitation.txt
```diff
@@ -0,0 +1,11 @@
+You have been invited to manage {% if domains|length > 1 %}multiple domains{% else %}the domain {{ domains.0.name }}{% endif %} on get.gov,
```
abroddrick, line 1: @dave-kennedy-ecs I spoke with @Katherine-Osos  today and since @therealslimhsiehdy updated the domain_invitation text in [her newest pr](https://github.com/cisagov/getgov/pull/1026/files#diff-6837eab823a4ed83fd1e89c91d8a372ee3e5b8c8d8c2ba2109c8c8d565af3957), she would like to match the verbiage in there as much as possible. Mind updating this text to follow that PR accordingly?  Feel free though to wait until that PR is merged before updating this.

src/registrar/management/commands/generate_test_transition_domains.py
dave-kennedy-ecs: Thank you.  I considered this, but used required=True on the parser.add_argument, which essentially does the same thing.

src/registrar/management/commands/generate_test_transition_domains.py
```diff
@@ -0,0 +1,60 @@
...
+        parser.add_argument(
+            "-e",
+            "--emails",
+            required=True,
+            dest="emails",
+            help="Comma-delimited list of email addresses to be used for testing",
+        )
+
+    def handle(self, **options):
+        """Delete existing TransitionDomains.  Generate test ones."""
```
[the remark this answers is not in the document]
dave-kennedy-ecs: Expanded the comment.  And handling logic in the parsing of command line arguments.  I do not think this script will be slated for reuse.

src/registrar/management/commands/generate_test_transition_domains.py
dave-kennedy-ecs: thanks

src/registrar/management/commands/send_domain_invitations.py
dave-kennedy-ecs: I intended this check to be here.  In this code, there is an ordered array of domains, sorted by username.  From that array, I create an array named emails_to_send of email_context to be sent, including data from one or more items from the first array in each item in the second array.  I have written checks that indicate when to add items to the second array.   The data to be added to the second array is stored in email_context while the domains array is looped through.

The first check indicates that the prior username in email_context does not match current username on the current item in the domains array.
The second check indicates that there is email_context to add to the second array.  This will not be the case when the first item in the domains array is processed.

src/registrar/management/commands/send_domain_invitations.py
dave-kennedy-ecs: thanks

src/registrar/management/commands/send_domain_invitations.py
dave-kennedy-ecs: Thank you.  Yes, I found this too, but request not available in this context.  So added to settings.py.

src/registrar/management/commands/send_domain_invitations.py
dave-kennedy-ecs: good suggestion.  i incorporated.

src/registrar/management/commands/send_domain_invitations.py
dave-kennedy-ecs: rethinking this.  i used debug intentionally so that this script could be run without the noise, and only errors would print.  but since the successful information is important to log for accountability, i have modified all of the debugs to infos.

src/registrar/management/commands/generate_test_transition_domains.py
dave-kennedy-ecs: i have now added the additional error handling to handle, and since redundant, did not add here as well.

src/registrar/templates/emails/transition_domain_invitation.txt
dave-kennedy-ecs: @abroddrick @Katherine-Osos @therealslimhsiehdy I reviewed the pr for 482 yesterday, and I have updated the text.  I added some screenshots to the comment.

Note that this updated text for the email simplifies the code, as I no longer need to add individual urls for each domain to the email.  If you approve these updated screenshots, I will go back and simplify the code on this PR just a bit.

src/registrar/management/commands/send_domain_invitations.py
zandercymatics: Ah, yup, I see your or statement now. Thanks for pointing that out - I misread it as 'and' so I assumed it would never execute. That check makes sense to do, especially given the context you've given

zandercymatics (approved):
LGTM!

abroddrick:
@zandercymatics did you run this locally and get it to work correctly? I don't see that checked off, and I want to make sure that was done here.

abroddrick (approved):
I will  approve as well.  These code comments are amazingly helpful and I want to say that your PR had one of the best descriptions of setup and what was done that I have seen yet. I really appreciate that as it made it much easier to review.

I know we are printing the domains and and emails when they fail, but at the end of the run could we add a print of the list of emails that triggered errors? I realize this may mean another list but it could be very helpful when trouble shooting in a production scenario.

zandercymatics:
@abroddrick Ah my apologies! I did yesterday and it worked well, just spaced that portion

PR #1048, opened 2023-09-20
Tickets #850, #912 - Epp Contact Getters / Security Contact Getter+Setter

### LINKED ISSUE #850: EPP Contact Get
### Issue Description

Get all contact registry information with tests. 

Retrieving Administrative contact information | Blocked by: #809
Retrieving Security email address | Blocked by: #809 
Retrieving Registrant Contant info | Blocked by: #809 
Retrieving Tech contact | Blocked by: #809 

When getting contacts if we don't have their public contact we should make a new public contact to match with that ID. This should only be relevant for transition domains. 

_No response_

### Issue Links

Blocked by: #809

### LINKED ISSUE #912: Get and Post the default security Email (EPP)
### Issue Description

On the domain management Security email page, have the security page pull from EPP and display to the security email to a user. When a user saves the email it should set the email on epp. 
AC:
- The default security email doesn't show to Domain Registrant. (see public contact default_security_contact)
-The security email page calls the correct domain.py command for setting the email address
- security email page pulls from domain.py to get the current security email
-remove the Set New Security Contact and Remove Secruity contact buttons in domain_change_form.html
-remove the corresponding SET_SECURITY_CONTACT and GET_SECURITY_EMAIL from admin.py (use their logic as examples for implementation)
-add unit tests accordingly (don't replicate what is already in test_models_domains.TestRegistrantContacts)

out of scope:setting the default security contact is not in scope and is set when the domain is created in epp. Expect all domains to have a default security email at start up

implementation: 
All by the removing buttons, not displaying the default are done, and added tests are done. Consider this a ticket for hardening up the existing code.
_No response_

### Issue Links
blocked by : #850 
_No response_

zandercymatics:
## Ticket

Resolves #850
Resolves #912  

## Changes

- Implemented getters for registrant_contact, security_contact, administrative_contact, and technical_contact using a `generic_contact_getter()` function
- Implemented caching behavior for "registrant" and "contacts"
- Modified `_add_registrant_to_existing_domain()` to account for duplicate PublicContacts
- Created a `PublicContact` mapper which deals with the 'InfoContactResultData' object returned from the registry
- Moved pre-existing logic for the `commands.InfoContact` call into a function called `_request_contact_info` for reusability
- Added getters for default contact types (security, admin, etc) in domain.py. Previously only had one for security_contact
- Created a `grab_contact_in_keys` function that can parse through a list of cached contacts and get the appropriate type we want
- Modified `_fetch_cache` so that it does a call to EPP to grabs InfoContact on registrant and other contacts. It then converts that into a PublicContact object and stores it on the cache (old method was dealing with strings / incorrect EPP objects)
-  Added a ` _registrant_to_public_contact` function which converts the string representation we get from EPP (just the ID) into a PublicContact object
- Modified the 'Add Security Email' to change to 'save' after it is edited
- Created a `dummyInfoContactResultData` helper in common.py which is used for creating a mock "InfoContactResultData" object (for testing contacts returned from EPPLib)
- Dummy data for tests with domains, modified other tests so that they are no longer using strings to represent objects but rather the appropriate class returned from EPPLib
- Tests for the getters / setters + modified existing related ones
- ~Fix a test interference bug with `TestDomainCreation`~ Removed - got fixed by another merge (test case change)  
- Added an import for 'Info' under EppLibWrapper (For 'result' objects, as those do not exist under 'common')

### Notes:
This is not fully testable locally as EPPLib is only contactable on cloud.gov. If you do not have access to my sandbox, you can pull this branch and do a cf push onto your own for testing purposes - or I can add you as an admin.
### Testing #912:
Navigate to any approved domain and click "Manage". Click security email and make alterations. These should persist on EPPLib and the registrar. You can verify it on EPPLib by using this set of commands:
**Creating the shell:**
cf ssh getgov-za
/tmp/lifecycle/shell
./manage.py shell
**Then run these inside the Python shell you created:**

domain, _ = Domain.objects.get_or_create(name="**name here**")

print(domain.security_contact.email)
^^This should match the UI^^

**If you'd like to verify with InfoContact:**
request = commands.InfoContact(id='**id_of_security_contact**')
response = registry.send(request, cleaned=True)
print(response)

You can also verify this through /admin
### Testing #850:
Same as above, but print the content for the following...
print(domain.security_contact)
print(domain.administrative_contact)
print(domain.technical_contact)
print(domain.registrant_contact)

You may need to set the value of these first. To do that, you can either user /admin or you can create it manually. To do so, set the contact to `your_edit = domain.get_default_{contact_type}_contact()`, edit the email, then set `domain.security_contact = your_edit` (or set your_edit.domain = domain)

You can also verify these through /admin by making a change, and verifying that the change was saved and is displaying correctly by using the InfoDomain or InfoContact commands
### Other notes:
As we rely on EPPLib to both set / get data for contacts, there is a performance penalty involved with this. As it is now, our setter will regenerate the entire cache when the object is accessed for the first time, or everytime the object is updated. As a result, this means that there is an unfavorable amount of loading time when accessing a domain. Once we can grab straight from our DB and/or cache, these issues dissipate but we do need an architectural change to resolve that main issue>
**With that in mind:** expect some performance issues when clicking a domain for the first time. This should not be as significant after everything is in cache so if it is, then that points towards something weird. 

- [x] Tested with multiple browsers, the suggestion is to use ones that the developer didn't (check off which ones were used)
  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


ops/scripts/manifest-sandbox-template.yaml
zandercymatics, line 1: This will be taken out after merge, just here to prevent crashing when testing

ops/scripts/manifest-sandbox-template.yaml
```diff
@@ -11,6 +11,7 @@ applications:
   command: ./run.sh
   health-check-type: http
   health-check-http-endpoint: /health
+  health-check-invocation-timeout: 30
```
zandercymatics, line 14: This will be taken out after merge, just here to prevent crashing when testing

zandercymatics (commented):
See below

src/registrar/tests/test_models_domain.py
```diff
@@ -81,17 +88,16 @@ def test_cache_used_when_avail(self):
...
+        # this is then transformed into PublicContact, and cache should NOT
+        # hold onto the DomainContact object
+        expectedUnfurledContactsList = [
+            common.DomainContact(contact="123", type="security"),
+        ]
+        expectedContactsList = [domain.security_contact]
         expectedHostsDict = {
             "name": self.mockDataInfoDomain.hosts[0],
-            "cr_date": self.mockDataInfoHosts.cr_date,
+            "cr_date": self.mockDataInfoDomain.cr_date,
```
zandercymatics, line 100: This is an error - will correct

src/registrar/models/domain.py
```diff
@@ -953,19 +1123,26 @@ def _fetch_cache(self, fetch_hosts=False, fetch_contacts=False):
...
             cleaned = {k: v for k, v in cache.items() if v is not ...}
 
             # statuses can just be a list no need to keep the epp object
             if "statuses" in cleaned.keys():
                 cleaned["statuses"] = [status.state for status in cleaned["statuses"]]
-            # get contact info, if there are any
+
+            # Registrant should be of type PublicContact
+            if "registrant" in cleaned.keys():
+                # Registrant, if it exists, should always exist in EppLib.
```
rachidatecs, line 1135: Do you mean "should always exist in the DB"?

src/registrar/models/public_contact.py
```diff
@@ -29,7 +29,8 @@ class ContactTypeChoices(models.TextChoices):
 
     def save(self, *args, **kwargs):
         """Save to the registry and also locally in the registrar database."""
-        if hasattr(self, "domain"):
+        skip_epp_save = kwargs.pop("skip_epp_save", False)
```
rachidatecs, line 32: Can possibly remove this control if we're not using it

src/registrar/templates/domain_security_email.html
```diff
@@ -21,7 +21,7 @@ <h1>Domain security email</h1>
     <button
       type="submit"
       class="usa-button"
-    >Add security email</button>
+    >{% if domain.security_contact is None or domain.security_contact.email == 'dotgov@cisa.dhs.gov'%}Add security email{% else %}Save{% endif %}</button>
```
rachidatecs, line 24: Nice!

src/registrar/tests/test_models_domain.py
```diff
@@ -103,15 +110,58 @@ def test_cache_nested_elements(self):
...
+        # The contact list should not contain what is sent by the registry by default,
+        # as _fetch_cache will transform the type to PublicContact
+        self.assertNotEqual(domain._cache["contacts"], expectedUnfurledContactsList)
+        self.assertEqual(domain._cache["contacts"], expectedContactsList)
 
         # get and check hosts is set correctly
         domain._get_property("hosts")
         self.assertEqual(domain._cache["hosts"], [expectedHostsDict])
 
-    def tearDown(self) -> None:
```
rachidatecs, line 112: Why did you remove the teardown?

rachidatecs:
Pushed and tested on my sandbox, dod not find any issues

src/registrar/models/domain.py
```diff
@@ -667,6 +649,200 @@ def __str__(self) -> str:
...
+            city=getattr(addr, "city", ""),
+            pc=getattr(addr, "pc", ""),
+            cc=getattr(addr, "cc", ""),
+            sp=getattr(addr, "sp", ""),
+            **streets,
+        )
+
+        return desired_contact
+
+    def _request_contact_info(self, contact: PublicContact):
```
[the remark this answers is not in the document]
zandercymatics: This is used in the `_registrant_to_public_contact()` function which exists on cache. The reason why we have this is because a registrant is the same type as other contacts. For registrant specifically, EppLib will return a string which is the id of the InfoContact object. For instance, if I had a contact object with id "eggs" then registrant="eggs". This isn't enough data for a PublicContact object, so we need to run the InfoContact command for more information to do the mapping. That coerced PublicContact object will then exist on the cache after it is 'converted' from its original form

Here is the code (clearer as a screenshot):
image
image

Its job is to effectively do the same thing that is being done here:
image

src/registrar/models/domain.py
zandercymatics: Removed this comment as it was unclear and did not have a value proposition

src/registrar/models/public_contact.py
zandercymatics: I think .pop removes it from the kwarg list and just returns if it existed or not. This is being used as a flag to determine if we want to do the entire EPPLib saving process or not

src/registrar/tests/test_models_domain.py
zandercymatics: Still exists! Just moved it to the beginning of the test

src/registrar/models/domain.py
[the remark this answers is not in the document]
rachidatecs: Oh so the fetch contacts in cache is divided up in 2 blocks, one for registrant and one for the other 3?

src/registrar/models/public_contact.py
rachidatecs: My point is that it seems that we're always saving (never passing skip_epp_save=False)

zandercymatics: @rachidatecs We are on line 1227 of domain.py: `public_contact.save(skip_epp_save=True)` as part of the `_get_or_create_public_contact()` function:
image

src/registrar/models/domain.py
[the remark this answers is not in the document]
zandercymatics: @rachidatecs Correct! Registrant is an edge case. It exists on a different property ("registrant" instead of "contacts") and EPPLib just returns a list of strings, instead of a list of DomainContact objects. I originally had this all contained inside one function but it just became harder to read, so I split it out

zandercymatics:
These changes are currently not on my sandbox. Shoot me a message when you'd like to test it!

src/registrar/templates/domain_security_email.html
```diff
@@ -21,7 +21,7 @@ <h1>Domain security email</h1>
     <button
       type="submit"
       class="usa-button"
-    >Add security email</button>
+    >{% if form.security_email.value is None or form.security_email.value == "dotgov@cisa.dhs.gov"%}Add security email{% else %}Save{% endif %}</button>
```
abroddrick, line 24: Did you see this in the figma from design somewhere? Or was this an ask from the design team? This feels off for our flow and normally I would just expect it to be a save button and add if we are generating a new field.

src/registrar/models/domain.py
```diff
@@ -1029,13 +1174,71 @@ def _fetch_cache(self, fetch_hosts=False, fetch_contacts=False):
...
+            ):
+                raise ValueError(
+                    "Requested PublicContact is out of sync "
+                    "with DB. Potential duplicate?"
+                )
+
+            # If it already exists, we can
+            # assume that the DB instance was updated
+            # during set, so we should just use that.
+            return existing_contact
```
abroddrick, line 1216: This worries me that we are passing this and storing it in a list. I think it should  work given manual testing but expanding your tests below would be wise here. I will comment there

src/registrar/tests/test_models_domain.py
```diff
@@ -103,15 +110,58 @@ def test_cache_nested_elements(self):
...
+        self.assertEqual(mapped.__dict__, expected_contact.__dict__)
+
+        # The mapped object should correctly translate to a DB
+        # object. If not, something else went wrong.
+        db_object = domain._get_or_create_public_contact(mapped)
+        in_db = PublicContact.objects.filter(
+            registry_id=domain.security_contact.registry_id,
+            contact_type=PublicContact.ContactTypeChoices.SECURITY,
+        ).get()
+        # DB Object is the same as the mapped object
```
abroddrick, line 163: cam we expand this test more such as modifying one object and maybe change it so you're accessing it in the list itself.

src/registrar/models/domain.py
```diff
@@ -1029,13 +1174,71 @@ def _fetch_cache(self, fetch_hosts=False, fetch_contacts=False):
...
-
             # replace the prior cache with new data
             self._cache = cleaned
 
         except RegistryError as e:
             logger.error(e)
 
+    def _get_or_create_public_contact(self, public_contact: PublicContact):
+        """Tries to find a PublicContact object in our DB.
+        If it can't, it'll create it."""
```
abroddrick, line 1185: add that it returns the PublicContat

src/registrar/templates/domain_security_email.html
zandercymatics: Yup... An ask from the design team:
https://cisa-corp.slack.com/archives/C05BDEA3C11/p1695069485950669

src/registrar/tests/test_models_domain.py
```diff
@@ -522,6 +575,128 @@ def test_update_is_unsuccessful(self):
...
+            self.mockTechnicalContact,
+            contact_id="technicalContact",
+            contact_type=PublicContact.ContactTypeChoices.TECHNICAL,
+        )
+
+        self.assertEqual(
+            self.domain_contact.technical_contact.email, expected_contact.email
+        )
+
+        # Checks if we grab the correct PublicContact...
```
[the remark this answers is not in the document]
zandercymatics: Yeah not a problem! Its just a stylistic thing but if its more confusing than anything I have no hesitation deleting them. Typically its like for documenting a chain of events that are broken up into multiple comments, such as:
Do x...
Then do y....
then do z...

src/registrar/models/domain.py
```diff
@@ -667,6 +649,183 @@ def __str__(self) -> str:
...
+        contact_id -> str: The given registry_id of the object (i.e "cheese@cia.gov")
+
+        contact_type -> str: The given contact type, (i.e. "tech" or "registrant")
+        """
+
+        if contact is None:
+            return None
+
+        if contact_type is None:
+            raise ValueError("contact_type is None")
```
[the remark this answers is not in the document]
zandercymatics: By error class, do you mean something like (with a better name) "class ContactError(Exception)"? I like that - think its a good value add

src/registrar/models/domain.py
```diff
@@ -667,6 +649,183 @@ def __str__(self) -> str:
...
+
+        if contact_id is None:
+            raise ValueError("contact_id is None")
+
+        if len(contact_id) > 16 or len(contact_id) < 1:
```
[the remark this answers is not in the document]
zandercymatics: This number comes from the definition for registry_id... Which makes me wonder if there is some built in way to do this?  To grab directly from it, I haven't tested it but I'd imagine you can do something like:
`PublicContact._meta.get_field('registry_id').max_length` but is there something built-in?
image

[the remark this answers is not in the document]
zandercymatics: This is defined by the max length of registry_id. I could do something like this:
```
@classmethod
def get_max_id_length(cls):
    return cls._meta.get_field('registry_id').max_length
```
but is there a good built in method that handles that for us and/or would be preferable?

src/registrar/models/domain.py
```diff
@@ -667,6 +649,183 @@ def __str__(self) -> str:
...
+            contacts = self._get_property(desired_property)
+        except KeyError as error:
+            # Q: Should we be raising an error instead?
+            logger.error(f"Could not find {contact_type_choice}: {error}")
+            return None
+        else:
+            # Grab from cache
+            cached_contact = self.grab_contact_in_keys(contacts, contact_type_choice)
+            if cached_contact is None:
+                raise ValueError("No contact was found in cache or the registry")
```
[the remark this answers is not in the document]
zandercymatics: Currently the page just returns a 500. This actually brings up a good question regarding this stuff:
For our application, is it preferred to display as many errors as possible to the end user for really weird edge cases, or just display a 500? My logic here was that `if cached_contact is None:` should essentially never occur, and if it does it'd we'd need to: a) address it but also b) not display that to an end user who may use it maliciously.

However this may be unnecessary in this context, what is our general assumption with this kind of stuff?

src/registrar/models/domain.py
```diff
@@ -953,19 +1104,24 @@ def _fetch_cache(self, fetch_hosts=False, fetch_contacts=False):
...
             # remove null properties (to distinguish between "a value of None" and null)
             cleaned = {k: v for k, v in cache.items() if v is not ...}
 
             # statuses can just be a list no need to keep the epp object
             if "statuses" in cleaned.keys():
                 cleaned["statuses"] = [status.state for status in cleaned["statuses"]]
-            # get contact info, if there are any
+
+            # Registrant should be of type PublicContact
+            if "registrant" in cleaned.keys():
```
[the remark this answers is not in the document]
zandercymatics: Fair enough

src/registrar/models/domain.py
```diff
@@ -953,19 +1104,24 @@ def _fetch_cache(self, fetch_hosts=False, fetch_contacts=False):
...
+                cleaned["registrant"] = self._registrant_to_public_contact(
+                    cleaned["registrant"]
+                )
+
             if (
                 # fetch_contacts and
-                "_contacts" in cleaned
+                "_contacts" in cleaned.keys()
```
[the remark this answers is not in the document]
zandercymatics: I'll get rid of this for all of these here

src/registrar/models/domain.py
```diff
@@ -974,30 +1130,19 @@ def _fetch_cache(self, fetch_hosts=False, fetch_contacts=False):
...
-                    }
+                    # Map the object we recieved from EPP to a PublicContact
+                    mapped_object = self.map_epp_contact_to_public_contact(
+                        data, domainContact.contact, domainContact.type
+                    )
 
+                    # Find/create it in the DB, then add it to the list
                     cleaned["contacts"].append(
-                        {k: v for k, v in contact.items() if v is not ...}
+                        self._get_or_create_public_contact(mapped_object)
```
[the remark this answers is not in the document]
zandercymatics: True, though the list will remain a constant size from my understanding (we can only ever have one of each type - correct if wrong) - so ideally it should be O(4). It seems counter-intuitive but for a list this small, a for loop would very likely be a bit faster. For very (and I mean _very_) tiny arrays, a for loop is actually a bit faster because from a computational standpoint, as you don't need to recompute the hashing function each time. That said the performance difference between the two is so small in that scenario that it doesn't make much of a difference. Here is a good stack overflow post which touches on the concept generally but doesn't directly relate: https://stackoverflow.com/questions/52542742/why-is-this-loop-faster-than-a-dictionary-comprehension-for-creating-a-dictionar

That said I think using a dict is fine here, but we should be using a hybrid approach rather than dictionaries all the way down as the old way did, if that makes sense.

For instance, I don't think converting the entire PublicContact object (for each object) to a dictionary then parsing/de-parsing it everytime we want to access it makes sense (essentially running a for loop on each property in PublicContact), it seems less computationally expensive to just store a list of references and directly point to that. 

As a middle ground, what do you think about this: a dictionary with the key as the type, and the value as the PublicContact object?
For instance, if we're assuming we won't expect duplicates:
```
{
"security_contact" : some_sec_contact,
"administrative_contact": some_admin_contact,
etc...
}
```

src/registrar/tests/test_views.py
```diff
@@ -1406,6 +1406,40 @@ def test_domain_your_contact_information_content(self):
         )
         self.assertContains(page, "Testy")
 
+    def test_domain_security_email_existing_security_contact(self):
+        """Can load domain's security email page."""
+        self.mockSendPatch = patch("registrar.models.domain.registry.send")
```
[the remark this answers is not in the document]
zandercymatics: Test interference reasons as the parent doesn't import MockEppLib (I believe). I think I did this initially and it would have required a refactor of multiple tests so this minimizes code interference

src/registrar/models/domain.py
zandercymatics: From testing and test cases (we check on _state), I believe the list evals to store a list of references to the specified PublicContact objects, not as copies of them - that is the reasoning for this implementation. So its like in c++ how you can have a pointer and get your data from that rather than making a copy of it. When doing it this way, objects are passed by reference in Python, which is why you can do stuff like x = PublicContact.get(etc...), x.save() and have x.save() save the DB object rather than creating a duplicate DB entry

src/registrar/tests/common.py
```diff
@@ -592,9 +678,27 @@ def mockSend(self, _request, cleaned):
...
+            match getattr(_request, "id", None):
+                case "securityContact":
+                    mocked_result = self.mockSecurityContact
+                case "technicalContact":
+                    mocked_result = self.mockTechnicalContact
+                case "adminContact":
+                    mocked_result = self.mockAdministrativeContact
+                case "regContact":
+                    mocked_result = self.mockRegistrantContact
+                case "123":
```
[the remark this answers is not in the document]
zandercymatics: Ah good catch. Nice, will add that!

src/registrar/tests/common.py
```diff
@@ -554,34 +555,119 @@ def __init__(
...
+                tr_date=...,
+                disclose=...,
+                vat=...,
+                ident=...,
+                notify_email=...,
+            )
+            return fake
 
     mockDataInfoDomain = fakedEppObject(
+        "lastPw",
```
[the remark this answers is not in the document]
zandercymatics: Not sure why its showing the entire thing as changed, but the change to contacts had to occur as it was not in parity with the data being returned from EppLib. EppLib doesn't return contacts as strings, but rather as DomainContact objects (with the exception of registrant, I think).

src/registrar/tests/test_models_domain.py
zandercymatics: Yeah no problem! I'll leave this "in_db" section and just add additional tests at the end, so we're expanding our test scope to test both the DB, the list, and the mapping behaviour

src/registrar/tests/test_models_domain.py
```diff
@@ -197,7 +247,10 @@ def test_duplicate_creation(self):
     def tearDown(self) -> None:
         DomainInformation.objects.all().delete()
         DomainApplication.objects.all().delete()
+        PublicContact.objects.all().delete()
         Domain.objects.all().delete()
+        User.objects.all().delete()
+        DraftDomain.objects.all().delete()
```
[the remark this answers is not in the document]
zandercymatics: test_approved_application_creates_domain_locally creates a DraftDomain object that doesn't get deleted - would cause issues if these cases expand and need to test on DraftDomain

src/registrar/models/domain.py
```diff
@@ -1029,13 +1174,71 @@ def _fetch_cache(self, fetch_hosts=False, fetch_contacts=False):
...
+
+        if db_contact.count() == 1:
+            existing_contact = db_contact.get()
+            # Does the item we're grabbing match
+            # what we have in our DB?
+            # If not, we likely have a duplicate.
+            if (
+                existing_contact.email != public_contact.email
+                or existing_contact.registry_id != public_contact.registry_id
+            ):
```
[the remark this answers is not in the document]
zandercymatics: Not opposed to that. The only case in which this could trigger is if EppLib makes an update on their end and its not updated for us, or if somehow a user goes in and makes changes through our DB (not with our setter).

abroddrick (approved):
Before committing see other comments.

Also before committing there is this test test_domain_security_email_form that is marked as skip and complete in #912, please fill it and get it passing before merging or notify so another ticket can be made to complete it.

Also setting a security email doesn't seem to update the Security Email field in the Domain Overview
image
This seems to be caused by domain_detail.html still doing domain.security_email instead of .get_security_email(). However there would have to be some logic to prevent the default contact from appearing on this page as well. I would have imagined this being included in the ticket, however, let me know if you want this to just be a follow on ticket as it might get a bit more complicated to fix.

zandercymatics:
@abroddrick Yeah I can get "domain_detail.html" fixed as well - thanks! Looks like thats one I missed - no idea how, I think its the code blindness / laser focus you get sometimes. It doesn't look too complicated change-wise.

zandercymatics:
Ignore the branching weirdness for a moment

zandercymatics:
@gabydisarli or @Katherine-Osos Would you be able to review the add a security contact / save button under security contact?

gabydisarli (approved):
LGTM!

PR #1092, opened 2023-09-29
Issue 848: Update/Create/Delete Nameservers

### LINKED ISSUE #848: Create, delete, and update nameserver settings at registry
### Issue Description

Implement, retrieve and set Name Servers within the domain.py file. When items are deleted, ensure delete is performed before the add command.
Implement tests as per #493 
NOTE: there are stubbed _update_or_create_host() and stubbed _delete_hosts functions that can be used. Delete if you choose not to use them

### Acceptance Criteria

1. Setting nameservers should change the domain to READY if there are between 2 and 13 (inclusive) nameservers. 
* Just need to add tests for this (add test for less than 2, more than 14 nameservers, and one that in between 2 and 13)

2. DELETE: Implement removing nameservers + _change the state_ (READY -> DNS NEEDED) if user has less then 2 nameservers
* Fix tests for these
* (Compare to list? And send delete command?)

3. UPDATE: If a nameserver has an IP address -- we need to update nameserver according (others _can_ has an IP address, dotgov _has_ to have a IP address)
* Can use an update command  to change the IP address
* If already has IP address and updates the IP address --> check if already have nameserver, and then go down update route

4.  Only allow changes if we are in READY or DNS NEEDED state
* Q: If state of nameserver is in Clienthold, should we even allow the user to edit the nameservers? (Assuming no)

This is partially implemented but not tested
_No response_

### Issue Links

Relates to: #493 
Related to #562

therealslimhsiehdy:
## Ticket

Resolves #848 

## Changes
Implement update, create and deleting nameservers within the domain.py file.
-can update existing nameservers
-delete a nameserver by just deleting a nameservers string and clicking save
-create a new nameserver (already worked)
-changes not allowed if  not ready or in dnsneeded

## Setup

Go to the nameserver page on a domain and add, delete and create new domains
-note error message displaying is handled in the next ticket
-the issue on the domain overview page where it displays a list is also not included in this ticket

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


src/registrar/tests/test_models_domain.py
```diff
@@ -523,7 +526,7 @@ def test_update_is_unsuccessful(self):
         raise
 
 
-class TestRegistrantNameservers(TestCase):
+class TestRegistrantNameservers(MockEppLib):
```
abroddrick, line 529: @neilmb Mind looking over what we are testing here and let us know if there are other cases we should unit test

src/registrar/models/domain.py
```diff
@@ -279,55 +282,173 @@ def _create_host(self, host, addrs):
...
+        # TODO-848: If duplicated nameserver names, throw error
+        for tup in listToConvert:
+            if len(tup) == 1:
+                newDict[tup[0]] = None
+            else:
+                newDict[tup[0]] = tup[1]
+        return newDict
+
+    def isSubdomain(self, nameserver):
+        return nameserver.find(self.name) != -1
```
neilmb, line 297: **comment (non-blocking)**: I think a more pythonic version of this is `self.name not in nameserver`.

src/registrar/models/domain.py
```diff
@@ -279,55 +282,173 @@ def _create_host(self, host, addrs):
...
+
+    def checkHostIPCombo(self, nameserver: str, ip: list):
+        if self.isSubdomain(nameserver) and (ip is None or ip == []):
+            raise ValueError("Nameserver %s needs to have an ip address" % nameserver)
```
neilmb, line 301: **comment (non-blocking)**: Could this error message explain that the nameserver needs an IP address because it is a subdomain?

src/registrar/models/domain.py
```diff
@@ -279,55 +282,173 @@ def _create_host(self, host, addrs):
...
+                "Nameserver %s cannot be linked "
+                "because %s is not a subdomain" % (nameserver, ip)
+            )
+        return None
+
+    def getNameserverChanges(self, hosts: list[tuple[str]]):
+        """
+        calls self.nameserver, it should pull from cache but may result
+        in an epp call
+        returns tuple of four values as follows:
```
neilmb, line 313: **comment**: I think it should be possible to type hint the return type of this as `-> tuple[list, list, dict, list]` as an additional help alongside the comment.

src/registrar/models/domain.py
```diff
@@ -279,55 +282,173 @@ def _create_host(self, host, addrs):
...
+        newHostDict = self._convert_list_to_dict(hosts)
+        deleted_values = []
+        updated_values = []
+        new_values = []
+
+        for prevHost in previousHostDict:
+            addrs = previousHostDict[prevHost]
+            # get deleted values-which are values in previous nameserver list
+            # but are not in the list of new host values
+            if prevHost not in newHostDict:
```
neilmb, line 331: **comment**: This might be easier in Python's set type as something like `deleted_values = set(previousHostDict).difference(newHostDict)`

src/registrar/models/domain.py
```diff
@@ -279,55 +282,173 @@ def _create_host(self, host, addrs):
...
+                self.dns_needed()
+                self.save()
+            except Exception as err:
+                logger.info(
+                    "nameserver setter checked for dns_needed state "
+                    "and it did not succeed. Warning: %s" % err
+                )
+        elif successTotalNameservers >= 2 and successTotalNameservers <= 13:
+            try:
+                print(
```
neilmb, line 440: **comment** We don't usually keep bare `print` statements around. Either use `logger.info` if it's worth keeping or delete it before the PR.

src/registrar/tests/common.py
```diff
@@ -577,22 +581,85 @@ def __init__(
...
     mockDataInfoHosts = fakedEppObject(
+        "lastPw",
+        cr_date=datetime.datetime(2023, 8, 25, 19, 45, 35),
+        addrs=["1.2.3", "2.3.4"],
+    )
+
+    mockDataHostChange = fakedEppObject(
         "lastPw", cr_date=datetime.datetime(2023, 8, 25, 19, 45, 35)
     )
+    infoDomainHasIP = fakedEppObject(
```
neilmb, line 625: Is this the right name for this object? I don't see an IP here at all.

src/registrar/tests/test_models_domain.py
```diff
@@ -532,9 +535,44 @@ def setUp(self):
...
+        super().setUp()
+        self.nameserver1 = "ns1.my-nameserver-1.com"
+        self.nameserver2 = "ns1.my-nameserver-2.com"
+        self.nameserver3 = "ns1.cats-are-superior3.com"
+
+        self.domain, _ = Domain.objects.get_or_create(
+            name="my-nameserver.gov", state=Domain.State.DNS_NEEDED
+        )
+
+    def test_get_nameserver_changes(self):
```
neilmb, line 547: **comment**: There's lots of different types of changes loaded into this one test, changing IPs, taking away IPs, etc. I'd love to see them in separate tests for each kind of change. I know it's kind of a drag, but it's a nice way to make sure that we're testing all the mutations one by one, and then maybe a big test like this that does a lot of things to make sure they work together.

src/registrar/tests/test_models_domain.py
```diff
@@ -544,9 +582,28 @@ def test_user_adds_one_nameserver(self):
...
             And `domain.is_active` returns False
         """
-        raise
 
-    @skip("not implemented yet")
+        # set 1 nameserver
+        nameserver = "ns1.my-nameserver.com"
+        self.domain.nameservers = [(nameserver,)]
+
+        # when you create a host, you also have to update at same time
```
neilmb, line 590: **comment**: This comment confused me because I didn't realize that these commands were being set up to use in testing with `check_call`. Maybe "when we create a hose, we should have updated at the same time"?

src/registrar/tests/test_models_domain.py
```diff
@@ -544,9 +582,28 @@ def test_user_adds_one_nameserver(self):
...
+
+        # checking if commands were sent (commands have to be sent in order)
+        expectedCalls = [
+            call(created_host, cleaned=True),
+            call(update_domain_with_created, cleaned=True),
+        ]
+
+        self.mockedSendFunction.assert_has_calls(expectedCalls)
+
+        # check that status is still NOT READY
```
neilmb, line 604: **suggestion**: Can you say more here about why this won't be ready yet?

src/registrar/tests/test_models_domain.py
```diff
@@ -578,21 +697,91 @@ def test_user_removes_some_nameservers(self):
                 to the registry
             And `domain.is_active` returns True
         """
-        raise
 
-    @skip("not implemented yet")
+        # Mock is set to return 3 nameservers on infodomain
+        self.extendedValues = True
```
neilmb, line 702: **suggestion**: This feels a little magical because of the way it interacts with _getattrshelper. Can we give it a more specific name or use a separate mock so we don't have to change it with this instance attribute?

src/registrar/tests/test_models_domain.py
```diff
@@ -637,9 +929,11 @@ def test_user_cannot_add_non_subordinate_with_ip(self):
                 which is not a subdomain of the domain and has IP addresses
             Then Domain raises a user-friendly error
         """
-        raise
+        dotgovnameserver = "mynameserverdotgov.gov"
+
+        with self.assertRaises(ValueError):
```
neilmb, line 934: Seeing these `ValueError` asserts makes me wonder if we want to use subclasses of `ValueError` for the different particular errors that could be raised when we try this.

src/registrar/tests/test_models_domain.py
neilmb: @abroddrick @dave-kennedy-ecs I like that this fills in all of the skipped tests, that seems like a great start. The last thing I think that I want to see is a round-trip or end-to-end test of the nameserver management view in the Django domain management application. I think that Dave's work should roll up to there so that we should be able to load `/domain/<id>/nameservers`, enter some change in the form, press `save`, and then check that the EPP calls were made and that the page now shows the new nameservers.

If that isn't all hooked up yet, then maybe make a skipped test in `src/registrar/tests/test_views.py` that describes doing that and that test can be filled in when the connections are made?

src/registrar/models/domain.py
therealslimhsiehdy: Oop 30 minutes and 4 print statements later I just realized you meant `self.name in nameserver` not `not in` 😆 😭

src/registrar/models/domain.py
therealslimhsiehdy: Unfortunately we need the values in this situation and set only takes and sorts on the keys 😅  Also major TIL that you could use `.difference` between we were just doing `set(old_dict)- set(new_dict) LOL

src/registrar/models/domain.py
therealslimhsiehdy: Apologies -- this is still a draft PR so thank you for bearing with me on the random print statements for local testing reasons and random helper function names! Will totally remove when the PR is more finalized (currently just a self reference to make sure I'm getting the right count) 😅

src/registrar/tests/common.py
therealslimhsiehdy: Great catch!

Katherine-Osos (approved):

src/registrar/utility/errors.py
```diff
@@ -8,3 +11,65 @@ class ExtraDotsError(ValueError):
...
+    """
+
+    _error_mapping = {
+        NameserverErrorCodes.MISSING_IP: "Nameserver {} needs to have an "
+        "IP address because it is a subdomain",
+        NameserverErrorCodes.GLUE_RECORD_NOT_ALLOWED: "Nameserver {} cannot be linked "
+        "because it is not a subdomain",
+        NameserverErrorCodes.INVALID_IP: "Nameserver {} has an invalid IP address: {}",
+        NameserverErrorCodes.TOO_MANY_HOSTS: (
+            "Too many hosts provided, you may not have more than " "13 nameservers."
```
CocoByte, line 54: Was this intended to be on separate lines?
(that said, if the linter isn't upset, I'm not either)

CocoByte (approved):
Admittedly new to python testing, but after reviewing Neil's comments and your code, I do not see any glaring issues.  Looks good to me!

src/registrar/models/domain.py
```diff
@@ -304,54 +470,62 @@ def dnssecdata(self, _dnssecdata: extensions.DNSSECExtension):
...
+                )
+        elif successTotalNameservers >= 2 and successTotalNameservers <= 13:
+            try:
+                self.ready()
+                self.save()
+            except Exception as err:
+                logger.info(
+                    "nameserver setter checked for create state "
+                    "and it did not succeed. Warning: %s" % err
+                )
```
neilmb, line 528: **Suggestion (non-blocking)**: I expected one final `else` or `elif` block for the too-many nameservers case. I think that's a separate ticket, but maybe it would be nice to have a "TODO" comment here with the ticket number or URL.

abroddrick: We do have a check at the top of this function for:
 if len(hosts) > 13:
            raise NameserverError(code=nsErrorCodes.TOO_MANY_HOSTS)

Thus I think it is impossible for the hosts to be greater than 13. Also if you add more than 13 it will throw an error above this line because of the registry error that results on update domain. With this do you still think that we need to check it again at the end to make sure more than 13 haven't appeared?

src/registrar/models/domain.py
```diff
@@ -282,6 +273,184 @@ def _create_host(self, host, addrs):
...
+        newDict: dict[str, Any] = {}
+
+        for tup in listToConvert:
+            if len(tup) == 1:
+                newDict[tup[0]] = None
+            elif len(tup) == 2:
+                newDict[tup[0]] = tup[1]
+        return newDict
+
+    def isSubdomain(self, nameserver: str):
```
dave-kennedy-ecs, line 293: is this logic sufficient for determining a subdomain?  or is a regular expression needed?  for example, isSubdomain('ns1-domain.com') would return True for domain with name 'domain.com'.  Is that ok?

therealslimhsiehdy: Great catch! Added regex in 🚀

PR #1095, opened 2023-09-29
Issue 1027: Refactor users, groups and permissions, logging on groups and permissions

### LINKED ISSUE #1027: Implement logging to capture changes made to Groups and Permissions, Refactor Groups and Permissions
### Issue Description

User roles/permissions should be handled in django admin groups and changes made to django-admin  groups must be logged. 

The log should contain an identifier of who made the change and what was changed. The log doesn't need to necessarily be in human readable format. Example: Timestamp: UUID 555555 has edited role 983 with attribute x y z.

### Acceptance Criteria

- [x] Log entry is made when a Group is created or deleted
- [x] Log entry is made when a permission is changed on a Group or a User
- [x] implement Groups with full permissions and analyst permissions in models and migrations
- [x] change fixtures to assign full permissions Group to perviously superusers and analyst permission group to previously staff (all fixtures users will be staff)
- [x] Refactor instances in admin.py where we are testing for is_staff or is_superuser to set permissions
- [ ] Refactor unit tests
- [x] Separate users and DAs/Domains fixtures
- [x] divide fixtures into 2 scripts, with one specific to users

https://cisa-corp.slack.com/archives/C03QM0JGSQG/p1694718697209909

rachidatecs:
## Ticket

Resolves #1027 

## Changes

- Separate fixtures in 2: users and data, and remove permissions assignment from fixtures
- New custom User Group that defines the group creations in the DB (`full_access_group` and `cisa_analyst_group`)
- Migration that populates ContentType and Permission (prerequisite for creation the 2 groups)
- Migration that creates the 2 groups
- Audit log permission changes on users and the custom user groups
- Refactor testing for permissions in admin.py (has_perm instead or is_superuser or has_group)
- Refactor testing for permissions in the Domain Permission classes `DomainPermissionView`, `DomainPermission`
- Unit tests for the data creating migration
- Refactor existing unit tests and mocks
- Update docs
- Refactor django admin user views to show groups instead of is_staff and is_superuser in list and detail views

## Context for reviewers

`test_place_and_remove_hold` now fails and logging indicates that `DomainAdmin.do_remove_client_hold` and `DomainAdmin.do_place_client_hold` do not even get called after the POST request (so I ruled out EPP connection or Mock issue). I also tried to rule out `DomainAdmin.has_change_permission` by hardcoding a True return, same with the permission checkers within the view permission classes. Thanks @dave-kennedy-ecs for helping me debug.

### Designers:

This PR changes how we're doing permissions at a structural level. Proofing of what used to be 'superuser' vs 'analyst' views and permissions are probably a good idea. Logging in still works the same way, everything else should still work the same way except that we now have `groups` and `permissions` instead of `superuser` and `staff`.

Screen Shot 2023-09-29 at 6 57 42 PM

## Setup

For our MVP, we create and maintain 2 admin roles:
Full access and CISA analyst. Both have the role `staff`.
Permissions on these roles are set through groups:
`full_access_group` and `cisa_analysts_group`. These
groups and the methods to create them are defined in
our `user_group` model and run in a migration.

We can edit and deploy new group permissions by:

1. editing `user_group` then:
2. Duplicating migration `0036_create_groups`
and running migrations

- Compose-up and test that permissions for what were perviously superuser and staff are still the same in django admin through groups.
- Confirm that group and permission creation got logged, make changes and confirm logs again.
- Confirm that he edit domain custom actions in django admin domain detail view still work as expected for both roles.
- Make a change to cisa_analyst_permissions in code and try triggering the migration as described above, confirm changes take effect in the application.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


docs/developer/user-permissions.md
```diff
@@ -48,3 +48,7 @@ future, as we add additional roles that our product vision calls for
 (read-only? editing only some information?), we need to add conditional
 behavior in the permission mixin, or additional mixins that more clearly
 express what is allowed for those new roles.
+
+# Admin User Permissions
+
+Refre to [Django Admin Roles](../django-admin/roles.md)
```
zandercymatics, line 54:
```suggestion
Refer to [Django Admin Roles](../django-admin/roles.md)
```

docs/django-admin/roles.md
```diff
@@ -1,21 +1,16 @@
 # Django admin user roles
 
-Roles other than superuser should be defined in authentication and authorization groups in django admin 
+For our MVP, we create and maintain 2 admin roles:
+Full access and CISA analyst. Both have the role `staff`.
+Permissions on these roles are set through groups:
+`full_access_group` and `cisa_analysts_group`. These
```
zandercymatics, line 6: A bit like how it was done previously in the now deleted "Superuser" section, I think we should roughly define where cisa_analysts_group is restricted.  I don't think it needs to be super detailed but having a reference here would be useful if we need to make a change. Could even be as simple as redirecting them to look at 
user_group for more context

docs/django-admin/roles.md
```diff
@@ -1,21 +1,16 @@
...
-
-Staff
-
-### Additional group permissions
-
-auditlog | log entry | can view log entry
-registrar | contact | can view contact
-registrar | domain application | can change domain application
-registrar | domain | can view domain
-registrar | user | can view user
```
zandercymatics, line 21: Something similarish to this if it makes sense to do. Again, I wouldn't say it needs to be all encompassing

docs/django-admin/roles.md
```diff
@@ -1,21 +1,16 @@
...
\ No newline at end of file
+1. editing `user_group` then:
+2. Duplicating migration `0036_create_groups`
```
zandercymatics, line 15: Silly question: what do you mean by duplicating migrations here? Like copy+pasting, running makemigrations / migrate multiple times, etc?

docs/operations/README.md
```diff
@@ -89,7 +89,8 @@ command in the running Cloud.gov container. For example, to run our Django
 admin command that loads test fixture data:
 
 ```
-cf run-task getgov-{environment} --command "./manage.py load" --name fixtures
+cf run-task getgov-{environment} --command "./manage.py load" --name fixtures--users
+cf run-task getgov-{environment} --command "./manage.py load" --name fixtures--applications
```
zandercymatics, line 93: I like the distinction here!

src/registrar/admin.py
```diff
@@ -187,43 +200,48 @@ class MyUserAdmin(BaseUserAdmin):
         "email",
         "Permissions",
         "is_active",
-        "is_staff",
-        "is_superuser",
+        "groups",
```
zandercymatics, line 203: Question: since I think we still have a is_staff variable, is that editable or is it not because of hierarchy (groups contains is_staff which means its read_only)?

src/registrar/admin.py
```diff
@@ -187,43 +200,48 @@ class MyUserAdmin(BaseUserAdmin):
...
+            return super().get_list_display(request)
+
+        # Customize the list display for analysts
+        return (
+            "email",
+            "first_name",
+            "last_name",
+            "group",
+            "status",
+        )
```
zandercymatics, line 225: This seems like something we could store in an external variable somewhere, i.e. "analyst_list_display"

src/registrar/admin.py
```diff
@@ -187,43 +200,48 @@ class MyUserAdmin(BaseUserAdmin):
...
     def get_readonly_fields(self, request, obj=None):
-        if request.user.is_superuser:
-            return ()  # No read-only fields for superusers
-        elif request.user.is_staff:
-            return self.analyst_readonly_fields  # Read-only fields for staff
-        return ()  # No read-only fields for other users
+        if request.user.has_perm("registrar.full_access_permission"):
+            return ()  # No read-only fields for all access users
+        # Return restrictive Read-only fields for analysts and
+        # users who might not belong to groups
```
zandercymatics, line 243: Weird that it wasn't this way before. Nice refactor!

src/registrar/admin.py
```diff
@@ -187,43 +200,48 @@ class MyUserAdmin(BaseUserAdmin):
...
-                "status",
-            )
-
-        # Use the default list display for non-staff users
-        return super().get_list_display(request)
+        # The full_access_permission perm will load onto the full_access_group
+        # which is equivalent to superuser. The other group we use to manage
+        # perms is cisa_analysts_group. cisa_analysts_group will never contain
+        # full_access_permission
+        if request.user.has_perm("registrar.full_access_permission"):
```
zandercymatics, line 214: It seems a little weird that Django chose to manage these as strings. It might be a bit of a refactor but we should really be storing these in a constant or a literal

src/registrar/tests/test_admin.py
```diff
@@ -51,6 +51,7 @@ def setUp(self):
         self.staffuser = create_user()
         super().setUp()
 
+    @skip("Why did this test stop working, and is is a good test")
```
zandercymatics, line 54: Thinking on this now, could it actually potentially be related to `self.client.login(username="staffuser", password=p)`? For instance, if `display_restricted_warning()` gets triggered it may not return a bad status code but it would stop changes from occurring. That said I do not know under what circumstances a user is put into "restricted"

src/registrar/tests/test_migrations.py
```diff
@@ -0,0 +1,51 @@
...
+        self.assertQuerysetEqual(
+            UserGroup.objects.filter(name="cisa_analysts_group"), [cisa_analysts_group]
+        )
+
+        # Assert that the full_access_group exists in the database
+        self.assertQuerysetEqual(
+            UserGroup.objects.filter(name="full_access_group"), [full_access_group]
+        )
+
+        # Test permissions for cisa_analysts)group
```
therealslimhsiehdy, line 34: **non-blocking super small nit:** remove the `)` in between cisa_analysts and group

therealslimhsiehdy (commented):
Did a first galance walkthrough of the code and this looks really good (great work btw, this is HUGE!) -- once I test locally I will approve 💯

src/registrar/tests/test_admin.py
rachidatecs: Shouldn't be. The user we're creating in common is not restricted - and both constructs (that user, the restricted check) existed before the PR and this test was running fine.

src/registrar/admin.py
rachidatecs: It is editable but I no longer think an analyst needs or should have eyes on, since permissions and roles are managed through groups.

src/registrar/admin.py
rachidatecs: Ideally, yes. Not worth it at this point imo.

docs/django-admin/roles.md
rachidatecs: Saving this migration file as a new migration and editing its dependency (probably best to have it run at the end of the Q)

src/registrar/admin.py
zandercymatics: Fair enough

src/registrar/admin.py
```diff
@@ -137,11 +138,19 @@ class MyUserAdmin(BaseUserAdmin):
...
-        "is_superuser",
+        "group",
         "status",
     )
 
+    # Let's define First group
+    # (which should in theory be the ONLY group)
+    def group(self, obj):
+        if obj.groups.filter(name="full_access_group").exists():
+            return "Super User"
```
zandercymatics, line 149: This seems to be inconsistent with how it is mentioned on the rest of the page. "Superuser" instead of "Super User"
image

If this is a design thing, I'd say that comes first but personally I think Superuser makes more sense in this context for consistency purposes

src/registrar/tests/test_migrations.py
```diff
@@ -0,0 +1,51 @@
...
+        """
+
+        # Get the UserGroup objects
+        cisa_analysts_group = UserGroup.objects.get(name="cisa_analysts_group")
+        full_access_group = UserGroup.objects.get(name="full_access_group")
+
+        # Assert that the cisa_analysts_group exists in the database
+        self.assertQuerysetEqual(
+            UserGroup.objects.filter(name="cisa_analysts_group"), [cisa_analysts_group]
+        )
```
zandercymatics, line 27: I would assume that this is implied by the .get call?  As it'd throw an err otherwise? A better check here would be to hard code some mimicked data and check  against that (for instance, a hardcoded UserGroup() object not present on the DB), as otherwise we're just doing a check on Djangos built in querying system which we can assume is reliable

src/registrar/tests/test_migrations.py
```diff
@@ -0,0 +1,51 @@
...
+            "change_draftdomain",
+            "analyst_access_permission",
+            "change_user",
+        ]
+
+        # Get the codenames of actual permissions associated with the group
+        actual_permissions = [p.codename for p in cisa_analysts_group.permissions.all()]
+
+        # Assert that the actual permissions match the expected permissions
+        self.assertListEqual(actual_permissions, expected_permissions)
```
zandercymatics, line 51: Nice!

zandercymatics (approved):
Works beautifully! Great work, this was a huge undertaking. I did a second pass of the code and noticed an issue with your test case - I think that ought to be resolved before merging, but I won't block you on that

src/registrar/admin.py
rachidatecs: How about 'God mode'?

src/registrar/tests/test_migrations.py
```diff
@@ -0,0 +1,51 @@
...
+
+        # Assert that the cisa_analysts_group exists in the database
+        self.assertQuerysetEqual(
+            UserGroup.objects.filter(name="cisa_analysts_group"), [cisa_analysts_group]
+        )
+
+        # Assert that the full_access_group exists in the database
+        self.assertQuerysetEqual(
+            UserGroup.objects.filter(name="full_access_group"), [full_access_group]
+        )
```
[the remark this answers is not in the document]
rachidatecs: Same with this.

src/registrar/tests/test_migrations.py
zandercymatics: This is OK - Rachid had mentioned that he is doing a test on migrations directly

src/registrar/tests/test_migrations.py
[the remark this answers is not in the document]
zandercymatics: This is OK - Rachid had mentioned that he is doing a test on migrations directly

src/registrar/admin.py
rachidatecs: You know, we should remove that superuser filter since it's now obsolete and can be misleading

Issue #923, opened 2023-08-23
Data Migration: Parse txt data into the transition_domain table


kitsushadow:
### Issue Description

- Parse and insert using the latest version of three files ([available in our Drive](https://drive.google.com/drive/folders/1TWxHUZexMqTK6VH9iXgopiX7OknX4vOC?usp=drive_link))
  - escrow_domain_contacts.daily.dotgov.GOV.txt
  - escrow_contacts.daily.dotgov.GOV.txt
  - escrow_domain_statuses.daily.dotgov.GOV.txt
- We need a script(s) to parse and ingest the above
- The script(s) should handle running multiple executions of the same or similar files
- This script(s) should be able to be run multiple times and it should check if a user/domain mapping is a duplicate and not create a new record if it IS a duplicate
  - It should check to see if the domain status is the same, if not update with the newest files domain status
- The status field will populate from the escrow_domain_statuses.daily.dotgov.GOV.txt. During parsing, "serverHold” fields will map to hold clientHold to hold and any ok state should map to Ready 
- When testing, double check that no other translation of states is needed. Do this by checking if there are any statuses that are not serverhold, client hold or OK in the original data set. Let Alysia know if some outliers are found and we can figure out what to do. 
- Reference scipts: 
  - getgov/src/registrar/management/commands/load_domains_data.py
  - getgov/src/registrar/management/commands/load_domain_invitations.py

### Additional Context (optional)

https://docs.google.com/document/d/1UkQZdAmcGprXWdHX_PuT155CCNPbf_Gf/edit?pli=1#heading=h.iqv2v8lecjpn

### Issue Links

Depends on: #921

PR #1115, opened 2023-10-03
Issue 926: transition domain object creation

### LINKED ISSUE #926: Data Migration: transition_domain domain object ingest and create/update
### Issue Description

Create a script that takes all the data from the transition_domain table from #923 and makes the corresponding objects in our main tables.
The script should:
- Creates a domain object that does NOT have a corresponding application for each domain listed in the table
  - it should not create duplicates
  - if a domain already exists, it should check the domain.state and if its not equal to transition_domain.status, set it to be the new value
- Creates a domainInvitation object for each of the emails in the username field and associates it with the corresponding domain object
  - emails should **NOT** be sent upon creation of the domain invitation objects
- This must be able to run in combination with the docker-compose run app ./manage.py "script_name"

https://docs.google.com/document/d/1UkQZdAmcGprXWdHX_PuT155CCNPbf_Gf/edit?pli=1#heading=h.iqv2v8lecjpn

### Issue Links

Depends on: #923

CocoByte:
Resolves 
#926 

## Changes

- Created script transfer_transition_domains_to_domains.py
- Updated model domain.py (to create workaround for executing updates to States without executing EPP calls)
- Updated model transition_domain.py (to fix a bug where "Hold" should be "On Hold")
- Created migration for model updates

## Setup

    
   - [ ] Run `docker-compose up` and navigate to /admin
   - [ ] Create some transition domains (see example below)
Select transition domain to change

### For issue #923 
  Open a terminal and run the following commands:
   
`cd src` (navigate to the src folder)

`docker compose run -T app ./manage.py transfer_transition_domains_to_domains --debug`

**NOTE**: 
`--debug` will print out additional, detailed logs

#### 1. Verify that the script...
- [ ] Adds new Domain entries (Check the Domain table on the admin page)
- [ ] Adds new Domain Invitation entries (check the Domain Invitation table on the admin page)
Pasted Graphic 1

#### 2. Re-execute the docker command and verify that the script...
- [ ] Does not add or update any data
Pasted Graphic 2

#### 3. Change some values in the test data (via the admin page) and verify that the script...
- [ ] Updates the corresponding entries

_In the example below, changed "bestCampsites" website from **OnHold** to **Ready**.
Re-ran the script an the output shows 1 updated domain_

Select transition domain to change

Pasted Graphic 5

- [ ] Meets all designs and user flow

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


src/registrar/models/transition_domain.py
```diff
@@ -5,14 +5,18 @@
 
 class StatusChoices(models.TextChoices):
     READY = "ready", "Ready"
```
abroddrick, line 7: Just realized this will make your other PR fail, this will definitely have to be kept in mind

src/registrar/management/commands/transfer_transition_domains_to_domains.py
```diff
@@ -0,0 +1,229 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument("--debug", action=argparse.BooleanOptionalAction)
+
+        parser.add_argument(
+            "--limitParse", default=0, help="Sets max number of entries to load"
+        )
+
+    def handle(  # noqa: C901
+        self,
+        **options,
```
abroddrick, line 48: like with the other PR I would ask that this get broken up into other functions to make it a bit more readable.

src/registrar/management/commands/transfer_transition_domains_to_domains.py
```diff
@@ -0,0 +1,229 @@
...
+        Domain.objects.bulk_create(to_create)
+
+        total_new_entries = len(to_create)
+        total_updated_domain_entries = len(updated_domain_entries)
+
+        logger.info(
+            f"""{termColors.OKGREEN}
+
+============= FINISHED ===============
+Created {total_new_entries} transition domain entries,
```
abroddrick, line 178: can we fix this indenting? and the other ones below that do the same thing

src/registrar/management/commands/transfer_transition_domains_to_domains.py
```diff
@@ -0,0 +1,229 @@
+"""Load domain invitations for existing domains and their contacts."""
```
abroddrick, line 1: remove comment

abroddrick (commented):
Didn't run or thoroughly review, just leaving comments for now

CocoByte:
I will merge from Main to fix failed tests.  But ALSO need to add domainInvitation piece, which I missed.  Will post in slack when this is updated.

zandercymatics (commented):
One thing to note is that while testing step 3,  if you change the domain name on an existing domain (for instance, test.gov => test123.gov), it will show up in the created column rather than the updated one. It works for every other field, though I am not sure if this is a bug or intended behavior. If its intended I'll approve
image

src/registrar/models/transition_domain.py
CocoByte: Good catch. In my latest updates I matched the transitionDomain model in both branches (along with the migration) so hopefully there will be no collisions there.

src/registrar/management/commands/transfer_transition_domains_to_domains.py
CocoByte: Indents fixed

src/registrar/management/commands/transfer_transition_domains_to_domains.py
CocoByte: Done

src/registrar/management/commands/transfer_transition_domains_to_domains.py
```diff
@@ -0,0 +1,229 @@
...
+                if debug_on:
+                    logger.info(
+                        f"""{termColors.WARNING}
+    > Found existing domain entry for: {transition_domain_name}, {current_state}
+    {termColors.ENDC}"""
+                    )
+                if transition_domain_status != current_state:
+                    if (
+                        transition_domain_status
+                        == TransitionDomain.StatusChoices.ON_HOLD
```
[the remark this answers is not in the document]
CocoByte: yes

src/registrar/management/commands/transfer_transition_domains_to_domains.py
```diff
@@ -0,0 +1,229 @@
...
+
+        total_new_entries = len(to_create)
+        total_updated_domain_entries = len(updated_domain_entries)
+
+        logger.info(
+            f"""{termColors.OKGREEN}
+
+============= FINISHED ===============
+Created {total_new_entries} transition domain entries,
+updated {total_updated_domain_entries} transition domain entries
```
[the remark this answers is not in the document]
CocoByte: I like it.  Implementing

src/registrar/management/commands/transfer_transition_domains_to_domains.py
CocoByte: Done

src/registrar/management/commands/transfer_transition_domains_to_domains.py
```diff
@@ -0,0 +1,323 @@
...
+                # DEBUG:
+                self.print_debug(
+                    debug_on,
+                    f"""{termColors.OKCYAN}
+                    Processing Transition Domain: {transition_domain_name}, {transition_domain_status}, {transition_domain_email}
+                    {termColors.ENDC}""",  # noqa
+                )
+
+                # get the existing domain
+                target_domain = Domain.objects.get(name=transition_domain_name)
```
abroddrick, line 224: Why are we not using .exists() here?

src/registrar/management/commands/transfer_transition_domains_to_domains.py
```diff
@@ -0,0 +1,323 @@
...
+                # for existing entry, update the status to
+                # the transition domain status
+                update_made = self.update_domain_status(
+                    transition_domain, target_domain, debug_on
+                )
+                if update_made:
+                    updated_domain_entries.append(transition_domain.domain_name)
+
+            except Domain.DoesNotExist:
+                already_in_to_create = next(
```
abroddrick, line 243: This is a bit hard to read I would add a comment here

src/registrar/management/commands/transfer_transition_domains_to_domains.py
```diff
@@ -0,0 +1,323 @@
...
+                        {termColors.ENDC}""",
+                    )
+                else:
+                    # no matching entry, make one
+                    new_entry = Domain(
+                        name=transition_domain_name, state=transition_domain_status
+                    )
+                    domains_to_create.append(new_entry)
+
+                    if transition_domain_email:
```
abroddrick, line 263: x is not a boolean you want to avoid checking Truthy in python as a rule. I would check against None/empty string vs doing the truthy check

src/registrar/management/commands/transfer_transition_domains_to_domains.py
```diff
@@ -0,0 +1,323 @@
...
+                    > Found existing domain entry for: {transition_domain_name}, {target_domain.state}
+                    {termColors.ENDC}""",  # noqa
+                )
+
+                # for existing entry, update the status to
+                # the transition domain status
+                update_made = self.update_domain_status(
+                    transition_domain, target_domain, debug_on
+                )
+                if update_made:
```
abroddrick, line 239: We can have duplicate domains with different users. We should be making a domain invitation for each user, even if the domain object exists. Hence since we have to repeat some of the code that is below, it may be better to do .exists() instead of waiting for an exception.

src/registrar/management/commands/transfer_transition_domains_to_domains.py
```diff
@@ -0,0 +1,323 @@
...
+                        new_domain_invitation = DomainInvitation(
+                            email=transition_domain_email.lower(), domain=new_entry
+                        )
+                        domain_invitations_to_create.append(new_domain_invitation)
+                    else:
+                        logger.info(
+                            f"{termColors.FAIL} ! No e-mail found for domain: {new_entry}"  # noqa
+                            f"(SKIPPED ADDING DOMAIN INVITATION){termColors.ENDC}"
+                        )
+                        skipped_domain_invitations.append(transition_domain_name)
```
abroddrick, line 273: since transition domains can have duplicates, couldnt a domain have one row with an actual email and another row with a missing email. In which case we don't want to add to skipped domain invitations if at least SOMEONE got an invitation for this domain. Would require 1.) when adding a domain invitation if that domain exists in skipped_domain_invitations make  remove it from skipped. 2.) Visa versa, don't add to skipped if it is in domain_invitiations_to_create

src/registrar/management/commands/transfer_transition_domains_to_domains.py
```diff
@@ -0,0 +1,323 @@
...
+                debug_max_entries_to_parse > 0
+                and total_rows_parsed >= debug_max_entries_to_parse
+            ):
+                logger.info(
+                    f"""{termColors.YELLOW}
+                    ----PARSE LIMIT REACHED.  HALTING PARSER.----
+                    {termColors.ENDC}
+                    """
+                )
+                break
```
abroddrick, line 311: dont think this is needed

src/registrar/management/commands/transfer_transition_domains_to_domains.py
```diff
@@ -0,0 +1,323 @@
+"""Load domain invitations for existing domains and their contacts."""
```
abroddrick, line 1: not existing domains, maybe reword "Load domain Invitations and Domain objects from the transition domain table"

abroddrick (changes requested):
Hate to do it, but have to request changes. There is a big where if a domain has multiple users we aren't making domain invitations for all those users. See example below, bob should have 3 invitations.
image

besides that just readme, and only minor suggestion changes that are non blocking.

src/registrar/management/commands/transfer_transition_domains_to_domains.py
CocoByte: That would be the proper way to do this.  I think part of me thought it was less efficient, checking if an object exists and then grabbing it is traversing the table twice...and makes this function more complex...plus we check for other exceptions anyways.  But I agree this isn't proper form so I will change it.

src/registrar/management/commands/transfer_transition_domains_to_domains.py
CocoByte: I'm going to change this to use .exists() in an if-else statement

src/registrar/management/commands/transfer_transition_domains_to_domains.py
CocoByte: good to know.  I'll update

src/registrar/management/commands/transfer_transition_domains_to_domains.py
CocoByte: K - will update the algorithm

src/registrar/management/commands/transfer_transition_domains_to_domains.py
CocoByte: needed to exit loop if parse limit has been reached

src/registrar/management/commands/transfer_transition_domains_to_domains.py
CocoByte: I thought I removed this comment...must have re-emerged in the course of my commits

abroddrick:
> One thing to note is that while testing step 3, if you change the domain name on an existing domain (for instance, test.gov => test123.gov), it will show up in the created column rather than the updated one. 

I think that should happen, because the domain name is really our unique id we get from the files.

abroddrick (approved):
Please at least react or reply to existing comments that have not received a response. But none are blocking and I see now invitations are made for all users, this works for me

PR #1116, opened 2023-10-03
Issue #1102: Change disclose behaviour

### LINKED ISSUE #1102: EPP Contact disclose Update
### Issue Description

After the EPP contact setting was written, it was determined that the disclose on the registry would default to not disclosing everything and that we merely have to set the disclose on things we want to show. Our current setup, is unnecessarily setting items to a disclose of false, when instead we just need to disclose the security contact email address IF it is not the default address

AC:
- Change the disclose function in domain.py so that it only ever sets the disclose on a security contacts email to True when the email is not the default email. Nothing else should ever receive the disclose set to true
- update tests accordingly

_No response_

### Issue Links

_No response_

zandercymatics:
## Ticket

Resolves #1102

## Changes

- Changed how we are handling disclose such that it only applies to DF.EMAIL on security_contact 
- Changed common.py to accommodate this change
- Added test cases

## Context for reviewers

The original `_disclose_fields()` function was not behaving correctly and was disclosing disclosed items.  EPP, by default, discloses all items unless otherwise specified. As such, that old logic was redundant and was excluding DF.email when it shouldn't have been. 
## Setup
This can be verified as so:
**Creating the shell:**
cf ssh getgov-za
/tmp/lifecycle/shell
./manage.py shell
**Then run these inside the Python shell you created:**
#This should create a contact with a default email, which should not be disclosed
domain, _ = Domain.objects.get_or_create(name="somekindofdomain.gov")

#Grab the appropriate contact
sec_id = domain.security_contact.registry_id
request = commands.InfoContact(id=sec_id)
response = registry.send(request, cleaned=True)

#Nothing should be disclosed
print(response.res_data[0].disclose)

#Test disclosure...
default = domain.get_default_security_contact()
default.email = "test123@mail.gov"
domain.security_contact = default

request = commands.InfoContact(id=sec_id)
response = registry.send(request, cleaned=True)

#Only email should be disclosed
print(response.res_data[0].disclose)

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


src/registrar/tests/common.py
```diff
@@ -626,15 +626,11 @@ def _convertPublicContactToEpp(
...
         DF = common.DiscloseField
-        fields = {DF.FAX, DF.VOICE, DF.ADDR}
-
-        if not disclose_email:
-            fields.add(DF.EMAIL)
+        fields = {DF.EMAIL}
 
         di = common.Disclose(
-            flag=False,
+            flag=disclose_email,
```
abroddrick, line 632: I think we should add  two tests for the _disclose_fields in the contacts test class and just check once for a security contact and once for a non security contact. That way if something changes down the road we have a test for this.

On visual inspection it all looks good

zandercymatics: @abroddrick Okay I added some tests. The one that checks for the disclosure is redundant but its good to have that test case out separated into its own thing, in case a change happens to that beefier one

therealslimhsiehdy (approved):
LGTM! I would have Alysia/Rachid do a one last glance in case there's a side case I'm not aware of, but this is great!

src/registrar/tests/test_models_domain.py
```diff
@@ -705,6 +705,123 @@ def test_updates_security_email(self):
...
+            (expected_tech, domain.technical_contact),
+        ]
+
+        # Test for each contact
+        for contact in contacts:
+            expected_contact = contact[0]
+            actual_contact = contact[1]
+            is_security = actual_contact.contact_type == "security"
+
+            expectedCreateCommand = self._convertPublicContactToEpp(
```
abroddrick, line 751: In general, here we are just copying the code here we are trying to test. That violates the basic point of our test. There is no point to this test if we are just testing a function against itself. I would manually write out the call we expect like in the other tests.

abroddrick (changes requested):
I can't get passed the use of _convertPublicContactToEpp to generate the expected call when that has the same code being used in the disclose function. Since we want to double check that code works correctly, checking it against itself seems like no matter what it will be true and isn't a great test. I would manually write out the command call just like is done in all other tests to check call content.

src/registrar/tests/test_models_domain.py
zandercymatics: Looks like there is a typo here on is_security. Should be testing on expected_contact.contact_type.

Anywho, I have no problem writing this out manually but in general, I am a bit confused as to how this is always going to return true. Since this is a helper function in common.py, that code is completely decoupled from the disclose function. It doesn't use _convertPublicContactToEpp from domain.py at all, its using the one in common.py
image

image

abroddrick: So first, yes I meant this test specifically. And it's not that you are exactly using the same function but it comes down to a coding best practices concept of limiting code logic within a test and if logic exists in a test, it itself should be tested. Here you are fundamentally checking the disclose is correct. However, you are using code to generate what the expected disclose should be without testing that code. Your code to generate the expected disclose is also almost _identical_ to the very disclose code you are testing. 
Look how close these codes are, basically the only difference is one is passing a boolean instead of making it. Left is the code we are testing, right is the code you are using to test the left...
image

abroddrick: I just do not view this as falling under the category as a valid test due to the duplication of code and the original intent of this test being to explicitly test the disclose. I already requested changes, but I won't block this again, if you want to merge you may.

zandercymatics: I understand your concern with how similar the checks are, I think its entirely reasonable to keep tests as simple as possible so that we actually perform a good test, rather than a bad one. In that case, we should be testing the other use cases of _convertPublicContactToEpp() to best conform to that practice, given that we do use it elsewhere. test_no_security_email is a great example because it needs to explicitly check if its not getting disclosed. See below... 

image
image

Maybe a good middle ground here would be a test case on the _convertPublicContactToEpp() function itself? What do you think?

PR #1125, opened 2023-10-04
Ticket #1120 + ContactError refinement: update domain overview verbiage

### LINKED ISSUE #1120: Update the verbiage to "None provided" when security email is not set on domain overview page
### Issue Description

Currently, if a security email is not set for a user, the domain overview page shows “None” under security email. 

Update to new verbiage: "None provided"

_No response_

### Issue Links

_No response_

zandercymatics:
## Ticket

Resolves #1120

## Changes

- Fixes and issue where security emails don't get displayed on the overview page
- The default text for a security email is now "None provided" on that page, instead of "None"
- Expanded the ContactError exception class so that we can further expand this in the future

## Context for reviewers
- So that we avoid redundant EPP calls, I am passing the security contact email by adding onto the context. That said, this seems somewhat inconsistent with what other pages are doing. Is there a better way to be doing this?

## Setup
Click the "manage" button for an approved domain. Note the default text on the overview page. Then, add a security email and note that it displays as expected.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


src/registrar/templates/domain_detail.html
```diff
@@ -46,8 +46,11 @@ <h2 class="margin-top-neg-1"> DNS name servers </h2>
     {% include "includes/summary_item.html" with title='Your contact information' value=request.user.contact contact='true' edit_link=url %}
 
     {% url 'domain-security-email' pk=domain.id as url %}
-    {% include "includes/summary_item.html" with title='Security email' value=domain.security_email edit_link=url %}
-
+    {% if security_email is not None and security_email != "dotgov@cisa.dhs.gov"%}
```
abroddrick, line 49: just browsing really quick, everything else on a fast inspection looks great. Here can we have this default email just pull from PublicContact.getDefaultSecurityEmail().email or even do domain.getDefaultSecurityContact().email? That way if the default changes we dont have to remember to change it in multiple places

src/registrar/views/domain.py
```diff
@@ -42,6 +42,15 @@ class DomainView(DomainPermissionView):
 
     template_name = "domain_detail.html"
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        security_email = self.get_object().get_security_email()
+        if security_email is None or security_email == "dotgov@cisa.dhs.gov":
```
abroddrick, line 48: same comment here, pull from domain.getdefaultsecurityEmail or get it from PublicContact instead of hardcoding it

src/registrar/models/utility/contact_error.py
```diff
@@ -1,2 +1,51 @@
+from enum import IntEnum
+
+
+class ContactErrorCodes(IntEnum):
+    """Used in the ContactError class for
+    error mapping.
+
+    Overview of contact error codes:
+        - 2000 CONTACT_TYPE_NONE
```
abroddrick, line 9: random, just curious, why 2000?

zandercymatics: I decided to somewhat follow the pattern from RFC 5730 - though very loosely (not literally). Not opposed at all to changing it, I just needed something to use here
image

abroddrick: The number doesn't matter so it's whatever you prefer. That is based on EPP standards so really happy if it matches or happy if it doesn't.

Katherine-Osos (approved):

rachidatecs:
I am unable to set a security email on a DNS Needed domain:

Screen Shot 2023-10-10 at 1 08 20 PM

zandercymatics:
@rachidatecs I may have broken something with that bug fix for deleted, will fix. Also worth noting that setting locally will not work as we cannot contact EPP

src/registrar/admin.py
```diff
@@ -222,6 +208,20 @@ def group(self, obj):
...
+
+    # Let's define First group
+    # (which should in theory be the ONLY group)
+    def group(self, obj):
+        if obj.groups.filter(name="full_access_group").exists():
+            return "Full access"
+        elif obj.groups.filter(name="cisa_analysts_group").exists():
+            return "Analyst"
+        return ""
+
```
zandercymatics, line 224: This was a merge conflict. I just moved it down so it was grouped with the other functions, but I can move it back if that is how it was in the original code

src/registrar/models/domain.py
```diff
@@ -1294,6 +1293,47 @@ def _fetch_cache(self, fetch_hosts=False, fetch_contacts=False):
...
+                    "name": getattr(data, "name", self.name),
+                    "registrant":  getattr(data, "name", registrant_id),
+                    "statuses": getattr(data, "statuses", ...),
+                    "tr_date": getattr(data, "tr_date", ...),
+                    "up_date": getattr(data, "up_date", ...),
+                }
+                cleaned = {k: v for k, v in cache.items() if v is not ...}
+
+                self._cache = cleaned
+
```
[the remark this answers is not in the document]
zandercymatics: Should we be raising an err in this scenario?

rachidatecs:
> @rachidatecs Worth noting that setting locally will not work as we cannot contact EPP but the 500 err should not be there

Doh.

zandercymatics:
@rachidatecs Okay - I fixed that err locally, I've decided to display an error to the user on the page that we cannot connect to the registry. Will require a designer to look at it and reapprove, but I think its a good solution

src/registrar/admin.py
rachidatecs: Ideally, keep this down and add a small comment around line 141 indicating that `group` in `list_display` is custom defined?

zandercymatics: Ah so like separating it back into its own location. Yeah, makes sense to me

src/registrar/models/domain.py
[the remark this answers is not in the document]
rachidatecs: NItpick, non-blocking: instead or raising err, add a "Unknown contact error" definition in contact_error.py and raise that?

src/registrar/admin.py
rachidatecs: It can stay where it is, but we should note that group is a custom column/prop defined in admin.py as opposed to the model.

src/registrar/views/domain.py
```diff
@@ -284,10 +298,21 @@ def form_valid(self, form):
...
+
+        # If we pass nothing for the sec email, set to the default
+        if new_email is None or new_email.strip() == "":
+            new_email = PublicContact.get_default_security().email
 
         domain = self.get_object()
         contact = domain.security_contact
+
+        # If no default is created for security_contact,
+        # then we cannot connect to EPPLib.
```
abroddrick, line 311:
```suggestion
        # then we cannot connect to the registry.
```

abroddrick (commented):
I want to approve because the security contact code to meet the AC of the ticket is met, but this added bug fix is blocking me from approving. If you are adding in some serious logic to handle a corner case then a test to double check your fix works is needed. Also I am not convinced we should be adding null data to the cache, especially by repeating what we have in fetch cache. Can you provide more background on this error? I would really support breaking out this bug handling in a different PR so that we can merge this and close the ticket this is associated with and then handle the bug on its own.

src/registrar/models/domain.py
```diff
@@ -1293,6 +1293,51 @@ def _fetch_cache(self, fetch_hosts=False, fetch_contacts=False):
...
+                if existing_contacts.count() > 0:
+                    for choice in contacts_dict:
+                        contacts_dict[choice] = existing_contacts.get(
+                            contact_type=choice
+                        ).registry_id
+                    # Edge case for registrant
+                    registrant = PublicContact.ContactTypeChoices.REGISTRANT
+                    registrant_id = existing_contacts.get(
+                        contact_type=registrant
+                    ).registry_id
```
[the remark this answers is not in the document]
zandercymatics: For some background on this:
When clicking the "view" button on a deleted domain on the registrar / .gov side, it'll error out (500) if cache gets cleared for that object - meaning you can't view deleted domains.

The reason for this is that, because cache is empty, it calls _fetch_cache. _fetch_cache will then call `_get_or_create_domain()` which inevitably invokes the function `pendingCreate()` (because it was deleted from the registry). Since pendingCreate violates the FSM rule of unknown => dns_needed (deleted => dns_needed), it fails. As a test, overriding that causes other issues with how the page is currently setup because it still has the save buttons, etc.

I chose to handle this in an exception as it was minimally code invasive, but there are multiple solutions to it, so I have no problem changing my approach.

As for your comment, I think that is a good idea generally, but not in the context of this particular solution. I personally would opt instead to outright not show the security contact page / info, plus the other stuff like nameservers etc if they will always be None

src/registrar/models/domain.py
```diff
@@ -1293,6 +1293,51 @@ def _fetch_cache(self, fetch_hosts=False, fetch_contacts=False):
         except RegistryError as e:
             logger.error(e)
 
+        except TransitionNotAllowed as err:
+            # Fixes a bug with _fetch_cache trying to create
+            # a deleted domain, as cache gets cleared on delete.
+            # Instead, serve what we have locally.
+            if self.state == self.State.DELETED:
+                logger.warning("Attempted to create a deleted domain")
+                data = self._cache
```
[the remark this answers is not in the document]
zandercymatics: Removed

src/registrar/models/domain.py
```diff
@@ -1293,6 +1293,51 @@ def _fetch_cache(self, fetch_hosts=False, fetch_contacts=False):
         except RegistryError as e:
             logger.error(e)
 
+        except TransitionNotAllowed as err:
+            # Fixes a bug with _fetch_cache trying to create
+            # a deleted domain, as cache gets cleared on delete.
```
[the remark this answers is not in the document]
zandercymatics: removed

src/registrar/models/domain.py
```diff
@@ -1293,6 +1293,51 @@ def _fetch_cache(self, fetch_hosts=False, fetch_contacts=False):
...
+            # Instead, serve what we have locally.
+            if self.state == self.State.DELETED:
+                logger.warning("Attempted to create a deleted domain")
```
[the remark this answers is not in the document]
zandercymatics: removed

src/registrar/models/domain.py
```diff
@@ -880,8 +880,8 @@ def get_contact_in_keys(self, contacts, contact_type):
 
             return self._handle_registrant_contact(desired_contact)
 
-        _registry_id: str
-        if contact_type in contacts:
+        _registry_id: str = ""
```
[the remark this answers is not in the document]
zandercymatics: Yeah. It didn't like that line without the type specified, then it didn't like the line without a default

abroddrick (approved):
Thanks for separating out the bug it looks good!

PR #1162, opened 2023-10-17
Ticket #1054: Connection pooling

### LINKED ISSUE #1054: Increase efficiency on epp calls
### Issue Description

The calls and current implementation for Epp results in a slow down on the epp pages especially when first going to them. Investigate how to make this more efficient excluding the idea of not checking hosts/contacts when not needed as this is covered in ticket #1053. Threading or multi-processing may be helpful for this but investigation should be done into how best to approach this

AC:
- collect ideas to make the caching more efficient and investigate a few choices
- investigate making the calls more efficient, by implementing code changes
- consider using a library to test the time efficiency of the changes to measure the efficiency change in a tangible way

Create any follow on tickets based on results of investigation as needed

_No response_

### Issue Links

blocked by: #1053

zandercymatics:
## Ticket

Resolves #1054 

## Changes

- Includes gevent and geventconnpool in the pipfile / piplock / requirements file
- Modifies client.py to incorporate pooling behavior. Three new variables are created on init: pool_options, _pool, and pool_status
- Changed the _send function in client.py, with an edge case included for timeouts.
- Changed the send function to check for an edge case where the pool isn't running, then restart it
- Added these functions in client.py: get_pool, _create_pool, kill_pool, _test_registry_connection_success, and start_connection_pool
- Modified socket.py to mesh with the pool
- New files (+ folder) in utility: pool.py, pool_error.py, pool_status.py. Used for pooling behavior
- Modified settings.py to allow for global pool settings, such as the pool size and the timeout period
- Added tests for the pool
- Changed ADR status from "In Review" to "Approved" as this ticket merging indicates that it was approved

## Context for reviewers
This ticket implements pooling, but it doesn't modify our caching behavior or remove redundant calls. Because of this, the calls to EPP itself are inherently inefficient. That said, we get a huge improvement just from pooling alone. From testing, 85-90% of the time we see improved load times. However, some calls will still "hang" on occasion (given that we have so many, the chance of this is high).

If load performance is consistently bad, then that means the pooling isn't working as we expect. If this only occurs rarely, then its likely just a slow EPP call. The general rule of thumb here is that this should not be happening often, but on rare occasion. 

Another thing to note is that if a catastrophic failure occurs with the pool (we cannot contact the registry or its otherwise broken), it will restart itself. I have not seen this occur without manually triggering it and/or creating an intentional timeout situation. If this happens, you can confirm this is happening by looking at my sandbox logs. This restart period should not  take very long 
## Context for designers
Since this touches on error handling, this may need to be checked. The current errors in place should be unchanged, though its worth testing just to make sure they still are consistent with what is expected.

## Setup
**Note:** When testing this locally, you will need to do a docker-compose down, docker-compose build, and docker-compose up because new packages were added.
Go to an approved domain, click manage domain and navigate between contacts and DNS. Try adding those, and note the performance difference between them. Should be substantially improved.

The EPP related buttons in /admin should also be working as expected.

**Note 2:** The AC mentions that the shared connection should be closed after a certain amount of time. However, in testing, this is inefficient and it makes more sense to keep this running due to minimal performance impact. If this is not desired, the "keepalive" setting should just be set to None and this will disable that behavior. With this disabled, the connection eventually gets culled after a timeout period (set by the registry / EPPLib) and a new pool will be spun up whenever a request is made for the first time since it got culled. This leads to inconsistent performance (as a new pool has to be spun up each time), it is much more consistent for the end user if we just keep this connection open - which we currently do.

- [x] Did dependency updates in Pipfile also get changed in requirements.txt?

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


zandercymatics:
These changes are not on my sandbox at this exact moment as I am doing a merge for an older PR. Will send an update message when they are back

zandercymatics:
This is back on my sandbox

src/registrar/models/domain.py
```diff
@@ -58,7 +59,7 @@ class Domain(TimeStampedModel, DomainHelper):
     G) Activation is controlled by the registry. It will happen automatically when the
        domain meets the required checks.
     """
-
+    # test comment for pushing to sandbox - will remove
```
abroddrick, line 62: **change before merging**: don't forget to delete this

src/registrar/tests/test_models_domain.py
```diff
@@ -873,7 +873,6 @@ def test_contact_getter_registrant(self):
             contact_id="regContact",
             contact_type=PublicContact.ContactTypeChoices.REGISTRANT,
         )
-
```
abroddrick, line 876: **non blocking suggestion**: undo accidentally removed line

zandercymatics:
@abroddrick Got those two changes in

Katherine-Osos (approved):

src/epplibwrapper/socket.py
```diff
@@ -14,24 +16,82 @@
...
+            # Occurs when an invalid creds are passed in - such as on localhost
+            except OSError as err:
+                logger.error(err)
+                return False
+            else:
+                self.disconnect()
+
+                # If we encounter a login error, fail
+                if self.is_login_error(response.code):
+                    logger.warning("was login error")
```
abroddrick, line 93:
```suggestion
                    logger.warning("A login error was found in test_connection_success")
```

Question: how can this every be true, when in client.send you're raising a LoginError if  the response code matches that of a login error.

src/epplibwrapper/utility/pool.py
```diff
@@ -0,0 +1,133 @@
...
+    """A connection pool for EPPLib.
+
+    Args:
+        client (Client): The client
+        login (commands.Login): Login creds
+        options (dict): Options for the ConnectionPool
+        base class
+    """
+
+    def __init__(self, client, login, options: dict):
```
zandercymatics, line 30: Should this change from a dict to just **options? Its a dict for intentional separation (so you _have_ to pass in an 'options' variable when using this), but if this is more confusing than not I can just remove it

src/epplibwrapper/tests/test_pool.py
```diff
@@ -0,0 +1,257 @@
...
+    pass
+
+logger = logging.getLogger(__name__)
+
+
+class TestConnectionPool(TestCase):
+    """Tests for our connection pooling behaviour"""
+
+    def setUp(self):
+        self.pool_options = {
```
abroddrick, line 31:
```suggestion
        # Mimic the settings added to settings.py
        self.pool_options = {
```

src/epplibwrapper/utility/pool.py
```diff
@@ -0,0 +1,133 @@
...
+        if "exc_classes" in options:
+            self.exc_classes = options["exc_classes"]
+
+        self.keepalive = None
+        if "keepalive" in options:
+            self.keepalive = options["keepalive"]
+
+        # Determines the period in which new
+        # gevent threads are spun up
+        self.spawn_frequency = 0.1
```
abroddrick, line 55: **comment**: maybe some context here as to what the measurement is. is that .1 of a second?

src/epplibwrapper/utility/pool.py
```diff
@@ -0,0 +1,133 @@
...
+            logger.error(message, exc_info=True)
+            # We want to raise a pool error rather than a LoginError here
+            # because if this occurs internally, we should handle this
+            # differently than we otherwise would for LoginError.
+            raise PoolError(code=PoolErrorCodes.NEW_CONNECTION_FAILED) from err
+
+    def _keepalive(self, c):
+        """Sends a command to the server to keep the connection alive."""
+        try:
+            # Sends a ping to EPPLib
```
abroddrick, line 80:
```suggestion
            # Sends a ping to the registry via EPPLib
```
**nitpick**

src/epplibwrapper/utility/pool.py
```diff
@@ -0,0 +1,133 @@
...
+
+                self.greenlets.clear()
+                self.conn.clear()
+
+                # Clear the semaphore
+                self.lock = BoundedSemaphore(self.size)
+            else:
+                logger.info("No connections to kill.")
+        except Exception as err:
+            logger.error("Could not kill all connections.")
```
abroddrick, line 110: **comment** you have this as an error in pool_error, why not use that here? It seems that error was made and never used

abroddrick (approved):
See comments first, some are just suggestions, but be sure to address (or at least read) before merging.

src/registrar/config/settings.py
```diff
@@ -534,6 +534,18 @@
 SECRET_REGISTRY_KEY_PASSPHRASE = secret_registry_key_passphrase
 SECRET_REGISTRY_HOSTNAME = secret_registry_hostname
 
+# Use this variable to set the size of our connection pool in client.py
+# WARNING: Setting this value too high could cause frequent app crashes!
```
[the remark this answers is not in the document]
zandercymatics: Good call, will leave a comment.

The reason is as follows: 
Having multiple pools means having to open multiple connections concurrently (same reason why we had slow load times before). Because of that slow load time, we had to change that timeout variable on our sandboxes. That timeout period is able to handle 6 connections being created at once, but if we set the pool to like 50 or 100 that could take a while to complete. If the pool needs to restart for any reason then it'll go through that process again

src/epplibwrapper/client.py
```diff
@@ -60,44 +64,99 @@ def __init__(self) -> None:
...
+                raise LoginError(
+                    "Couldn't connect to the registry after three attempts"
+                )
+            with self._pool.get() as connection:
+                response = connection.send(command)
+        except Timeout as t:
+            # If more than one pool exists,
+            # multiple timeouts can be floating around.
+            # We need to be specific as to which we are targeting.
+            if t is timeout:
```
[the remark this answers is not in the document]
zandercymatics: Yup! Basically, lets say that we are running three pools at once. Two of them are dead, but the third is running fine. gevent.timeouts are _global_ so if this line doesn't exist, it will kill all three pools rather than the dead ones.
https://www.gevent.org/api/gevent.timeout.html
image

src/epplibwrapper/client.py
```diff
@@ -60,44 +64,99 @@ def __init__(self) -> None:
...
-            logger.warning(message, cmd_type, exc_info=True)
+            # For linter
+            text = "failed to execute due to a registry login error."
+            message = f"{cmd_type} {text}"
+            logger.error(message, exc_info=True)
             raise RegistryError(message) from err
         except Exception as err:
-            message = "%s failed to execute due to an unknown error." % err
-            logger.warning(message, cmd_type, exc_info=True)
+            message = f"{cmd_type} failed to execute due to an unknown error."
```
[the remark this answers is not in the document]
zandercymatics: I did not know that about our production environment, I'll add those!

src/epplibwrapper/socket.py
```diff
@@ -14,24 +16,82 @@
...
         """Close the connection."""
         try:
             self.client.send(commands.Logout())
             self.client.close()
         except Exception:
             logger.warning("Connection to registry was not cleanly closed.")
+
+    def send(self, command):
+        """Sends a command to the registry.
+        If the response code is >= 2000,
```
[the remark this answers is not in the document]
zandercymatics: That is fair, I'll specify instead what error type is returned and the situation in which that happens. Error codes >= 2000 indicate a bad login/response from the registry basically (defined in the RegistryError class)

src/epplibwrapper/socket.py
zandercymatics: True, this I think is a code leftover from when I was manually doing code checking here. Will remove as I changed the LoginError behavior since then

dave-kennedy-ecs (commented):
(Non-blocking comment) I have run code locally and in sandbox.  My only suggestion has to do with the initialization.  It seems to me that the initialization of the pool of connections happens on the first EPP request, not on app initialization.  This causes the first access of a domain (by a user) to take more than 19 seconds with the current configuration.  This may be acceptable.  I think it would be better if the initialization could happen at app initialization rather than on demand.

Per transaction improvement is noticeable.

This is a very complicated ticket.  And I think it was well done.  Nice job, Zander.

zandercymatics:
@dave-kennedy-ecs That is interesting. Also quite strange that it is taking that long. Looking into if its feasible to do but would definitely be worth splitting into another ticket.

PR #1183, opened 2023-10-20
Add Vicky to fixtures

rachidatecs:


src/registrar/fixtures_users.py
```diff
@@ -86,6 +86,12 @@ class UserFixture:
             "first_name": "Kristina",
             "last_name": "Yin",
         },
+        {
+            "username": "ac49d7c1-368a-4e6b-8f1d-60250e20a16f",
+            "first_name": "Vicky",
+            "last_name": "Chin",
+            "email": "szu.chin@associates.cisa.dhs.gov",
```
zandercymatics, line 93: Just making sure: does she want to use this email for Admin?

zandercymatics (approved):
LGTM. Only thing is the cisa email being used for the admin account which works kind of backward in my head to the pattern others have used, but its a non issue if she prefers it

src/registrar/fixtures_users.py
therealslimhsiehdy: +1 to Zanders comment, but it's non-blocking.

If Vicky doesn't wan't to keep swapping between two email tabs, we may want to keep up with the some_email for Admin and then some_email+1 for Analyst pattern

therealslimhsiehdy (approved):

dave-kennedy-ecs (approved):

PR #1182, opened 2023-10-20
Remove leading spaces to fix our issue template

h-m-f-t:
Fixes our issue template, which got borked in #1137.

PR #1137, opened 2023-10-06
More issue template refinement

h-m-f-t:
This follows the PR at #1113 and improves our issue template. 

* I've revised the text a little more so it should fit on 2 lines in situ (expected). 
* To make it easier for people to know which emoji to use, I've added the Github :emoji: names in the description text. 
* I also struck the placeholder text for the A/C, which felt too heavy.

This PR doesn't change required fields: only "Issue description and context" is mandatory.

PR #1113, opened 2023-10-03
Update issue template to add A/C and clarity

h-m-f-t:
This PR updates our `issue` issue template to add acceptance criteria and attempts clearer language.

## Context for reviewers

* It should be easy for our team and other interested people to submit issues to this repo. Though we should avoid overprescription, including common fields and reminders makes the work of creation better for the creators. 
* There is also a `story` issue template. Perhaps we should cut that and include relevant content here? It seems like we could recommend an `issue`-writer start with a user story and drop the other template.
* We currently use issues to determine ground truth for feature development and regularly edit issues to improve them – even ones other people submitted. That's not a common approach, and someone who wasn't expecting it might be offended by it. Should we give a heads up in our template and say e.g., "Issues may be revised by the product team to improve our understanding."?

PR #1184, opened 2023-10-20
Fix double emojis in issue template

h-m-f-t:
This is a small fix to #1182, which was a small fix to #1137, which was a small fix to #1113. This PR replaces :emojicode: with square brackets and extends a hyperlink.

(Maybe issue templates were a bad idea??)

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue-default.yml
+++ b/.github/ISSUE_TEMPLATE/issue-default.yml
@@ -12,7 +12,7 @@ body:
     attributes:
       label: Issue description
       description: |
-        Describe the issue so that someone who wasn't present for its discovery can understand why it matters. Use full sentences, plain language, and good [formatting](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax).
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/issue-default.yml
```diff
@@ -31,7 +31,7 @@ body:
     attributes:
       label: Links to other issues
       description: |
-        "Add issue #numbers this relates to and how (e.g., 🚧 :construction: Blocks, ⛔️ :no_entry: Is blocked by, 🔄 :repeat: Relates to)."
+        "Add issue #numbers this relates to and how (e.g., 🚧 [construction] Blocks, ⛔️ [no_entry] Is blocked by, 🔄 [arrows_counterclockwise] Relates to)."
```
zandercymatics, line 34: I think this square bracket notation w/out a link may have been broken in older github versions, but now it renders normally. If this still renders you'd need to use an escape character so like \[construction\] but I don't think its an issue. I tested it in a markdown file and it worked so I honestly don't see why it wouldn't here

h-m-f-t: Square brackets also worked for me. We'll find out when it's merged whether it actually works on GitHub!

abroddrick (approved):
I wanted to approve just based on the description of the PR alone!  Feel free to merge this in.

PR #1211, opened 2023-10-26
Issue 1192: Enable the deletion of NS records #1 and #2 if there are enough records in the formset

### LINKED ISSUE #1192: Unable to remove the first or second element on nameservers page
### Current Behavior

If you try to delete the first or second nameserver, even with more than 2 nameservers on that page, you are unable to do so because the page is marking the first two boxes as required. 

### Expected Behavior

You should be able to delete the first two nameservers if you already entered more than 2 nameservers.

For **refinement**: do we want to allow the first two nameservers to always be deletable? If we did, a warning pop-up should occur as that will take their domain offline, so if we want to allow that it should be another ticket. 
Result of refinement: DO NOT do a pop-up, and don't allow the first two nameservers to always be deletable. Just only allow delete if they have more than 2 nameservers prior to the delete. 

Additional acceptance criteria:
- [ ] Amend error message with clarifying text to indicate that at least 2 NS are required.

### Steps to Reproduce

1.Go to the nameservers page for a domain
2. fill in 2 or more nameservers
3. try to delete the first or second nameserver
4. notice you are unable to. 

### Environment

_No response_

### Additional Context

image

### Issue Links

_No response_

rachidatecs:
## Ticket

Resolves #1192 

## Changes

- Wrap the label text in a span for NS labels so that we can do JS textContent swaps without destroying abbr child nodes
- JS to manipulate the DOM:
  - Enable/disable* delete buttons on NS 1 and 2 as needed
  - Clean up the clone source on add and make sure the refactor of forms in the formset is clean as well, on delete (required, abbr, etc.)

## Context for reviewers

Note: We did not write any tests for the JS.

## For designers:

Try to add more than 13 nameservers. Experiment with the delete (especially NS1 and NS2 when you have lots of NSs and when you have 3, 2 NSs). We ran ANDI and it's OK with the USWDS light grey on the disabled delete button (another approach could be to completely remove the buttons where applicable but I liked disabling a bit more, as it feels more intuitive as to what's going on). Check all possible error scenarios (less than 2 NS, bad NS, bad IP, subdomain with no IP, IP with no subdomain, IP with no NS at all).

## Setup

Add and delete forms in the formset.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [x] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


gabydisarli:
non-blocking, but it would be nice to have the form default to just two name server field sets since those are the only two required, and then the user could add more if they wanted to.

gabydisarli (approved):
This looks good to me. There is a new registry error that I would like to get implemented but I am waiting on confirmation from @h-m-f-t and @michelle-rago. 

Here is the copy:
https://docs.google.com/document/d/12qxl6evZw16dT-TB9iu3YLjKdwwL5KqIYQJV4uncAXg/edit#heading=h.30j0zll

Note that there are specific page versions of the error message for user entered "bad data" and then there is a generic system can't connect message.

gabydisarli:
> This looks good to me. There is a new registry error that I would like to get implemented but I am waiting on confirmation from @h-m-f-t and @michelle-rago.
> 
> Here is the copy: https://docs.google.com/document/d/12qxl6evZw16dT-TB9iu3YLjKdwwL5KqIYQJV4uncAXg/edit#heading=h.30j0zll
> 
> Note that there are specific page versions of the error message for user entered "bad data" and then there is a generic system can't connect message.

@rachidatecs the error message for the name servers page is approved 🎉

src/registrar/forms/domain.py
```diff
@@ -23,19 +23,28 @@ class DomainAddUserForm(forms.Form):
...
+        required=False,
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(DomainNameserverForm, self).__init__(*args, **kwargs)
+
+        # add custom error messages
+        self.fields["server"].error_messages.update(
+            {
+                "required": "A minimum of 2 Name Servers are required.",
```
abroddrick, line 45: should we capitalize Name Servers when it's not in a title? tagging @gabydisarli

abroddrick (commented):
image
If I have 3 nameserver boxes but only two filled out, I can still click the nameservers delete button for nameserver 1 or 2 but if I delete the third box (even though it's empty) now I can't delete nameserver 1 or 2. Is this the intent? Shouldn't we instead remove disable the delete button still if only two nameserver boxes are filled out?

therealslimhsiehdy:
**Bug:** If I have 13 nameservers and hit save -- I'm still able to click "Add another name server" and still have a 14th nameserver pop up and save it

rachidatecs:
> image If I have 3 nameserver boxes but only two filled out, I can still click the nameservers delete button for nameserver 1 or 2 but if I delete the third box (even though it's empty) now I can't delete nameserver 1 or 2. Is this the intent? Shouldn't we instead remove disable the delete button still if only two nameserver boxes are filled out?

In that case, submitting the form will throw an error saying "2 nameservers are required" so it's not a "bad" deadend.

More context:
We're not taking into account whether an NS record is filled or not when controlling for the delete buttons. If we want to do so, we'd need to iterate over all records and test for content on page load + on every add or delete, which is not too bad.

More UI complications:
What if a user has multiple records, deletes data in some of them, ends up with eg 2 filled and 3 empties. In that case, the delete buttons would be disabled which would be weird UX (why can't I delete a blank record) unless we disable the delete buttons on filled and keep them on blanks. Seems like a very fussy UI/UX to me.

rachidatecs:
> **Bug:** If I have 13 nameservers and hit save -- I'm still able to click "Add another name server" and still have a 14th nameserver pop up and save it
> 
> Screenshot 2023-10-30 at 2 18 19 PM
> And +1 to Alysia's screenshot error as well, I'm running into that bug as well 😅

Should be an easy fix (basically call the JS function that checks for delete btn status on page load). I'll handle.

src/registrar/assets/js/get-gov.js
```diff
@@ -397,9 +437,20 @@ function prepareDeleteButtons(formLabel) {
...
-        addButton.classList.add("display-none")
+        // addButton.classList.add("display-none");
+        addButton.setAttribute("disabled", "true");
+      }
+
+      if (isNameserversForm && forms.length >= 2) {
+        // Enable the delete buttons on the nameservers
+        forms.forEach((form, index) => {
+          Array.from(form.querySelectorAll('.delete-record')).forEach((deleteButton) => {
+            // deleteButton.classList.remove("display-none");
```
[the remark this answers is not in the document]
rachidatecs: I left these comments in place in case we decide we want to completely hide btns instead of adding a disabled attribute. I can remove.

rachidatecs:
> non-blocking, but it would be nice to have the form default to just two name server field sets since those are the only two required, and then the user could add more if they wanted to.

This is actually a bit harder than it looks, since in an empty form the JS looks at the 3rd (non-required) record to clone on that add button. We're already cleaning this data when we delete down to 2 and then add back (after page load), so it can totally be done but in another quick fix ticket instead of this PR

src/registrar/forms/domain.py
gabydisarli: No, name servers should not be capitalized, good catch!

michelle-rago:
Side note: "IP Address" should be "IP address" (lowercase a). Can someone fix that? Thanks!

rachidatecs:
> IP Address

Fixed

abroddrick (approved):
Approving as the current bug where you can't submit an already created nameserver is going to be solved in Dave's PR. Everything else looks good

PR #1225, opened 2023-10-27
A bit more verbosity in send_domain_invitations command

neilmb:
We use the `send_domain_invitations` command during our data migration to send invitation emails to new users. If we don't give the `-s` option then it doesn't send emails, but it also doesn't print anything out about what it would do. This makes it a little bit more chatty and prints the emails that it would have sent.

@dave-kennedy-ecs just some tweaks on your script that were useful as we played around with it this afternoon

### DIFF
--- a/src/registrar/management/commands/send_domain_invitations.py
+++ b/src/registrar/management/commands/send_domain_invitations.py
@@ -45,8 +45,10 @@ class Command(BaseCommand):
         self.transition_domains = TransitionDomain.objects.filter(
             email_sent=False,
         ).order_by("username")
+        logger.info("Found %d transition domains", len(self.transition_domains))
 
         self.build_emails_to_send_array()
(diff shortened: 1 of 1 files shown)


src/registrar/management/commands/send_domain_invitations.py
```diff
@@ -58,6 +60,12 @@ def handle(self, **options):
             logger.info("done sending emails and updating transition_domains")
         else:
             logger.info("not sending emails")
+            for email_context in self.emails_to_send:
+                logger.info(
+                    "would send email to %s for %s",
+                    email_context["email"],
+                    email_context["domains"],
+                )
```
zandercymatics, line 68: (optional) One thing that popped into my head was that using Nicolles `TerminalColors` class would be neat here (yellow?) for visual distinction, but I don't think that's used elsewhere in this file

abroddrick (approved):
LGTM

PR #1227, opened 2023-10-30
Add link to rotate_application_secrets.md

h-m-f-t:
A very small documentation update that simply adds a link Login.gov's recommendation for generating a public cert.


neilmb (approved):
Great pointer back to where we probably found it in the first place.

PR #1228, opened 2023-10-30
fix migration 43 - domain expiration date

rachidatecs:
## Ticket

Does not resolve a ticket, this is a hotfix

## Changes

- Fix migration issue (0043 - domain expiration)


rachidatecs:
Creating another PR that uses a named branch to trigger workflows

PR #1236, opened 2023-10-30
Issue 1101 - Regex form validation on nameservers

### LINKED ISSUE #1101: Regex form validation on nameserver entry field
### Issue Description

To avoid sending an EPP call to set a nameserver that is an invalid format already, have front end error checking on the nameserver field. This error checking should check if the nameserver string entered in the text box is a valid based on the [rfc952](https://datatracker.ietf.org/doc/html/rfc952.txt)

AC:
- The first character must be an alpha **or digit** character.
- The last character must not be a minus sign or period.
- must be longer than 1 character
- **max of 253 ascii characters including delimiting dots**
- **each label (region between dot delimiters) should be 1-63 octets long.**
- alphabet **(a-z) lowercase**, digits (0-9), minus sign (-), and period (.). No whitespaces allowed.
- check the RFC for additional items that can be checked on the front end to ensure better error handling

_No response_

### Issue Links

_No response_

dave-kennedy-ecs:
## Ticket

Resolves #1101

## Changes

- Form validation on nameservers to satisfy the following conditions:
  - The first character must be an alpha or digit character.
  - The last character must not be a minus sign or period.
  - must be longer than 1 character
  - max of 253 ascii characters including delimiting dots
  - each label (region between dot delimiters) should be 1-63 octets long.
  - alphabet (a-z) lowercase, digits (0-9), minus sign (-), and period (.). No whitespaces allowed.

## Context for reviewers

Prior to this change, when invalid domain was entered, form errored and error message read, "Value was wrong." and the form was blanked out.  After this change, the form should identify properly the field which is invalid and display a proper error message.

To test, go to domain management page, to the nameservers section.  Enter in invalid host names.  Some examples: localhost, gobbledegook, example.gov, www.dddddddddddddddddddd.gov

After clicking Save, proper error message should display.  And form values should still be present in the form.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### DIFF
--- a/src/registrar/forms/domain.py
+++ b/src/registrar/forms/domain.py
@@ -51,9 +51,8 @@ class DomainNameserverForm(forms.Form):
 
         ip_list = self.extract_ip_list(ip)
 
-        if ip and not server and ip_list:
-            self.add_error("server", NameserverError(code=nsErrorCodes.MISSING_HOST))
-        elif server:
+        # validate if the form has a server or an ip
+        if ip and ip_list or server:
             self.validate_nameserver_ip_combo(domain, server, ip_list)
 
         return cleaned_data
@@ -84,6 +83,20 @@ class DomainNameserverForm(forms.Form):
                     "ip",
                     NameserverError(
                         code=nsErrorCodes.MISSING_IP, nameserver=domain, ip=ip_list
+                    ),
+                )
+            elif e.code == nsErrorCodes.MISSING_HOST:
+                self.add_error(
+                    "server",
+                    NameserverError(
+                        code=nsErrorCodes.MISSING_HOST, nameserver=domain, ip=ip_list
+                    ),
+                )
+            elif e.code == nsErrorCodes.INVALID_HOST:
+                self.add_error(
+                    "server",
+                    NameserverError(
+                        code=nsErrorCodes.INVALID_HOST, nameserver=server, ip=ip_list
                     ),
                 )
             else:
--- a/src/registrar/forms/domain.py
+++ b/src/registrar/forms/domain.py
@@ -44,6 +44,8 @@ class DomainNameserverForm(forms.Form):
         cleaned_data = super().clean()
         self.clean_empty_strings(cleaned_data)
         server = cleaned_data.get("server", "")
+        # remove ANY spaces in the server field
+        server = server.replace(" ", "")
         ip = cleaned_data.get("ip", None)
         # remove ANY spaces in the ip field
         ip = ip.replace(" ", "")
--- a/src/registrar/forms/domain.py
+++ b/src/registrar/forms/domain.py
@@ -46,6 +46,9 @@ class DomainNameserverForm(forms.Form):
         server = cleaned_data.get("server", "")
         # remove ANY spaces in the server field
         server = server.replace(" ", "")
+        # lowercase the server
+        server = server.lower()
+        cleaned_data["server"] = server
         ip = cleaned_data.get("ip", None)
         # remove ANY spaces in the ip field
         ip = ip.replace(" ", "")
--- a/src/registrar/forms/domain.py
+++ b/src/registrar/forms/domain.py
@@ -52,7 +52,7 @@ class DomainNameserverForm(forms.Form):
         ip_list = self.extract_ip_list(ip)
 
         # validate if the form has a server or an ip
-        if ip and ip_list or server:
+        if (ip and ip_list) or server:
             self.validate_nameserver_ip_combo(domain, server, ip_list)
 
         return cleaned_data
--- a/src/registrar/models/domain.py
+++ b/src/registrar/models/domain.py
@@ -305,12 +305,33 @@ class Domain(TimeStampedModel, DomainHelper):
         return bool(regex.match(nameserver))
 
     @classmethod
+    def isValidDomain(cls, nameserver: str):
+        """Checks for validity of nameserver string based on these rules:
+        - first character is alpha
+        - last character is not - or .
+        - all characters alpha, 0-9, -, or .
+        - 2 character min, 24 character max
+        """
+        # pattern to test for valid domain
+        # pattern = r'^[a-zA-Z][a-zA-Z0-9-.]{0,22}[a-zA-Z0-9]$'
+        pattern = r"^[a-zA-Z][a-zA-Z0-9-.]*(\.[a-zA-Z0-9-]+){2}[a-zA-Z0-9]$"
+
+        # attempt to match the pattern
+        match = re.match(pattern, nameserver)
+
+        # return true if nameserver matches, and length less than 25;
+        # otherwise false
+        return bool(match) and len(nameserver) < 25
+
+    @classmethod
     def checkHostIPCombo(cls, name: str, nameserver: str, ip: list[str]):
         """Checks the parameters past for a valid combination
         raises error if:
             - nameserver is a subdomain but is missing ip
             - nameserver is not a subdomain but has ip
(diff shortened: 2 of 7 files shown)


erinysong:
Is there any information available on the registrar for users to know the naming requirements for their nameservers? In other words, are these requirements accessible for the users to also know?
image

gabydisarli:
> Is there any information available on the registrar for users to know the naming requirements for their nameservers? In other words, are these requirements accessible for the users to also know? image

Good call out
@h-m-f-t and I discussed this and we decided having the example name server would be good enough for now. We will also be adding content like this to this help page on the website. However, we're not linking to that page on this page, which might be a good idea. Thoughts?

Also just want to note that we won't be mentioning the white space thing as those should be automatically removed when entered.

gabydisarli:
I know this isn't part of this PR but hoping you can fix these small typos that I've just noticed. There are several instances of hatch marks used in the alert instead of proper quotations. Can we fix these? Image shows the hatch marks in red and the proper quotation marks in green.

 Screenshot 2023-10-31 at 2 13 26 PM

gabydisarli:
@dave-kennedy-ecs I'm not able to trigger any of the front-end formatting validation on the name servers field. I'm only getting the registry error "Value entered was wrong." Am I missing something?

dave-kennedy-ecs:
I'm sorry, we moved this PR over to the [bl sandbox](https://getgov-bl.app.cloud.gov/) because there was another PR (for 996) staged on the dk sandbox

dave-kennedy-ecs:
> I know this isn't part of this PR but hoping you can fix these small typos that I've just noticed. There are several instances of hatch marks used in the alert instead of proper quotations. Can we fix these? Image shows the hatch marks in red and the proper quotation marks in green.
> 
> Screenshot 2023-10-31 at 2 13 26 PM

Are you sure the red and green were not reversed?  I have replaced all 4 of these with quotation marks.

erinysong:
@gabydisarli that sounds good! I think having some sort of way to have this page display nameserver requirements may be good, since that could give some more insight into why a nameserver request may not be saving

src/registrar/forms/domain.py
```diff
@@ -51,9 +51,8 @@ def clean(self):
 
         ip_list = self.extract_ip_list(ip)
 
-        if ip and not server and ip_list:
-            self.add_error("server", NameserverError(code=nsErrorCodes.MISSING_HOST))
-        elif server:
+        # validate if the form has a server or an ip
+        if ip and ip_list or server:
```
neilmb, line 55: **suggestion (non-blocking)**: Could we parenthesize this conditional to make it more clear what the intention is? I think that Python's interpretation is going to be
```suggestion
        if (ip and ip_list) or server:
```
but I think it's easier to read if we don't have to remember the precedence rules.

src/registrar/models/domain.py
```diff
@@ -304,13 +309,40 @@ def isSubdomain(cls, name: str, nameserver: str):
         regex = re.compile(full_pattern)
         return bool(regex.match(nameserver))
 
+    @classmethod
+    def isValidDomain(cls, nameserver: str):
```
neilmb, line 313: **question (non-blocking)**: It's not very easy to find, but we already have a `Domain.string_could_be_domain()` method with similar regexes in it. Can we clarify somehow if there is a difference between how this and that behave and where we should be using each one?

src/registrar/models/domain.py
```diff
@@ -304,13 +309,40 @@ def isSubdomain(cls, name: str, nameserver: str):
         regex = re.compile(full_pattern)
         return bool(regex.match(nameserver))
 
+    @classmethod
```
neilmb, line 312: **suggestion (blocking)**: This function seems like a perfect candidate for some simple unit tests that check that it performs correctly on various types of input that either conform with the rules or don't. Maybe in `registrar/tests/test_models_domain.py` or `test_models_domain_helper.py`?

neilmb (commented):
Looks like fine code, but the checking function could use some unit tests and we might already have a function that checks this, or at least something like this?

gabydisarli:
> > I know this isn't part of this PR but hoping you can fix these small typos that I've just noticed. There are several instances of hatch marks used in the alert instead of proper quotations. Can we fix these? Image shows the hatch marks in red and the proper quotation marks in green.
> > Screenshot 2023-10-31 at 2 13 26 PM
> 
> Are you sure the red and green were not reversed? I have replaced all 4 of these with quotation marks.

No they are not reversed. The green circle show's what they should look like ” vs "

src/registrar/forms/domain.py
dave-kennedy-ecs: done

src/registrar/models/domain.py
dave-kennedy-ecs: There was some redundant code, which became obsolete in previous iterations.  I removed redundant code and obsolete code.  Part of the reason for doing this, rather than refactoring to the older model is that (1) the older model was processing all tuples at once and not one form at a time, and (2)  I am anticipating that there could be a future desire for different NameserverErrors and error messages for users to be raised for different conditions, and would like these exceptions to be raised from the same area in the code for code readability.

src/registrar/models/domain.py
dave-kennedy-ecs: @neilmb I have added test cases and pushed.  Would you mind taking a look again?

neilmb (commented):
Tests look great @dave-kennedy-ecs, thanks for adding those.

neilmb (approved):
Let's ship this when you're ready @dave-kennedy-ecs

gabydisarli:
Let's change the format validation message to "Enter a name server in the required format, like ns1.example.com" Note that there is no period at the end in order not to confuse users into thinking a period is required at the end. Ideally we'd have one summary error at the top of page and then each field throwing an error would throw an error. 
[Figma link](https://www.figma.com/file/aJbuDLJNNikqQObTuCNKQa/GetGov_Designs_ECS-Truss?type=design&node-id=1493%3A95184&mode=design&t=9Y3UmJoLl6Uj6QWD-1)

gabydisarli:
> Let's change the format validation message to "Enter a name server in the required format, like ns1.example.com" Note that there is no period at the end in order not to confuse users into thinking a period is required at the end. Ideally we'd have one summary error at the top of page and then each field throwing an error would throw an error. [Figma link](https://www.figma.com/file/aJbuDLJNNikqQObTuCNKQa/GetGov_Designs_ECS-Truss?type=design&node-id=1493%3A95184&mode=design&t=9Y3UmJoLl6Uj6QWD-1)

I'm also wondering about what should happen if the max of 253 ASCII characters is exceeded or the labels between dot delimiters exceeds 63. Like what happens if someone adds ns1.ABCDEFGHIJKLMNOPQRSTUVWXYZABCDEFGHIJKLMNOPQRSTUVWXYZABCDE.com?

dave-kennedy-ecs:
I have truncated the long error messages.  And updated the text of the error message.

gabydisarli:
Noticed another weird behavior, not sure if it's a bug, but, if I enter the two name server minimum, hit save, and then copy on name server into the 1st or 2nd field I'm able to save without getting the "minimum 2 required" error message.

dave-kennedy-ecs:
@gabydisarli I am able to reproduce.  It is not related to this particular ticket, but an edge case that we can consider handling in another ticket.

gabydisarli (approved):

Issue #1160, opened 2023-10-17
Configure the manage.get.gov application in Login.gov production


neilmb:
Login.gov production (as opposed to the sandbox) needs to be configured for our manage.get.gov application and that configuration takes time to be published by them. We should get it set up with plenty of lead time, but it needs to be after we set up the manage.get.gov domain name.

neilmb:
This configuration is done and the code for configuring it is running as expected on stable.

PR #1245, opened 2023-11-01
Configure a Login.gov production OIDC provider

### LINKED ISSUE #1151: Use multiple application instances on our stable environment
### Story

As a user
I want the registrar to be available as often as possible
so that I can manage my domains

### Acceptance Criteria

- [ ] Cloud.gov app in stable environment uses more than one instance: https://cloud.gov/docs/management/multiple-instances/
- [x] Cloud.gov app in stable environment uses zero-downtime deployments: https://cloud.gov/docs/deployment/production-ready/#zero-downtime-deploy

### Additional Context

In preparation for launch, we should consider reviewing other items on https://cloud.gov/docs/deployment/production-ready/ of which these are two and we have many of them already.

This ticket is exempt from automated testing requirement.

### Issue Links

_No response_

neilmb:
As part of #1160 we need to configure the application running on manage.get.gov to be able to use the production Login.gov environment rather than the identity sandbox environment like the rest of our sandboxes use. Fortunately, we made ourselves a design for that using a dict of `OIDC_PROVIDERS` and a settings variable for `OIDC_ACTIVE_PROVIDER`.

## Changes

This makes the `OIDC_ACTIVE_PROVIDER` load from an environment variable and adds the production login.gov provider information. In the stable environment's manifest, we set the `OIDC_ACTIVE_PROVIDER` environment variable to match, while all other environments will continue using the default value of `identity sandbox` when that environment variable is unset.

Also closes #1151 by moving us to two instances in the stable manifest.

## Setup

This is a hard one to test because it depends on having matching private keys for each environment in the `DJANGO_SECRET_LOGIN_KEY` environment variable.  You can at least test that this change keeps your sandbox or local install in working order by pulling the branch, running `docker compose up` and making sure that you can log in through the Identity sandbox.

### DIFF
--- a/ops/manifests/manifest-stable.yaml
+++ b/ops/manifests/manifest-stable.yaml
@@ -4,7 +4,7 @@ applications:
   buildpacks:
     - python_buildpack
   path: ../../src
-  instances: 1
+  instances: 2
   memory: 512M
   stack: cflinuxfs4
   timeout: 180
--- a/ops/manifests/manifest-stable.yaml
+++ b/ops/manifests/manifest-stable.yaml
@@ -23,6 +23,8 @@ applications:
     DJANGO_LOG_LEVEL: INFO
     # default public site location
     GETGOV_PUBLIC_SITE_URL: https://beta.get.gov
+    # Which OIDC provider to use
+    OIDC_ACTIVE_PROVIDER: login.gov production
   routes:
     - route: getgov-stable.app.cloud.gov
   services:
--- a/ops/manifests/manifest-staging.yaml
+++ b/ops/manifests/manifest-staging.yaml
@@ -4,7 +4,7 @@ applications:
   buildpacks:
     - python_buildpack
   path: ../../src
-  instances: 1
+  instances: 2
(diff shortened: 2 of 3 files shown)


ops/manifests/manifest-stable.yaml
```diff
@@ -4,7 +4,7 @@ applications:
   buildpacks:
     - python_buildpack
   path: ../../src
-  instances: 1
+  instances: 2
```
abroddrick, line 7: Does this make the application have two instances? if so I think that is covered in another ticket that you may want to link to this PR with the resolves keyword. Also should we have the same 2 instances on staging so staging is as close to stable as possible?

abroddrick (approved):
Tested and can still log in locally. See  one non blocking comment

ops/manifests/manifest-stable.yaml
neilmb: Good point, I'll mention the other ticket. And I can make the change to staging too, I think that sounds good.

PR #1355, opened 2023-11-16
Ticket #1243: Import Organization Data

### LINKED ISSUE #1243: Import "organization address" data
### Issue description

We need to import the "organization address" information from the legacy registry data to the new .gov registrar database, matching it with the appropriate existing "organization" records, so that all users can view this on the "Organization name and mailing address" page for domain management.

### Acceptance Criteria:

- [ ] Organization data, corresponding to each transition domain is updated and populated with the supplied "organization address" data in the `transition domain` and `domain information` tables, having been sourced from yyyymmdd.organization.adhoc.dotgov.txt file using the corresponding orgid.
- [ ] The following fields are populated:
  - Street address (this will have all the contents of the "orgstreet" source field. `Street address line 2` will be left blank.
  - city
  - state
  - zip
- [ ] Migration file created

- [ ] because this will be added in late in the game, a check should be added so that if this data is ALREADY set on a particular domain don't change it. (IE a user already went in and provided more accurate information)

### Additional Context:

(confirm this slack thread)
See [Slack Thread ](https://cisa-corp.slack.com/archives/C05BDEA3C11/p1698420433132029)for discussion.

For spreadsheet view of the data, see [this example google doc](https://docs.google.com/spreadsheets/d/1Wf2jrfVGj5yQCE7I66HweDQZokMTQ8vUWqMXh19u5TY/edit#gid=36490367).

### Data Details
NOTE: The data being migrated is sourced from a few different files, so some mapping is necessary. Refer to the existing example pulling data from the organization file. 

For each domain in `yyyymmdd.domainadditionaldatalink.adhoc.dotgov.txt`, the column orgid is a key into yyyymmdd.organization.adhoc.dotgov, which contains the detailed organization data.

- Source Organization data is as follows:
  - orgid (KEY)
  - orgname
  - orgstreet
  - orgcity
  - orgstate
  - orgzip
  - orgcountrycode

### Links to other issues

_No response_

zandercymatics:
## Ticket

Resolves #1243

## Changes

- Added script `load_organization_data.py` 
- Added documentation for script usage in `data_migration.md`
- Modified `send_domain_invitations.py` to include a minor bug fix when trying to .get() a non-existent TransitionDomain (broke test cases)
- Modified `extra_transition_domain_helper` to add the `OrganizationDataLoader` class which loads org data then adds it to TransitionDomains
- Modified the `ExtraTransitionDomain` class so that we can pass in a list of files we want to parse (rather than loading all 6). `transition_domain_arguments.py` was also modified for this reason
- Added a migration file with changes on TransitionDomain for the missing fields
- Modified `transition_domain.py` to include all relevant fields
- Modified `test_transition_domain_migrations` and added the `TestOrganizationMigration` class to test changes
- Modified `errors.py` to include some custom errors for the load_org script

## Context for reviewers
This ticket is meant to address the fact that during MVP data migration, we did not include Organization Information with that. 

Given that to do so requires modifying existing fields (which could corrupt user data), this script is designed to be cautious. For instance, we are utilizing bulk_update and only modifying a defined set of fields. Additionally, if a user has modified ANY of the fields we want to modify, we assume that we should not parse that specific field and it will skip it. 

There is a "preparing to parse" step which involves creating a list of proposed changes. If an error is encountered along the way, an exception is intentionally thrown. Rather than allowing for a manual override, it is better to modify the corrupt data then attempt to rerun the script.

This script runs pretty quickly, so if you already have all of the data on your sandbox, you can run it on the full dataset.

## Setup
Getgov-rjm will contain the code for this PR, and I will run the script there. This contains the entire dataset, and some data oddities to test the script. This is where you would go to inspect the related CSV.

When it comes to running this on your own, you will need to run the migration scripts to test this PR. To do so, refer to `data_migration.md`. The data you will use is located [here](https://drive.google.com/file/d/195DjlREITUBJB64tIcjchNBbHikY3dmk/view?usp=drive_link). For the purposes of this PR, keep everything inside of the migrationdata folder. This means that you can also do a `cf push` to move that data to your sandbox of choice, if you so desire.

After you've run those, simply run `./manage.py load_organization_data migrationFilepaths.json --debug`. 

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


src/registrar/management/commands/utility/extra_transition_domain_helper.py
```diff
@@ -751,6 +751,257 @@ def _infer_filename(self, regex: re.Pattern, matched_file_name, default_file_nam
...
+            )
+        else:
+            self.parse_logs.create_log_item(
+                file_type,
+                LogCode.WARNING,
+                f"Updated existing {var_name} to '{changed_value}' on {domain_name}",
+                domain_name,
+                not self.debug,
+            )
+
```
zandercymatics, line 1004: All of this is just boilerplate copied from the old code. Really this logic should be moved into a class if we wanted to use it again, but given where we are at - its not really as important for the PR. Its a little overkill for just parsing two files (as opposed to the many of the other script), but it saved time by using what already exists and it is rather efficient. 

This portion of the code may be confusing to follow without context to the last PR, let me know if it is.

src/registrar/management/commands/utility/extra_transition_domain_helper.py
```diff
@@ -775,52 +1026,49 @@ def __init__(self, options: TransitionDomainArguments):
...
+                (
+                    EnumFilenames.AUTHORITY_ADHOC,
+                    options.authority_adhoc_filename,
+                    AuthorityAdhoc,
+                    "authorityid",
+                ),
+            ]
 
-        self.file_data = self.populate_file_data(pattern_map_params)
+        self.file_data = self.populate_file_data(options.pattern_map_params)
```
zandercymatics, line 1071: This is a small modification to our "file organizer" (which basically stores data about each file, where its located, and how I want to parse it). Making this changes means that we can selectively choose which files to parse by passing in an option.

Again, this is overkill for two files but this approach is faster and reduces code duplication

src/registrar/management/commands/load_organization_data.py
```diff
@@ -0,0 +1,251 @@
...
+            "--domain_additional_filename",
+            help="Defines the filename for additional domain data",
+        )
+
+        # Serves as a organization_adhoc_filename override
+        parser.add_argument(
+            "--organization_adhoc_filename",
+            help="Defines the filename for domain type adhocs",
+        )
+
```
zandercymatics, line 49: Should these be left in? They are overrides for the filename but they aren't necessary to keep

src/registrar/management/commands/load_organization_data.py
```diff
@@ -0,0 +1,251 @@
...
+                        options[key] = value
+
+            except Exception as err:
+                logger.error(
+                    f"{TerminalColors.FAIL}"
+                    "There was an error loading "
+                    "the JSON responsible for providing filepaths."
+                    f"{TerminalColors.ENDC}"
+                )
+                raise err
```
zandercymatics, line 80: Most of this is copied over from the last PR. I am unsure if this needs more clarity.

src/registrar/management/commands/load_organization_data.py
```diff
@@ -0,0 +1,251 @@
...
+            domain_name__in=[item.domain_name for item in desired_objects],
+        )
+
+        # This indicates that some form of data corruption happened.
+        if len(desired_objects) != len(transition_domains):
+            raise LoadOrganizationError(code=LoadOrganizationErrorCodes.TRANSITION_DOMAINS_NOT_FOUND)
+
+        # Start with all DomainInformation objects
+        domain_informations = DomainInformation.objects.all()
+        domain_informations_dict = {di.domain.name: di for di in domain_informations}
```
zandercymatics, line 166: This is only used to determine if a given domain we are looping over actually exists. Should this be kept in?

Its worth noting that this script is already pretty fast. It can run on the entire dataset in under a minute, so performance isn't an issue.

src/registrar/management/commands/load_organization_data.py
```diff
@@ -0,0 +1,251 @@
...
+        # we can assume that they modified this information themselves - thus we should not update it.
+        domain_informations = domain_informations.filter(
+            domain__name__in=[td.domain_name for td in transition_domains],
+            address_line1__isnull=True,
+            city__isnull=True,
+            state_territory__isnull=True,
+            zipcode__isnull=True,
+        )
+
+        filtered_domain_informations_dict = {di.domain.name: di for di in domain_informations}
```
zandercymatics, line 180: Should be self-explanatory, but the reason this is stored in a dict is to increase eval speed. This changed the time it took to run the script from ~15 minutes down to >1 minute.

src/registrar/management/commands/load_organization_data.py
```diff
@@ -0,0 +1,251 @@
...
+        # Fetches all DomainInformations in one query.
+        # If any related organization fields have been updated,
+        # we can assume that they modified this information themselves - thus we should not update it.
+        domain_informations = domain_informations.filter(
+            domain__name__in=[td.domain_name for td in transition_domains],
+            address_line1__isnull=True,
+            city__isnull=True,
+            state_territory__isnull=True,
+            zipcode__isnull=True,
+        )
```
zandercymatics, line 178: This is assuming that if the user changed ANY of the given org fields, then we should just skip over it. Basically, it assumes that the later will probably change these fields later if they changed one.

The intent is to minimize accidental data corruption, but it doesn't need to be this way. Instead we can check each field individually and make the determination that way (update if null, leave alone if not). Thoughts on this?

src/registrar/management/commands/send_domain_invitations.py
```diff
@@ -152,6 +152,12 @@ def send_email(self, email_data):
...
-                this_transition_domain.save()
+                transition_domains = self.transition_domains.filter(username=this_email, domain_name=domain_name)
+                if len(transition_domains) == 1:
+                    this_transition_domain = transition_domains.get()
+                    this_transition_domain.email_sent = True
+                    this_transition_domain.save()
+                elif len(transition_domains) > 1:
+                    logger.error(f"Multiple TransitionDomains exist for {this_email}")
+                else:
+                    logger.error(f"No TransitionDomain exists for {this_email}")
```
zandercymatics, line 163: Fixes a bug where .get fails when it probably shouldn't. Just logs it instead. Included in this PR because of test case reasons, but can be taken out + test cases adjusted.

src/registrar/management/commands/utility/extra_transition_domain_helper.py
```diff
@@ -775,52 +963,47 @@ def __init__(self, options: TransitionDomainArguments):
...
+                    OrganizationAdhoc,
+                    "orgid",
+                ),
+                (
+                    EnumFilenames.AUTHORITY_ADHOC,
+                    options.authority_adhoc_filename,
+                    AuthorityAdhoc,
+                    "authorityid",
+                ),
+            ]
```
zandercymatics, line 1004: This is just a one line modification of the old code but get doesn't see it that way, lol. This basically just specifies what files to load from and how to parse them.

This wasn't an option you could pass in before, so this change just makes it so we don't have to load all 6 files in OrganizationDataLoader and instead we can just define the two we would want to parse.

src/registrar/management/commands/load_organization_data.py
```diff
@@ -0,0 +1,263 @@
...
+
+    def handle(self, migration_json_filename, **options):
+        """Process the objects in TransitionDomain."""
+
+        # === Parse JSON file === #
+        json_filepath = os.path.join(options["directory"], migration_json_filename)
+
+        # If a JSON was provided, use its values instead of defaults.
+        with open(json_filepath, "r") as jsonFile:
```
abroddrick, line 57: **non blocking comment**: if you know you have a section of that does a specific task, that 100 percent should be a function. This block reads a json so a function with something like that in it's name makes sense here instead.

src/registrar/management/commands/load_organization_data.py
```diff
@@ -0,0 +1,263 @@
...
+            # load JSON object as a dictionary
+            try:
+                data = json.load(jsonFile)
+
+                skipped_fields = ["domain_additional_filename", "organization_adhoc_filename"]
+                # Iterate over the data from the JSON file. Skip any unused values.
+                for key, value in data.items():
+                    if value is None or value.strip() == "":
+                        continue
```
abroddrick, line 66: **nonblocking comment: ** continue and break should be used only when there is no other logical option as they are less easy to follow when reading code. Instead could you just do the negative here. If value is not None and strippedValue!="" and options not in skipped_field:

src/registrar/management/commands/load_organization_data.py
```diff
@@ -0,0 +1,263 @@
...
+
+            except Exception as err:
+                logger.error(
+                    f"{TerminalColors.FAIL}"
+                    "There was an error loading "
+                    "the JSON responsible for providing filepaths."
+                    f"{TerminalColors.ENDC}"
+                )
+                raise err
+        # === End parse JSON file === #
```
abroddrick, line 81: **nonblocking comment** this is the other clue that a function is needed, if you have to denote where the logical code break ends

src/registrar/management/commands/load_organization_data.py
zandercymatics: Sure, not a problem. I did that in a few other places so I may need to change that too. Is this generally good practice in Python? It varies by language

src/registrar/management/commands/load_organization_data.py
zandercymatics: Totally reasonable, I was actually thinking that too - I just wasn't sure if it was overboard

src/registrar/utility/errors.py
```diff
@@ -57,6 +57,44 @@ def __str__(self):
         return f"{self.message}"
 
 
+class LoadOrganizationErrorCodes(IntEnum):
```
abroddrick, line 60: **blocking suggestion** I think this should be moved out of the usual utility, as the data migration code is logically separate from the purpose of the application and these error messages should NOT be used anywhere else, thus we don't want them in a global utility.

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -19,6 +19,233 @@
 from .common import less_console_noise
 
 
+class TestOrganizationMigration(TestCase):
+    def setUp(self):
+        """ """
```
abroddrick, line 24: **comment**:  empty comment string here should be filled in

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -19,6 +19,233 @@
...
+        self.test_domain_contact_filename = "test_domain_contacts.txt"
+        self.test_contact_filename = "test_contacts.txt"
+        self.test_domain_status_filename = "test_domain_statuses.txt"
+
+        # Files for parsing additional TransitionDomain data
+        self.test_agency_adhoc_filename = "test_agency_adhoc.txt"
+        self.test_authority_adhoc_filename = "test_authority_adhoc.txt"
+        self.test_domain_additional = "test_domain_additional.txt"
+        self.test_domain_types_adhoc = "test_domain_types_adhoc.txt"
+        self.test_escrow_domains_daily = "test_escrow_domains_daily"
```
abroddrick, line 39: **comment/non-blocking:** Given this script is just supposed to be a one off, this is red flag for me to see so much setup for a test that then requires other files to be set. I would think the script would just take one file only, as we are NOT going to re-run the data migration at all, so we do not need the other pieces of the migration scripts touched or expanded on.

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -19,6 +19,233 @@
...
+        Domain.objects.all().delete()
+        DomainInformation.objects.all().delete()
+        DomainInvitation.objects.all().delete()
+        TransitionDomain.objects.all().delete()
+
+        # Delete users
+        User.objects.all().delete()
+        UserDomainRole.objects.all().delete()
+
+    def run_load_domains(self):
```
abroddrick, line 54: **comment**: as a general rule our PR review checklist says "new functions have comments, these are new functions probably should have comments for their purpose.

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -19,6 +19,233 @@
...
+            return_value=True,
+        ):
+            call_command(
+                "load_transition_domain",
+                self.migration_json_filename,
+                directory=self.test_data_file_location,
+            )
+
+    def run_transfer_domains(self):
+        call_command("transfer_transition_domains_to_domains")
```
abroddrick, line 68: **comment/non-blocking** honestly this command call coveys its purpose more  then just "run transfer_domains". Some meaning gets lost here with hiding it in the function whose name doesn't describe the purpose as well. Comment though would solve this

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -82,6 +310,19 @@ def run_master_script(self):
                 disablePrompts=True,
             )
 
+    def run_load_organization_data(self):
```
abroddrick, line 313: why is this added to the test migration script section?

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -19,6 +19,233 @@
...
+        expected_missing_domain_invitations,
+    ):
+        """Does a diff between the transition_domain and the following tables:
+        domain, domain_information and the domain_invitation.
+        Verifies that the data loaded correctly."""
+
+        missing_domains = []
+        duplicate_domains = []
+        missing_domain_informations = []
+        missing_domain_invites = []
```
abroddrick, line 101: why are we checking domain invites? That should not have changed

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -19,6 +19,233 @@
...
+        for transition_domain in TransitionDomain.objects.all():  # DEBUG:
```
abroddrick, line 102: **Comment** You should never need a for loop in a unit test. If you provide a reasonable amount of test data then you should be able to double check that your expected output occurred. Here you do not actually check that all the items successfully ended up in the transition domain table to begin with.

abroddrick: side note, why is # DEBUG there?

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -19,6 +19,233 @@
...
+            transition_domain_name = transition_domain.domain_name
+            transition_domain_email = transition_domain.username
+
+            # Check Domain table
+            matching_domains = Domain.objects.filter(name=transition_domain_name)
+            # Check Domain Information table
+            matching_domain_informations = DomainInformation.objects.filter(domain__name=transition_domain_name)
+            # Check Domain Invitation table
+            matching_domain_invitations = DomainInvitation.objects.filter(
```
abroddrick, line 111: **comment** This doesn't seem needed for what you changed, right?

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -19,6 +19,233 @@
...
+        total_duplicate_domains = len(duplicate_domains)
+        total_missing_domain_informations = len(missing_domain_informations)
+        total_missing_domain_invitations = len(missing_domain_invites)
+
+        total_transition_domains = len(TransitionDomain.objects.all())
+        total_domains = len(Domain.objects.all())
+        total_domain_informations = len(DomainInformation.objects.all())
+        total_domain_invitations = len(DomainInvitation.objects.all())
+
+        print(
```
abroddrick, line 135: remove this, we don't keep prints in a unit test, it will print at a different time then using logger.

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -19,6 +19,233 @@
...
+            address_line="93001 Arizona Drive",
+            city="Columbus",
+            state_territory="Oh",
+            zipcode="43268",
+        )
+        expected_transition_domain.id = transition.id
+
+        self.assertEqual(transition, expected_transition_domain)
+
+    def test_load_organization_data_domain_information(self):
```
abroddrick, line 200: Comments on new functions

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -19,6 +19,233 @@
...
+        # Third, test that we didn't corrupt any data
+        expected_total_transition_domains = 9
+        expected_total_domains = 5
+        expected_total_domain_informations = 5
+        expected_total_domain_invitations = 8
+
+        expected_missing_domains = 0
+        expected_duplicate_domains = 0
+        expected_missing_domain_informations = 0
+        # we expect 1 missing invite from anomaly.gov (an injected error)
```
abroddrick, line 235: **comment/non-blocking** same question as before, why is this in this test at all

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -19,6 +19,233 @@
...
+        _domain = Domain.objects.filter(name="fakewebsite2.gov").get()
+        domain_information = DomainInformation.objects.filter(domain=_domain).get()
+
+        self.assertEqual(domain_information.address_line1, "93001 Arizona Drive")
+        self.assertEqual(domain_information.city, "Columbus")
+        self.assertEqual(domain_information.state_territory, "Oh")
+        self.assertEqual(domain_information.zipcode, "43268")
+
+    def test_load_organization_data_integrity(self):
+        """Validates data integrity with the load_org_data command"""
```
abroddrick, line 218: data integrity is very vague, maybe checks that all domain information and domain objects were created

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -19,6 +19,233 @@
...
+            expected_total_transition_domains,
+            expected_total_domains,
+            expected_total_domain_informations,
+            expected_total_domain_invitations,
+            expected_missing_domains,
+            expected_duplicate_domains,
+            expected_missing_domain_informations,
+            expected_missing_domain_invitations,
+        )
+
```
abroddrick, line 247: **blocking comment** The goal was to retroactively run organization data on a filled table loading into transition domains and then updating the domain information object. I would expect a test 1.) checking that transition domain was updated with organization data, assert organization data was updated +assert that other fields did NOT change. 2.) a test that organization fields (and only those fields) changed on the corresponding domain information objects. 3.) a test that if organization address was ALREADY changed (ie it exists in domain information already), then assert that the organization address in domain information is NOT updated.

While none of these were 100percent done, I can call #1 good you check everything on the transition domain for one of the domains. I am not seeing 2 and 3 here at all, feel free to correct me here but if you don't see these satisfied as well they need to be added in.

abroddrick (changes requested):
Still reviewing but started with the tests and errors. Please see the comment marked as blocking at the bottom of your test file and see the comment about moving the errors.py. I will continue to review the rest of the code as well

src/registrar/tests/test_transition_domain_migrations.py
zandercymatics: You are correct, those are not needed. Cleaned up that portion

src/registrar/management/commands/load_organization_data.py
```diff
@@ -0,0 +1,293 @@
...
+        # If any related organization fields have been updated,
+        # we can assume that they modified this information themselves - thus we should not update it.
+        domain_informations = DomainInformation.objects.filter(
+            domain__name__in=[td.domain_name for td in transition_domains],
+            address_line1__isnull=True,
+            city__isnull=True,
+            state_territory__isnull=True,
+            zipcode__isnull=True,
+        )
+        filtered_domain_informations_dict = {di.domain.name: di for di in domain_informations if di.domain is not None}
```
zandercymatics, line 181: This assumes that if any of those fields are modified, then a user likely had their hands on it - thus we shouldn't update.

This may be overly cautious, or it may not. It doesn't have to be this way, instead we can just do a check to not update fields that already have data in them. Thoughts?

src/registrar/management/commands/load_organization_data.py
```diff
@@ -0,0 +1,293 @@
...
+
+        # === Log results and return data === #
+        if len(self.domains_failed_to_update) > 0:
+            failed = [item.domain_name for item in self.domains_failed_to_update]
+            logger.error(
+                f"""{TerminalColors.FAIL}
+                Failed to update. An exception was encountered on the following Domains: {failed}
+                {TerminalColors.ENDC}"""
+            )
+            raise LoadOrganizationError(code=LoadOrganizationErrorCodes.UPDATE_DOMAIN_INFO_FAILED)
```
zandercymatics, line 195: To avoid any potential data corruption, this assumes that we should just make the script fail rather than proceed anyway

src/registrar/tests/test_transition_domain_migrations.py
zandercymatics: For 2), I assume you mean modifying the `test_load_organization_data_domain_information` test such that includes the other fields. Will make that change.

For 3), I agree with adding that. Makes sense to test that behavior.

src/registrar/tests/test_transition_domain_migrations.py
zandercymatics: Comment leftover from copy+paste

src/registrar/tests/test_transition_domain_migrations.py
zandercymatics: Comment leftover from copy + paste. Fixed.

zandercymatics:
@abroddrick Got those test case changes in

abroddrick (approved):
Ran it with the full data set and had 0 domain information objects without their address fields. Looks great. All comments are non blocking though please be sure to read/address ones where appropriate before merging

PR #1376, opened 2023-11-21
Issue #1338: Google Analytics CSP Bug

### LINKED ISSUE #1338: Bug: Google Analytics on manage.get.gov not sending data due to Content Security Policy issue
### Issue description

We're not receiving data in Google Analytics for manage.get.gov. The tag is on the pages in the right place. Is there some setting or code on manage.get.gov that could be blocking GA? 

image

The Chrome console tools say:

> Refused to load the script 'https://www.googletagmanager.com/gtag/js?id=G-PZ5QSP6QPL' because it violates the following Content Security Policy directive: "default-src 'self'". Note that 'script-src-elem' was not explicitly set, so 'default-src' is used as a fallback.
> 
> login/:9 Refused to execute inline script because it violates the following Content Security Policy directive: "default-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-UraOrhktOjKQSK/hGnJDVePveQUKN3Rqy/ukrAozrx0='), or a nonce ('nonce-...') is required to enable inline execution. Note also that 'script-src' was not explicitly set, so 'default-src' is used as a fallback.

So...does something in https://github.com/cisagov/manage.get.gov/blob/c457f862fc10a15fe5f313c846011ed513658899/src/registrar/config/settings.py need to change?

### Acceptance criteria

- [ ] Data can be viewed on the Google Analytics platform

### Additional context

- One recommended solution https://hershkoy.medium.com/resolving-google-analytics-4-content-security-policy-issues-9a01e18952d2
- [Stack Overflow convo](https://stackoverflow.com/questions/75940522/google-analytics-issues-refusing-to-connect-because-it-violates-the-following-c)

### Links to other issues

#1329

CocoByte:
## Ticket 1338

Resolves #1338 

## Changes

- Updated settings.py to whitelist google analytics scripts
- Updated base.html with Nonce for inline script (whitelists the inline script)

## Setup

Set IS_PRODUCTION to True in your .env file

### Verify Locally:
Docker-compose up
Navigate to manage.get.gov, open inspect and verify the following;
- [ ] google analytics scripts have loaded in the header
- [ ] there are no CSP error messages in the console 

### Verify in Sandbox and Google Analytics:
(Note: this will require some setup that we do not want to be committed back in to the code repo)

#### 1 - Create a Google Analytics stream that links up to your sandbox
- If you don't already have an analytics account, go to https://analytics.google.com/analytics/web/#/provision/create and create a test account (I just used my gmail and added random info for company details).  I recommend unselecting things like data sharing (bottom of the screen in step 1)
image
- When you get to step 5 - "Data collection" select "Web"
Pasted Graphic 4
- Put in your sandbox info (see example below) and click "Create Stream"
Pasted Graphic 5

#### 2 - Copy-paste the stream ID over the one we use in base.html
When your stream first gets created, it will not be active.  This is because we need to configure the code with the ID of your stream.  Copy the Measurement ID shown in your stream details...
Pasted Graphic 6
- Paste this ID over the ones In base.html
Pasted Graphic 7

#### 4 - Change your sandbox .env to set IS_PRODUCTION=True (or just remove the IF statement in the code before pushing -- not as clean, but we do not want to commit this code anyways. NOTE: do not comment out the IF statement. That will do nothing.  You literally have to delete (or move) the if statement to free the code.)

image

#### 5 - You should now be able to see some google analytic metrics.  Verify the following:
- [ ] The header loads just like you observed in Local

image

- [ ] Google analytics shows data collection is active (may need to refresh the analytics page) 
- [ ] Clicking around in your sandbox generates some activity (eg. a page view)

image

- [X] Met the acceptance criteria, or will meet them in a subsequent PR
- [X] Added at least 2 developers as PR reviewers (only 1 will need to approve)
- [X] Messaged on Slack or in standup to notify the team that a PR is ready for review

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### DIFF
--- a/src/registrar/config/settings.py
+++ b/src/registrar/config/settings.py
@@ -304,9 +304,8 @@ CSP_DEFAULT_SRC = allowed_sources
 # explicitly set
 CSP_FRAME_ANCESTORS = allowed_sources
 CSP_FORM_ACTION = allowed_sources
-CSP_SCRIPT_SRC_ELEM = allowed_sources_scripts
-
-CSP_CONNECT_SRC = allowed_sources_scripts
+CSP_SCRIPT_SRC_ELEM = ["'self'", "https://www.googletagmanager.com/"]
+CSP_CONNECT_SRC = ["'self'", "https://www.google-analytics.com/"]
 CSP_INCLUDE_NONCE_IN = ["script-src-elem"]
 
 # Cross-Origin Resource Sharing (CORS) configuration
--- a/src/registrar/config/settings.py
+++ b/src/registrar/config/settings.py
@@ -297,12 +297,26 @@ SERVER_EMAIL = "root@get.gov"
 
 # Content-Security-Policy configuration
 # this can be restrictive because we have few external scripts
-allowed_sources = ("'self'",)
+allowed_sources = (
+    "'self'"
+)
+allowed_sources_scripts = [
+    "'self'",
+    "https://www.googletagmanager.com/",
+    "https://www.google-analytics.com/"
+]
 CSP_DEFAULT_SRC = allowed_sources
-# Most things fall back to default-src, but these two do not and should be
+# Most things fall back to default-src, but the following do not and should be
 # explicitly set
 CSP_FRAME_ANCESTORS = allowed_sources
 CSP_FORM_ACTION = allowed_sources
+CSP_SCRIPT_SRC_ELEM = allowed_sources_scripts
+CSP_SCRIPT_SRC = allowed_sources_scripts
+CSP_CONNECT_SRC = allowed_sources_scripts
+CSP_INCLUDE_NONCE_IN = [
+    'script-src',
+    'script-src-elem'
+    ]
 
 # Cross-Origin Resource Sharing (CORS) configuration
 # Sets clients that allow access control to manage.get.gov
--- a/src/registrar/config/settings.py
+++ b/src/registrar/config/settings.py
@@ -297,14 +297,8 @@ SERVER_EMAIL = "root@get.gov"
 
 # Content-Security-Policy configuration
 # this can be restrictive because we have few external scripts
-allowed_sources = (
-    "'self'"
-)
-allowed_sources_scripts = [
-    "'self'",
-    "https://www.googletagmanager.com/",
-    "https://www.google-analytics.com/"
-]
+allowed_sources = "'self'"
+allowed_sources_scripts = ["'self'", "https://www.googletagmanager.com/", "https://www.google-analytics.com/"]
 CSP_DEFAULT_SRC = allowed_sources
 # Most things fall back to default-src, but the following do not and should be
 # explicitly set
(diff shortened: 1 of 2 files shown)


abroddrick:
@CocoByte how do we test that google analytics is actually working locally? I imagine there is an interface that we need to request access to?

therealslimhsiehdy (commented):
Thank you for letting me know that instead of commenting out the if statement in base.html, guess one has to delete those lines out before pushing to staging 🤣 I was able to see the google tag both working + able to see it active. 

However I don't know enough about Google Analytics security/context so I will let someone else with more knowledge approve. Otherwise this looks great!

Screenshot 2023-11-21 at 11 08 49 AM

Screenshot 2023-11-21 at 11 10 18 AM

abroddrick:
> However I don't know enough about Google Analytics security/context so I will let someone else with more knowledge approve. Otherwise this looks great!
> 

@therealslimhsiehdy even if you don't know alot about it, if you tested it and looked at the code and found nothing wrong that is 100 percent grounds to just go ahead and approve. The developer can wait for another review if they see fit. Please click the approve button, if you did the review steps and have nothing that is blocking

src/registrar/config/settings.py
```diff
@@ -297,12 +297,16 @@
 
 # Content-Security-Policy configuration
 # this can be restrictive because we have few external scripts
-allowed_sources = ("'self'",)
+allowed_sources = "'self'"
+allowed_sources_scripts = ["'self'", "https://www.googletagmanager.com/", "https://www.google-analytics.com/"]
```
neilmb, line 301: **suggestion (non-blocking)**: We never use this `allowed_sources_scripts` variable now, so I think you can remove this line.

src/registrar/config/settings.py
```diff
@@ -297,12 +297,16 @@
...
 CSP_DEFAULT_SRC = allowed_sources
-# Most things fall back to default-src, but these two do not and should be
+# Most things fall back to default-src, but the following do not and should be
 # explicitly set
 CSP_FRAME_ANCESTORS = allowed_sources
 CSP_FORM_ACTION = allowed_sources
+CSP_SCRIPT_SRC_ELEM = ["'self'", "https://www.googletagmanager.com/"]
```
neilmb, line 307: **suggestion (non-blocking)**: These two items seem like they are under the previous comment, but that doesn't describe why they are being set. Can we put in another comment to describe why they are here?
```suggestion

# Google analytics requires that we relax our otherwise
# strict CSP by allowing scripts to run from their domain
# and inline with a nonce, as well as allowing connections back to their domain
CSP_SCRIPT_SRC_ELEM = ["'self'", "https://www.googletagmanager.com/"]
```

abroddrick (approved):
Seems like magic but appears to work, would love to learn more about how this works! LGTM

src/registrar/config/settings.py
neilmb, line 301:
```suggestion
```

src/registrar/config/settings.py
```diff
@@ -297,12 +297,16 @@
 
 # Content-Security-Policy configuration
 # this can be restrictive because we have few external scripts
-allowed_sources = ("'self'",)
+allowed_sources = "'self'"
```
neilmb, line 300: **suggestion (blocking)**: It is surprising, but allowed_sources actually needs to be a one-element tuple here instead of a single string. The parentheses and the comma are required.
```suggestion
allowed_sources = ("'self'",)
```

neilmb (commented):
LGTM, two small suggestions and one that needs fixing.

therealslimhsiehdy (approved):

src/registrar/config/settings.py
CocoByte: oh yes, good catch

src/registrar/config/settings.py
CocoByte: Oh wow...I actually don't remember removing the tuple enclosure...I wonder if linting did.....Thanks for the catch

Issue #1388, opened 2023-11-22
Change "DS Data" to "DS data" everywhere


michelle-rago:
### Issue description

"DS Data" should be "DS data" in the HTML page title, navigation, heading, error messages, and body text.

<img width="1042" alt="DS-data" src="https://github.com/cisagov/manage.get.gov/assets/60157596/e9fc3cb2-3d9f-4860-95c1-958d6fcca34d">

### Acceptance criteria

- [ ] "DS Data" is "DS data" (sentence case) in all user-facing content

michelle-rago:
cc @PaulKuykendall for appropriate ticket tagging

PR #1389, opened 2023-11-22
Changed DNS sidebar nav / Put DNS options in a list on DNS landing page

michelle-rago:
## Changes

- Changed "Name servers" to "DNS name servers"
- Changed "DS Data" to "DS data" (related to https://github.com/cisagov/manage.get.gov/issues/1388)
- Bulleted DNS options on DNS landing page

### DIFF
--- a/src/registrar/templates/domain_dns.html
+++ b/src/registrar/templates/domain_dns.html
@@ -12,9 +12,11 @@
   <p>You can enter your name servers, as well as other DNS-related information, in the following sections:</p>
 
   {% url 'domain-dns-nameservers' pk=domain.id as url %}
-  <p><a href="{{ url }}">DNS name servers</a></p>
+<ul>  
+<li><a href="{{ url }}">DNS name servers</a></li>
 
   {% url 'domain-dns-dnssec' pk=domain.id as url %}
-  <p><a href="{{ url }}">DNSSEC</a></p>
+  <li><a href="{{ url }}">DNSSEC</a></li>
+</ul>
 
 {% endblock %}  {# domain_content #}
--- a/src/registrar/templates/domain_dns.html
+++ b/src/registrar/templates/domain_dns.html
@@ -13,7 +13,7 @@
 
   {% url 'domain-dns-nameservers' pk=domain.id as url %}
 <ul>  
-<li><a href="{{ url }}">DNS name servers</a></li>
+<li><a href="{{ url }}">Name servers</a></li>
 
   {% url 'domain-dns-dnssec' pk=domain.id as url %}
   <li><a href="{{ url }}">DNSSEC</a></li>
--- a/src/registrar/templates/domain_sidebar.html
+++ b/src/registrar/templates/domain_sidebar.html
@@ -24,7 +24,7 @@
             <a href="{{ url }}"
               {% if request.path == url %}class="usa-current"{% endif %}
             >
-                DNS name servers
+                Name servers
             </a>
           </li>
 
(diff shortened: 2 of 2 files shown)


h-m-f-t:
Can you say a bit more about this change, @michelle-rago? Doesn't it make the nav:
* DNS
  * DNS name servers

michelle-rago:
> Can you say a bit more about this change, @michelle-rago? Doesn't it make the nav:
> 
> * DNS
>   
>   * DNS name servers

It does. We'd like the navigation to be consistent with the headings in the main content area. @Katherine-Osos and I discussed this in our [content review spreadsheet](https://docs.google.com/spreadsheets/d/18wwmEioSr6BU9Y6G6ihqzS7P1OaE65Ifoipg9-Gxcjo/edit?disco=AAABArd9MhU).

Katherine-Osos:
@h-m-f-t I agree with Michelle that the primary heading should match the navigation. An alternative would be to remove "DNS" in the page heading, but that didn't seem right to us ...

@gabydisarli since you worked on this section, do you have an opinion on the name of the page?

neilmb (approved):

abroddrick (commented):
LGTM

abroddrick (approved):
Meant to click approve not comment last time

Katherine-Osos (approved):

gabydisarli:
> @h-m-f-t I agree with Michelle that the primary heading should match the navigation. An alternative would be to remove "DNS" in the page heading, but that didn't seem right to us ...
> 
> @gabydisarli since you worked on this section, do you have an opinion on the name of the page?

I understand your reasoning on primary headings matching navigation, but I think in this case, because this is a child page of the parent DNS page it's implied that this page is part of the DNS section, sort of like DNS: Name servers in a way. If y'all don't agree with that characterization, maybe we change the page heading to just "Name servers", then? I agree with @h-m-f-t that saying it twice in the nav is awkward.

h-m-f-t:
+1 to Gaby's comment. Though I could live with this header/nav inconsistency, I think the better step is making the page heading "Name servers" instead of making the nav read "DNS name servers".

Katherine-Osos:
@michelle-rago Just following up on this update. Are you ok with making the page heading "Name servers"?

michelle-rago:
@Katherine-Osos I still think it should be "DNS name servers." I wouldn't expect a user to know what section they're in and I think "DNS name servers" makes this as explicit as possible. Users could be looking for "DNS" or "name sever" as the thing that they want. 

We are using DNS to mean Domain Name System so we're not repeating the term "name server." I get that some folks might think that though. @gabydisarli @h-m-f-t Do you think "DNS name servers" runs the risk of confusing anyone or do you think it just looks repetitive (or like we don't know what we're talking about)?

h-m-f-t:
>  I wouldn't expect a user to know what section they're in and I think "DNS name servers" makes this as explicit as possible

I think the storytelling we do at https://manage.get.gov/domain/[number]/dns and the emails we'll send gives users the context they need to know what to look for. The URL changes to /[number]/dns/nameserver, too. We could even link directly to this section to minimize confusion.

> Do you think "DNS name servers" runs the risk of confusing anyone or do you think it just looks repetitive (or like we don't know what we're talking about)?

I don't know that it will confuse anyone, but I agree it looks repetitive in the nav and feels unnecessarily redundant.

Even so, I defer to @michelle-rago judgment. You make the call. The whole DNS section will change substantially in the next ~6 months and I strongly suspect we won't have a standalone "name servers" section (at least as a default).

michelle-rago:
Thanks for going on this journey, folks. I changed it to "Name servers" sans the "DNS." I realized that's also more consistent with our intro text, in addition to the justification y'all provided.

Katherine-Osos (approved):

PR #1420, opened 2023-11-30
#1378 Display error message when availability API fails

### LINKED ISSUE #1378: Availability endpoint on the request form not working
### Current Behavior

Entering any domain into the check domain availability box on the application workflow fails. 

### Expected Behavior

Domains that are available should result in a the green success message indicating that this domain is available to request.

### Steps to Reproduce

1. in the request form go to register/dotgov_domain/
2. enter a domain name that should be available
3. see that it shows red

### Environment

_No response_

### Additional Context

image

### Issue Links

_No response_

erinysong:
## Ticket

Resolves #1378 

## Changes

- Displays an error message when availability call fails for any reason and specifies the API failure is due to a connection issue, rather than just saying the domain is unavailable which is misleading.
image

## Context for reviewers

There are some cases in which the domain availability check will intermittently fail due to connection pool errors, leading to false information that a domain is unavailable. In this case, we would want to inform the user that the issue is coming from the EPP client's end rather than the domain that they are inputting.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [x] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


src/api/tests/test_available.py
```diff
@@ -101,15 +101,16 @@ def test_available_domain_dotgov(self):
         self.assertTrue(json.loads(response.content)["available"])
 
     def test_error_handling(self):
-        """Calling with bad strings raises an error."""
+        """Calling with bad strings returns error (error verifying) or unavailable (invalid domain)."""
```
abroddrick, line 104: I think the error message portion should be a different test since it's a output of the api we are checking

src/api/views.py
```diff
@@ -72,8 +73,8 @@ def check_domain_available(domain):
         else:
             # domain search string doesn't end with .gov, add it on here
             return Domain.available(domain + ".gov")
-    except Exception:
-        return False
+    except Exception as err:
```
abroddrick, line 76: I think since you are just raising the error here you can just remove the try/except block entirely

src/api/tests/test_available.py
```diff
@@ -101,15 +101,16 @@ def test_available_domain_dotgov(self):
...
         bad_string = "blah!;"
         request = self.factory.get(API_BASE_PATH + bad_string)
         request.user = self.user
         response = available(request, domain=bad_string)
         self.assertFalse(json.loads(response.content)["available"])
-        # domain set to raise error returns false for availability
-        error_domain_available = available(request, "errordomain.gov")
-        self.assertFalse(json.loads(error_domain_available.content)["available"])
+        # domain set to raise error returns false for availability and error message
+        error_domain_response = available(request, domain="errordomain.gov")
```
abroddrick, line 111: This should be it's own test since it's a different output we are checking

src/api/views.py
```diff
@@ -30,7 +30,8 @@
     ),
     "invalid": "Enter a domain using only letters, numbers, or hyphens (though we don't recommend using hyphens).",
     "success": "That domain is available!",
-    "error": "Error finding domain availability.",
+    "error": "Error finding domain availability. Please wait a few minutes and try again.\
```
abroddrick, line 33: **blocking change**: The slack thread with Katherine mentioned using our already-designer approved Registry connection error message (a few lines up from the message I sent you earlier). It is in registrar/utility/errors.py and it is the one associated with Cannot_contact_registry. If you can reference that one directly that would be ideal, otherwise I would copy and paste it here instead.

abroddrick (changes requested):
Couple minor things, and the big thing is just using the error message that Katherine requested.

src/api/views.py
abroddrick: That being said, @Katherine-Osos would erin's suggestion for a registry connection error above work as a replacement or would you prefer the one already suggested in my message "We’re experiencing a system connection error. Please wait a few minutes
and try again. If you continue to receive this error after a few tries,
contact help@get.gov"

Katherine-Osos: Erin's suggestion works!

Katherine-Osos: I was going to say Erin's suggestion works ... but I realized this same message should appear when searching for a domain in the request flow (right?). In that case, we should offer the contact email, so let's use the existing message we use elsewhere: 

"We’re experiencing a system connection error. Please wait a few minutes and try again. If you continue to receive this error after a few tries, contact [help@get.gov](mailto:help@get.gov)"

erinysong: Updated the error message to make the error message to be the same as the generic registry error by pulling from the constants set in our GenericError class - the only concern I have is that it does this by referencing a helper function _error_mapping in GenericError and I was wondering if this was good practice. It's not exactly a private function and I was wondering if we should make any code format changes that would be good for this @abroddrick

Katherine-Osos (approved):

src/api/tests/test_available.py
erinysong: Updated to split bad string inputs and error handling into separate test cases

src/api/views.py
abroddrick: I totally see what you're saying, normally you wouldn't directly access something with a leading _. I would say just add a function in GenericError  getRegistryConnection error and have it just be one line that has a return with the _error_mapping[...]. That would probably be a bit easier to read anyways

src/registrar/models/utility/domain_helper.py
```diff
@@ -29,19 +29,19 @@ def validate(cls, domain: str | None, blank_ok=False) -> str:
         if not isinstance(domain, str):
             raise ValueError("Domain name must be a string")
         domain = domain.lower().strip()
-        if domain == "":
-            if blank_ok:
-                return domain
-            else:
-                raise errors.BlankValueError()
+        if domain == "" and not blank_ok:
```
erinysong, line 32: ended up making this change to satisfy a lint requirement but let me know if we want to change it back and I can add an exception to this function on the linter

src/api/views.py
erinysong: Sounds good! Just updated to make a class function that gives the correct error message for the corresponding code

src/registrar/models/utility/domain_helper.py
```diff
@@ -29,19 +29,19 @@ def validate(cls, domain: str | None, blank_ok=False) -> str:
...
         if "." in domain:
             raise errors.ExtraDotsError()
         if not DomainHelper.string_could_be_domain(domain + ".gov"):
             raise ValueError()
-        if not check_domain_available(domain):
-            raise errors.DomainUnavailableError()
+        try:
+            if not check_domain_available(domain):
+                raise errors.DomainUnavailableError()
+        except Exception:
```
abroddrick, line 43: **non-blocking Suggestion:** always tried to avoid using a generic Exception. This is because you (and others maintaining the code) should know what type of errors can result from a function. Here it should be a RegistryError correct? In which case, I would catch a RegistryError instead. Double check that is the correct error

src/api/views.py
```diff
@@ -66,14 +67,11 @@ def check_domain_available(domain):
...
-        else:
-            # domain search string doesn't end with .gov, add it on here
-            return Domain.available(domain + ".gov")
-    except Exception:
-        return False
+    if domain.endswith(".gov"):
+        return Domain.available(domain)
+    else:
+        # domain search string doesn't end with .gov, add it on here
+        return Domain.available(domain + ".gov")
```
zandercymatics, line 74: Nice, I like the refactor here

src/api/views.py
```diff
@@ -66,14 +67,11 @@ def check_domain_available(domain):
     a match.
     """
```
abroddrick, line 68: **change before merging** Now that it throws an exception, the comment should indicate "throws RegistryError" to indicate that it has that potential

src/registrar/utility/errors.py
```diff
@@ -56,6 +60,10 @@ def __init__(self, *args, code=None, **kwargs):
     def __str__(self):
         return f"{self.message}"
 
+    @classmethod
+    def get_error_message(self, code=None):
```
abroddrick, line 64: nice!

abroddrick (approved):
LGTM, see only a couple comments. Only the comment updated is a must have

src/registrar/models/utility/domain_helper.py
```diff
@@ -29,19 +29,19 @@ def validate(cls, domain: str | None, blank_ok=False) -> str:
...
+            raise errors.RegistrySystemError()
```
zandercymatics, line 44:
```suggestion
        except Exception as err:
            raise errors.RegistrySystemError() from err
```
Typically when we re-raise errors like this, we want to add this "from err" syntax. The reason for this is because when you just raise an error like this normally, you lose information about the exception that caused it. For instance, if you had a syntax error on line 42, it wouldn't return that a syntax error occurred in the stack trace, but just that a registry system error was encountered. That `from` basically maintains that data and wraps it up with the error you throw

src/registrar/models/utility/domain_helper.py
```diff
@@ -29,19 +29,19 @@ def validate(cls, domain: str | None, blank_ok=False) -> str:
...
             domain = domain[:-4]
         if "." in domain:
             raise errors.ExtraDotsError()
         if not DomainHelper.string_could_be_domain(domain + ".gov"):
             raise ValueError()
-        if not check_domain_available(domain):
-            raise errors.DomainUnavailableError()
+        try:
+            if not check_domain_available(domain):
+                raise errors.DomainUnavailableError()
```
zandercymatics, line 42: (Q) was line 41 throwing an exception?

zandercymatics (commented):
Nice work! Just one small blocking comment, but nothing too major. Still testing out your branch, so I'll create a follow-up PR once I am finished with that

src/registrar/models/utility/domain_helper.py
erinysong: I believe so! I think lines 41-42 were the original code and we put it in a try catch since we're now throwing a RegistryError when the availability API fails

src/registrar/models/utility/domain_helper.py
erinysong: ah thank you for letting me know and did not know that about specifying exceptions through as err... from err! Appreciate you taking the time to write out an explanation on that and just updated with the suggested fix

src/registrar/models/utility/domain_helper.py
erinysong: Ah got it - I ended up catching RegistryError and also saving using as/from err - thank you both for the explanations and suggestions!

Issue #1435, opened 2023-12-04
Create a method that decides whether a Login.gov user should be identity verfied or not


neilmb:
### Issue description

In #1433 we are building a two-tiered log in procedure where some users can be logged in without needing to be identity verified at Login.gov, but the rest of the users will need to be logged in with Login's identity verification process. In order for that OIDC work to function, we need to design a method say a `classmethod` like `User.needs_identity_verification(email, uuid)` in our `src/registrar/` code base that can make that determination.

Our best description right now of the individuals who don't need to be identity verified are: (1) any existing user who is a domain manager of a domain (that is, they have an entry in `UserDomainRole` for their `User`) OR (2) any new incoming user who is a domain manager for one of the domains that we inputted from Verisign (that is, their email address appears in the `username` field of a `TransitionDomain`) OR (3) any new incoming user who is being invited to be a domain manager (that is, their email address is in `DomainInvitation` for an invitation that is not yet "retrieved").

### Acceptance criteria

- [x] There is a single method that our OIDC library can call to find out whether it should require identity verification
- [x] The new method is simple, clear, and well-documented so that we can change its behavior from a policy perspective
- [x] The new method has unit tests that express the positive and negative cases of all three busines-cases listed above

### Links to other issues

OIDC implementation in #1433, research and motivation in #1414

neilmb:
This comment belongs here:

Just another implementation note that we probably have to keep around a table of email addresses to check on "needs identity verification" status rather than using a property in the User table. Or both. Because most of the legacy domain managers from the previous system have not logged in and we won't have User entries for them until they do. So we'll just need to keep a table of email addresses and check whether a new user who is logging in is on that list? Or maybe we can just use that table of email addresses to set the needs_identity_verification property on their newly created User?

But however we do it, we're going to need a new table to store these "special" email addresses.

Issue #1433, opened 2023-12-04
Modify our OIDC library to allow differentiating identity-verification levels for different users


neilmb:
### Issue description

As part of #1414, we want to make it possible to send users to the Login.gov OIDC provider with a value of `http://idmanagement.gov/ns/assurance/ial/1` for `acr_values` (see https://github.com/cisagov/manage.get.gov/blob/c569db24f35a4f088c0bd177bc84606e271f7b3a/src/djangooidc/oidc.py#L103 and https://github.com/cisagov/manage.get.gov/blob/c569db24f35a4f088c0bd177bc84606e271f7b3a/src/registrar/config/settings.py#L521) BUT then for certain users, we want to NOT log them in at that point, instead we would want to redirect them BACK to Login.gov with `acr_values` of `http://idmanagement.gov/ns/assurance/ial/2` to require those users to be identity verified in order to log in.

Our `src/djangooidc` library will need some changes to be able to handle this pattern. We will need to add an attribute on our `User` model (`needs_identity_verification`?) that we check from inside `djangooidc` to determine which users need this treatment. We will also need to modify the `behaviour` configuration inside of `djangooidc` to allow for configuring two distinct settings for `acr_value`.

Because this is an invasive change to a crucial user-facing experience, we want to be sure that it is working as we expect. We should make unit tests for the oidc login flow (some unit tests are here https://github.com/cisagov/manage.get.gov/tree/c569db24f35a4f088c0bd177bc84606e271f7b3a/src/djangooidc/tests) that completely mock out the requests to Login.gov, but otherwise give us a fairly complete end-to-end example of logging in. The actual redirect request that we could mock out is at https://github.com/cisagov/manage.get.gov/blob/c569db24f35a4f088c0bd177bc84606e271f7b3a/src/djangooidc/oidc.py#L137.

### Acceptance criteria

- [ ] Some users can log in WITHOUT being identity-verified with Login.gov
- [ ] Some other users cannot log in WIHTOUT being identity-verified with Login.gov (this should be the default for "unrecognized" users).
- [ ] Unit testing for both aspects of this flow is in place.

### Additional context

Login's engineering staff is available on #login-partner-support Slack channel or via Zendesk ticket https://zendesk.login.gov/ for consulting about this. They have been very helpful in the past, so reaching out earlier rather than later would be best. https://github.com/18F/identity-oidc-sinatra/ has a working example of this pattern of two consecutive redirects with different `acr_value` in a Ruby web app.

[...]

PR #1441, opened 2023-12-05
Switch to allowing non-identity-verified logins

neilmb:
In advance of #1433, for now we want to allow people to log in using Login.gov WITHOUT being identity verified. This simple change makes that switch. People logging in to the system will need to have Login.gov accounts where they can verify control of the email address used, but they will not need to be personally identity-verified using their SSN/ID/phone/etc.

We anticipate that when #1433 and #1435 are ready, that this will be superseded, but for now this is right for our current group of domain managers.

@h-m-f-t @amcelya @michelle-rago

### DIFF
--- a/src/registrar/config/settings.py
+++ b/src/registrar/config/settings.py
@@ -518,7 +518,7 @@ OIDC_PROVIDERS = {
             "response_type": "code",
             "scope": ["email", "profile:name", "phone"],
             "user_info_request": ["email", "first_name", "last_name", "phone"],
-            "acr_value": "http://idmanagement.gov/ns/assurance/ial/2",
+            "acr_value": "http://idmanagement.gov/ns/assurance/ial/1",
         },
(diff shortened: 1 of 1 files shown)


therealslimhsiehdy (commented):
**Question:** To confirm, `2` meant that we would have to go through login.gov and going to `1` means we will not have to do identity proofing, correct? If so, is it same to assume that the link is like a "flag" set up?

Also! Is there a way to possibly test this?

neilmb:
Good question @therealslimhsiehdy! Everything still goes through Login.gov. The current "2" indicates to Login that each account used to log in needs to be "identity verified" on their side. The proposed "1" indicates to Login that the accounts just need to have control of the email address that they use.  See https://developers.login.gov/oidc/authorization/ under "Type of Service Level".

therealslimhsiehdy (approved):

PR #1445, opened 2023-12-06
Ticket #1444: Make transfer_transition_domains_to_domains work incrementally

### LINKED ISSUE #1444: Make our transfer_transition_domains_to_domains script able to work incrementally
### Issue description

We are wanting to use our data migration pathway a second time to add a small number of additional domains, but it was not designed originally to be used A SECOND TIME after the initial migration. In particular, the transfer script overwrites the domain's status, creation date, and expiration date for every domain in the `TransitionDomain` table, and if we use the second half to migrate `DomainInformation` then it would also overwrite a domain's contact information. For domains that have been managed since our initial launch, this overwriting is not acceptable.

To be able to use the transfer script incrementally, we should add a `transferred` boolean field to the `TransitionDomain` object with a default value of `False`, but in our migration, all of the existing `TransitionDomain` objects should be marked as `transferred=True`.  Then, in the transfer command, we should replace the loop over `TransitionDomain.objects.all()` with a loop over `TransitionDomain.objects.filter(transferred=False)` and at the end of each loop iteration we should mark it as `transition_domain.transferred = True`.

### Acceptance criteria

- [ ] We can use `transfer_transition_domains_to_domains` twice without the second run changing any properties related to the domains transferred on the first run.

### Additional context

This emerged as a possible issue when working towards #1425. It is also complicated because we might use the information in `TransitionDomain` as part of #1435 for reference to which email addresses don't need to be identity verified.

### Links to other issues

_No response_

zandercymatics:
## Ticket

Resolves #1444

## Changes

- Add a "processed" boolean flag on TransitionDomains
- Change the transfer_transition_domains_to_domains script such that it only edits domains with the flag processed set to True, and have it set that flag to True after it finishes execution
- Test cases

## Context for reviewers
This ticket concerns the transfer_transition_domains_to_domains command. Previously, rerunning on stable this would be destructive in that it would overwrite existing data. This change fixes that by adding a boolean flag to do so. 

## Setup
**Testing locally**
1. Grab the latest [migration data](https://drive.google.com/file/d/195DjlREITUBJB64tIcjchNBbHikY3dmk/view?usp=drive_link) and unzip it into migrationdata/
2. Run the load_transition_domains command with --limitParse set to 10
3. Check /admin and verify that all TransitionDomains have the "processed" flag set to False
4. Run the transfer_transition_domains_to_domains command
5. Check /admin and verify that all TransitionDomains have the "processed" flag set to True
6. From the added data choose a Domain object to modify. Change the expiration_date field. Write this down somewhere.
7. From the added data choose a DomainInformation object to modify. Change the "organization_name" field. Write this down somewhere.
8. Run the load_transition_domains command with --limitParse set to 20
9. Check /admin and verify that 10 TransitionDomains have the "processed" flag set to False, and the other 10 are set to True.
10. Run the transfer_transition_domains_to_domains command
11. Check /admin and verify that all TransitionDomains have the "processed" flag set to True
12. Check your modified domain, and verify that your data hasn't been corrupted by this process.
13. Continue steps 8-9, but with the [twenty_eight file](https://drive.google.com/file/d/1TnJ6CUjRBwODa3w8_ltU_Yjc-KGiudOe/view?usp=drive_link)

**Testing on sandboxes**
This PR has a migration change, but given the nature of this ticket, we actually want to upload migration data _before_ we do a cf push. This is because we want to test if all existing domains are given the "processed" flag.
1. Grab the latest [migration data](https://drive.google.com/file/d/195DjlREITUBJB64tIcjchNBbHikY3dmk/view?usp=drive_link)
2. Follow these [instructions](https://github.com/cisagov/manage.get.gov/blob/main/docs/operations/data_migration.md) for uploading files to our sandboxes. Run the migration scripts as specified, though do not send out emails. You can either use the entire dataset or a subsection of it.
3. After that is done, verify that the data has been loaded correctly on your sandbox. Then, do a cf push and load this branch.
4. SSH into your branch, run `tmp/lifecycle/shell`, then run `./manage.py migrate`
5. Navigate to `Transition Domains` in /admin. All existing TransitionDomains should have the flag "processed"
6. Choose a `Domain` object to modify. Modify the expiration date. Write this down.
7. Choose a `DomainInformation` object to modify. Modify the organization_name. Write this down.
8. Following the upload instructions from before, upload the [twenty_eight file](https://drive.google.com/file/d/1TnJ6CUjRBwODa3w8_ltU_Yjc-KGiudOe/view?usp=drive_link). Rerun the `load_transition_domains` script, then run the  
`transfer_transition_domains_to_domains` script.
9. In the script output, verify that the transfer_transition_domains script didn't run for every TransitionDomain, only those specified (should be no more than 28). All pre-existing domains should be unmodified.
10. View the Domain and DomainInformation objects that you modified. These should be identical to how they were before.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


src/registrar/management/commands/load_transition_domain.py
```diff
@@ -548,7 +548,9 @@ def handle(  # noqa: C901
                                 )
                                 existing_entry.status = new_entry_status
                             existing_entry.email_sent = new_entry_emailSent
+                            existing_entry.processed = True
```
abroddrick, line 551: **blocking question** @zandercymatics this looks like for each transition domain that already exists we still risk updating the status instead of just skipping them. They should be skipped, unless you have some other thought in mind here?

abroddrick: Setting pre-existing ones to false could be set in the migration file itself utilizing a similiar set up to what Rachid did with users. Alternatively, you could pass a flag to mark all existing transition domains as processed. (though flags are annoying lol)

abroddrick: Also just noticed that new-email_emailSent defaults to false, so this could also result in an email sending on domain invitations as is

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -21,6 +21,113 @@
...
+        # Test that a given TransitionDomain isn't set to "processed"
+        transition_domain_object = TransitionDomain.objects.get(domain_name="fakewebsite3.gov")
+        self.assertFalse(transition_domain_object.processed)
+
+        self.run_transfer_domains()
+
+        # Test that a given TransitionDomain is set to "processed" after we transfer domains
+        transition_domain_object = TransitionDomain.objects.get(domain_name="fakewebsite3.gov")
+        self.assertTrue(transition_domain_object.processed)
+
```
abroddrick, line 129: Looking through the code I am also seeing update_organization_data_for_all() function in extra_transition_domain_helper is doing objects.all on transition domain instead of filtering out the processed ones, that should be changed, I believe

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -21,6 +21,113 @@
...
+        # Rerun transfer domains
+        self.run_transfer_domains()
+
+        # Ensure that domain hasn't changed
+        actual_domain = Domain.objects.filter(name="fakewebsite3.gov").get()
+        self.assertEqual(changed_domain, actual_domain)
+
+        # Ensure that DomainInformation hasn't changed
+        actual_domain_information = DomainInformation.objects.filter(domain=changed_domain).get()
```
abroddrick, line 111: The issue here is that you manually adding domain info changes after running loadtransitiondomain on an empty db. But in reality we want to test the opposite, that running it on an existing database object that will not have any processed value will still result in the domain information object not being changed if it exists.

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -21,6 +21,113 @@
...
+        self.assertEqual(changed_domain_information, actual_domain_information)
+
+    def test_transition_domain_is_processed(self):
+        """
+        This test checks if a domain is correctly marked as processed in the transition.
+        """
+        self.run_load_domains()
```
abroddrick, line 118: again you would want to already have existing transition domains here before running. Then after tis load, check that the original transition domains with domain infos etc in the table are set to true, and check that the new ones are false (just as you are in the line below)

src/registrar/tests/test_transition_domain_migrations.py
zandercymatics: Nice catch

src/registrar/tests/test_transition_domain_migrations.py
zandercymatics: Agreed but we still want to test what happens if we need to run this script twice (for example, the EPP connection cuts out midway through running it)

src/registrar/tests/test_transition_domain_migrations.py
zandercymatics: Okay, looked into this again - made me realize something: Isn't the intent of the load_org script to run on already processed domains? I.e., we run the transfer script then afterwards, we update the given domain later with Org info?

src/registrar/management/commands/load_transition_domain.py
abroddrick: So this tells me that the ONLY way to really test this functionality is on a sandbox, not locally. As the setup should be:
1.) have transition domains already loaded on a sandbox
2.) upload this code
3.) run migrations
4.) add the files and follow the rest of the setup.

src/registrar/tests/test_transition_domain_migrations.py
abroddrick: yeah but did you cover the case I mention, am I just missing it? again just a comment, not a blocking thing given we need this sooner so not really worried a ton on code quality stuff

zandercymatics: @abroddrick See below...
image

src/registrar/tests/test_transition_domain_migrations.py
zandercymatics: @abroddrick I don't know if that is accurate, these get marked as processed after the bulk_create of DomainInformation which is where this information is updated/added at. This is more or less the end of the script.
image

PR #1454, opened 2023-12-06
Dashboard: adjust placeholder messaging

Katherine-Osos:
## Ticket

The placeholder messaging for "domain requests" didn't match the wording for "domains." It also mentioned "active," but the table will display rejected and withdrawn requests, as well. 

Updated the placeholder messaging within the "domain requests" table when no requests have been created yet.

### DIFF
--- a/src/registrar/templates/home.html
+++ b/src/registrar/templates/home.html
@@ -138,7 +138,7 @@
       aria-live="polite"
     ></div>
     {% else %}
-    <p>You don't have any active domain requests right now</p>
+    <p>You don't have any domain requests yet</p>
     <!-- <p><a href="{% url 'application:' %}" class="usa-button">Start a new domain request</a></p> -->
     {% endif %}
   </section>
--- a/src/registrar/templates/home.html
+++ b/src/registrar/templates/home.html
@@ -87,7 +87,7 @@
       aria-live="polite"
     ></div>
     {% else %}
-    <p>You don't have any registered domains yet</p>
+    <p>You don't have any registered domains.</p>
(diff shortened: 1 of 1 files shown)


src/registrar/templates/home.html
```diff
@@ -138,7 +138,7 @@ <h2>Domain requests</h2>
       aria-live="polite"
     ></div>
     {% else %}
-    <p>You don't have any active domain requests right now</p>
+    <p>You don't have any domain requests yet</p>
```
abroddrick, line 141: what no sad face emoji?

abroddrick (approved):

src/registrar/templates/home.html
Katherine-Osos: I was trying to match the wording in the Domains table, but I think your suggestion makes sense. What do you think of the word "registered" in the Domains message?

You don't have any **registered** domains yet

michelle-rago: I can't think of anything better, but we should explain "registered" in the help content. (I don't know what that means technically. I would remove "yet" in all cases.

Katherine-Osos: @michelle-rago Where does it say "Your domain applications?"

michelle-rago: Oh, just realized it is for the screenreader. `<caption class="sr-only">Your domain applications</caption>

Katherine-Osos:
@michelle-rago Ready for you to review this again!

michelle-rago (approved):
Super!

PR #1461, opened 2023-12-07
Add language to say that an available domain is not guaranteed

### LINKED ISSUE #781: Content: Clarify that a domain that appears "available" is not guaranteed
### Story
As an applicant
I want to understand that my requested domain is not guaranteed
so that I can set appropriate expectations for myself and my team.

### Acceptance Criteria
On the ".gov Domain" step in the application process, language appears that clarifies a requested domain is not guaranteed.

### Additional Context
During user testing, most people understood what to input, but it could’ve been more explicit that the domain they entered was not guaranteed, even if available. When prompted, most participants understood the process they would go through to get the domain, but the language could be more explicit.

### Issue Links
_No response_

michelle-rago:
## Ticket

Resolves #781 

## Changes

- Added language to let a user know, more explicitly, that an available domain isn't automatically theirs

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

## Screenshots

image

### DIFF
--- a/src/api/views.py
+++ b/src/api/views.py
@@ -34,7 +34,7 @@ DOMAIN_API_MESSAGES = {
     "invalid": "Enter a domain using only letters, numbers, or hyphens (though we don't recommend using hyphens).",
     "success": "That domain is available! We’ll try to give you the domain you want, "
                "but it's not guaranteed. After you complete this form, we’ll "
-                "evaluate whether your request meets our requirements.",
+               "evaluate whether your request meets our requirements.",
     "error": GenericError.get_error_message(GenericErrorCodes.CANNOT_CONTACT_REGISTRY),
 }
 
--- a/src/api/views.py
+++ b/src/api/views.py
@@ -33,8 +33,8 @@ DOMAIN_API_MESSAGES = {
     ),
     "invalid": "Enter a domain using only letters, numbers, or hyphens (though we don't recommend using hyphens).",
     "success": "That domain is available! We’ll try to give you the domain you want, " 
-               "but it's not guaranteed. After you complete this form, we’ll evaluate whether your " 
-               "request meets our requirements.",
+               "but it's not guaranteed. After you complete this form, we’ll "
+                "evaluate whether your request meets our requirements.",
     "error": GenericError.get_error_message(GenericErrorCodes.CANNOT_CONTACT_REGISTRY),
 }
 
--- a/src/api/views.py
+++ b/src/api/views.py
@@ -32,7 +32,7 @@ DOMAIN_API_MESSAGES = {
         "Read more about choosing your .gov domain.</a>".format(public_site_url("domains/choosing"))
     ),
     "invalid": "Enter a domain using only letters, numbers, or hyphens (though we don't recommend using hyphens).",
-    "success": "That domain is available! We’ll try to give you the domain you want, " 
+    "success": "That domain is available! We’ll try to give you the domain you want, "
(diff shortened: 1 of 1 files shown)


src/api/views.py
```diff
@@ -32,7 +32,9 @@
         "Read more about choosing your .gov domain.</a>".format(public_site_url("domains/choosing"))
     ),
     "invalid": "Enter a domain using only letters, numbers, or hyphens (though we don't recommend using hyphens).",
-    "success": "That domain is available!",
+    "success": "That domain is available! We’ll try to give you the domain you want, "
```
erinysong, line 35: Had some time to look into the linting error: First thing that came to mind is that DOMAIN_API_MESSAGES is a 1-to-1 key-value dictionary that maps a message-category to the message content text. The code interpreter/compiler is likely confused because it's reading the mapping for success message-category as 3 items (for the 3 lines we added to accommodate our new longer success text) when each message category is only supposed to have 1 item assigned to it. Most likely this is what's causing the linter to fail (and would also lead to this message erroring when we try passing it into the UI possibly). 

**Solution**
When we want to write a single text string that consists of multiple lines of code, I put a backwards slash at the end of the line to indicate to the code interpreter that my I want my single string to include what's on the next line ([More info here](https://note.nkmk.me/en/python-long-string/#:~:text=In%20Python%2C%20a%20backslash%20(%20%5C,continue%20on%20the%20next%20line.)). I can push an update to this branch to update this!

erinysong (changes requested):

src/api/views.py
erinysong: Pushing a change that updates the following API message:
**Before**
```
"success": "That domain is available! We’ll try to give you the domain you want, "
           "but it's not guaranteed. After you complete this form, we’ll "
           "evaluate whether your request meets our requirements.",
```

**After**
```
"success": "That domain is available! We’ll try to give you the domain you want, \
               but it's not guaranteed. After you complete this form, we’ll \
               evaluate whether your request meets our requirements.",
```

michelle-rago: @erinysong Thank you!!!!!

erinysong (approved):
I personally pushed changes to this PR branch so would like for someone else to take a look and verify that it works on my end. Otherwise LGTM!

src/api/views.py
```diff
@@ -32,7 +32,9 @@
         "Read more about choosing your .gov domain.</a>".format(public_site_url("domains/choosing"))
     ),
     "invalid": "Enter a domain using only letters, numbers, or hyphens (though we don't recommend using hyphens).",
-    "success": "That domain is available!",
+    "success": "That domain is available! We’ll try to give you the domain you want, \
```
therealslimhsiehdy, line 35: **Question (nonblocking):** Would we also want to change the "success" wording on the 11ty website as well for the availability button?

therealslimhsiehdy (approved):
Screenshot 2023-12-11 at 11 39 53 AM

Great catch + resolution for the linter issue @erinysong!

Katherine-Osos (approved):

Katherine-Osos:
LGTM!

src/api/views.py
michelle-rago: > **Question (nonblocking):** Would we also want to change the "success" wording on the 11ty website as well for the availability button?

Good callout! I don't think we'd make the message exactly the same because the user won't be in the midst of completing the form, but we can clarify the difference between "available" and "you're getting it!"

PR #1487, opened 2023-12-12
Change email links to point to manage.get.gov and the correct object ID

neilmb:
## Changes

In our soft soft-launch, we noticed that the approval emails had the wrong host for the registrar (registrar.get.gov instead of manage.get.gov) and that the links in the "approved" email were using the ID of the application instead of the ID of the approved domain.

This changes all of the links in our emails to use manage.get.gov and in the approved email, it fixes the links to point to the ID of the approved domain.

## Setup

These are going to best be reviewed by inspection. If you want to try testing it, you need to be on a sandbox that has AWS email-sending credentials set up. We should test this as part of manual testing once this goes onto staging.


Katherine-Osos:
@neilmb I'll be reviewing all emails this sprint, so should I just test this then? i.e. approve this ticket for now and make sure all the links look ok when I do my email review?

neilmb:
> @neilmb I'll be reviewing all emails this sprint, so should I just test this then? i.e. approve this ticket for now and make sure all the links look ok when I do my email review?

@Katherine-Osos I think that's best to approve this and review the results as part of your larger review

Katherine-Osos (approved):

PR #1491, opened 2023-12-13
Issues 1464, 1468:  Auth / contact / user refactor to solve issues introduced by switch to ial1 for some users (RJM sandbox).

### LINKED ISSUE #1464: Change displayed User firstname/lastnames to use the Contact table instead.
### Issue description

After changing our login.gov setup to no longer require identity proofing we realized that first and last names in the User table are disappearing. Login.gov is returning an empty first/last name combo for those fields now; however, we are accepting that value and overwriting any stored values we have in the user table. This issue will continue as long as we don't require identity proofing. The main issue with this comes to play on django admin, as that then leaves user emails as the only thing appearing in user related drop downs. This isn't the most user friendly. 

A simple solution is just to make sure any references we have to user's first and last names goes off of the Contact table and not the User table. The reasoning is, a user can manually update their contact table info and ensure that what we have is correct.

### Acceptance criteria

- [ ] All user centered drop downs  on django admin should now display the first name and last name from the Contact table if it exists. These include but aren't limited to submitter, creator, and investigator.

### Additional context

_No response_

### Links to other issues

Pick up #1468 with this ticket, and solve both together, if possible.

### LINKED ISSUE #1468: BUG: On auth to DjAdmin, username data is replaced by UUID
### Current Behavior

Noticed that the Username displayed on the top of the Django Admin banner has recently changed to display the long UUID, per the attached screenshot. This was noticed during 12/7/23 Refinement.

Image

**_This corresponds with recent changes to the login.gov identity proofing strategy, and seems to be related to that configuration change. See links below_**

### Expected Behavior

First and Last name should not be overwritten as a result of the call to Login.gov.

### Steps to Reproduce

1. Confirm (with someone else logged in as Superuser) that your FNAME and LNAME are set.
2. Login to staging/admin as usual.
3. Observe that your UUID is displayed in the top banned, instead of your name
4. Confirm that your FNAME and LNAME have been deleted from your User record.

### Environment

Staging Admin

### Additional Context

Upon Further investigation, I noticed that the FNAME and LNAME are being wiped out upon login. Paul verified this with Vicky.

Facts:
#1 - The UUID now being displayed is mine, so that correlates.
#2 - I checked Vicky's UserID (while in Superuser mode) and saw that her name (FAKEY) was set.
Screenshot 2023-12-07 at 4 42 14 PM

#3 - After Vicky logged in, her Fname and Lname (at least) had been deleted.
Screenshot 2023-12-07 at 4 45 09 PM

This resulted in her UUID being displayed in the top banner, rather than her name.

User's be displayed in the top banner of Django Admin. However, this is just a sympton of the data curruption that's happening.

### Issue Links

Suspect that this is Related to: #1433 and #1435 as these changes correlate with the bug arrival.

Fix in tandem with #1464

rachidatecs:
## Ticket

Resolves #1464
Resolves #1468 

## Changes

- When updating a user in djangooidc backends, update without overwriting first_name and last_name. Overwrite first_name and last_name if not empty string.
- Script to copy over first and last names from contact to linked user if the first and last names in the linked user are both none/ empty.
- Change save() in Contact to overwrite the first and last name in User.
- Unit tests for all the above.

## Context for reviewers

The solution is 3 independent parts that all work together to resolve the data corruption and make sure it stops happening. The script will restore the lost data, the changes to the oidc backend will stop ial1 from deleting first and last names, and the changes to the Contact's save() will ensure users can enter a first and last name in case ial2 is never pulling them from login.gov (transition users, invited users, users with existing domain roles that for some reason do not have first and last info in a linked contact(unlikely)).

## Setup

- Login, go to admin, see hello firstname in the top right.
- Delete the first and last names for a user that has a linked contact with first and last name values. Run `docker compose run -T app ./manage.py copy_names_from_contacts_to_users --debug`, refresh admin and see the user's first and last repopulated from contact.
- Change a contact's first and last names, save, see them updated in the linked user (can be done in Django admin).

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


src/registrar/tests/test_admin.py
```diff
@@ -1164,7 +1164,7 @@ def test_alphabetically_sorted_fk_fields_domain_information(self):
         tested_fields = [
             DomainInformation.authorizing_official.field,
             DomainInformation.submitter.field,
-            DomainInformation.creator.field,
+            # DomainInformation.creator.field,
```
rachidatecs, line 1167: 3 exclusions to keep these 2 tests passing after the edit to common.AuditedAdminMockData.dummy_user

src/registrar/tests/common.py
```diff
@@ -161,7 +162,7 @@ def dummy_user(self, item_name, short_hand):
         user = User.objects.get_or_create(
             first_name="{} first_name:{}".format(item_name, short_hand),
             last_name="{} last_name:{}".format(item_name, short_hand),
-            username="{} username:{}".format(item_name, short_hand),
+            username="{} username:{}".format(item_name, str(uuid.uuid4())[:8]),
```
rachidatecs, line 165: After the edit on Contact.save, this dude breaks unless we make sure it's unique

src/registrar/management/commands/copy_names_from_contacts_to_users.py
```diff
@@ -0,0 +1,248 @@
...
+        logger.info(
+            f"""{TerminalColors.OKCYAN}
+            ========= Adding Domains and Domain Invitations =========
+            {TerminalColors.ENDC}"""
+        )
+        (
+            skipped_contacts,
+            eligible_users,
+            processed_users,
+        ) = self.process_contacts(
```
neilmb, line 236: **comment (non-blocking)**: I think that this signature is confusing. It takes the (empty) lists as arguments so that it can append to them, but then it returns the appended lists as well? Why not one or the other? 

Either take no lists as arguments and return the lists from inside the function, or take the list objects to be added to as arguments but then you don't need to return them. I personally think the first option of no arguments and just returning the lists is most pythonic. Then you won't have to initialize the empty lists in this calling function.

src/registrar/models/contact.py
```diff
@@ -59,6 +59,16 @@ def get_formatted_name(self):
         names = [n for n in [self.first_name, self.middle_name, self.last_name] if n]
         return " ".join(names) if names else "Unknown"
 
+    def save(self, *args, **kwargs):
+        # Call the parent class's save method to perform the actual save
+        super().save(*args, **kwargs)
+
+        # Update the related User object's first_name and last_name
+        if self.user:
```
neilmb, line 67: **question (non-blocking):**: This is very clever to keep user objects in sync with contacts rather than vice versa. But for an existing user who does log in with identity verification, if their `given_name` and `family_name` have changed with Login.gov (can this ever happen?) will their `User` object get updated by not their `Contact` object? Is that what we want?

src/registrar/tests/test_copy_names_from_contacts_to_users.py
```diff
@@ -0,0 +1,98 @@
...
+    def setUp(self):
+        """We cannot setup the user details because contacts will override the first and last names in its save method
+        so we will initiate the users, setup the contacts and link them, and leave the rest of the setup to the test(s).
+        """
+
+        self.user1 = User.objects.create(username="user1")
+        self.user2 = User.objects.create(username="user2")
+        self.user3 = User.objects.create(username="user3")
+        self.user4 = User.objects.create(username="user4")
+        # The last user created triggers the creation of a contact and attaches itself to it. @Neil wth is going on?
```
neilmb, line 21: **comment**: It's signals again. https://github.com/cisagov/manage.get.gov/blob/main/src/registrar/signals.py#L13 should create the Contact for a user every time the new User object is saved.

src/registrar/tests/test_models.py
```diff
@@ -654,3 +654,48 @@ def test_identity_verification_with_new_user(self):
         """A new user who's neither transitioned nor invited should
         return True when tested with class method needs_identity_verification"""
         self.assertTrue(User.needs_identity_verification(self.user.email, self.user.username))
+
+
+class TestContact(TestCase):
+    def setUp(self):
+        self.email = "mayor@igorville.gov"
+        self.user, _ = User.objects.get_or_create(email=self.email, first_name="Rachid", last_name="Mrad")
```
neilmb, line 662: **Comment (non-blocking)**: It's a little unusual for us to use our actual first and last names in our test cases. Should this just be "John Doe" or "Firstname Lastname" instead?

neilmb (approved):
This is great work, especially nice unit testing. One question that still stands is what we want to happen if someone's name gets updated with Login.gov. I don't know if that can even happen with their identity verification product, but I can't see where that change would propagate here, or even if we want it to.

dave-kennedy-ecs:
> This is great work, especially nice unit testing. One question that still stands is what we want to happen if someone's name gets updated with Login.gov. I don't know if that can even happen with their identity verification product, but I can't see where that change would propagate here, or even if we want it to.

It would not propogate here.  The user will at most once to log in with IAL2, and that is the only time that their name comes over from login.gov.  On subsequent logins, they will do IAL1, and the first name and last name will not be in the payload from login.gov.  

If the first_name and last_name do somehow come over from login.gov, this code would handle it and overwrite the first_name and last_name in the user.  But we are not seeing the first_name and last_name coming over in IAL1, so in essence, login.gov name changes don't propogate here.

src/registrar/models/contact.py
dave-kennedy-ecs: No, this will not get updated from login.gov, per my other comment.  It is not that we can't handle receiving this from login.gov, it is just that login.gov is not sending it when user logs in with IAL1.

src/registrar/tests/test_models.py
rachidatecs: Jeff Lebowski it is

src/djangooidc/tests/test_backends.py
```diff
@@ -0,0 +1,99 @@
...
+        User.objects.all().delete()
+
+    def test_authenticate_with_create_user(self):
+        """Test that authenticate creates a new user if it does not find
+        existing user"""
+        # Ensure that the authenticate method creates a new user
+        user = self.backend.authenticate(request=None, **self.kwargs)
+        self.assertIsNotNone(user)
+        self.assertIsInstance(user, User)
+        self.assertEqual(user.username, "test_user")
```
zandercymatics, line 27: Nice checks

src/registrar/management/commands/copy_names_from_contacts_to_users.py
```diff
@@ -0,0 +1,236 @@
...
+                    > Found linked user for contact:
+                    {contact} {contact.email} {contact.first_name} {contact.last_name}
+                    > The linked user is {eligible_user} {eligible_user.username}
+                    {TerminalColors.ENDC}""",  # noqa
+                )
+
+                # ---- UPDATE THE USER IF IT DOES NOT HAVE A FIRST AND LAST NAMES
+                # ---- LET'S KEEP A LIGHT TOUCH
+                if not eligible_user.first_name and not eligible_user.last_name:
+                    # (expression has type "str | None", variable has type "str | int | Combinable")
```
zandercymatics, line 115: That is odd. I've ran into similar issues in the past with the linter. I think ignoring this is fine as long as there is no foreseeable way that this can cause issues, and I think this is probably the case given the db definition.

src/registrar/management/commands/copy_names_from_contacts_to_users.py
```diff
@@ -0,0 +1,236 @@
...
+                    eligible_user.last_name = contact.last_name  # type: ignore
+                    eligible_user.save()
+                    processed_user = eligible_user
+
+                return (
+                    eligible_user,
+                    processed_user,
+                )
+
+            except Exception as E:
```
zandercymatics, line 127:
```suggestion
            except Exception as error:
```
(nitpick / optional) `E` doesn't follow our naming conventions, I think `error` may be more descriptive. Tiny nitpick though

src/registrar/management/commands/copy_names_from_contacts_to_users.py
```diff
@@ -0,0 +1,236 @@
...
+                logger.warning(
+                    f"""
+                    {TerminalColors.FAIL}
+                    !!! ERROR: An exception occured in the
+                    User table for the following user:
+                    {contact.email} {contact.first_name} {contact.last_name}
+
+                    Exception is: {E}
```
zandercymatics, line 135:
```suggestion
                    Exception is: {error}
```
(You'll have to change it here too if you decide to)

src/registrar/management/commands/copy_names_from_contacts_to_users.py
```diff
@@ -0,0 +1,236 @@
...
+
+    # ======================================================
+    # ===================    USER    =====================
+    # ======================================================
+    def update_user(self, contact: Contact, debug_on: bool):
+        """Given a contact with a first_name and last_name, find & update an existing
+        corresponding user if her first_name and last_name are null.
+
+        Returns the corresponding User object.
+        """
```
zandercymatics, line 93: (Nice to have) Could you note in the function description that this returns a tuple than just a user object?

src/registrar/management/commands/copy_names_from_contacts_to_users.py
```diff
@@ -0,0 +1,236 @@
...
+                )
+                sys.exit()
+        else:
+            return None, None
+
+    # ======================================================
+    # ================= PROCESS CONTACTS  ==================
+    # ======================================================
+
+    # C901 'Command.handle' is too complex
```
zandercymatics, line 146:
```suggestion
```
May be an accidental copy+paste remnant, remove?

src/registrar/management/commands/copy_names_from_contacts_to_users.py
```diff
@@ -0,0 +1,236 @@
...
+    def process_contacts(
+        self,
+        debug_on,
+        skipped_contacts=[],
+        eligible_users=[],
+        processed_users=[],
+    ):
+        for contact in Contact.objects.all():
+            # DEBUG:
```
zandercymatics, line 155:
```suggestion
```
Same as above. I always end up copying this one in, I feel the pain

src/registrar/management/commands/copy_names_from_contacts_to_users.py
```diff
@@ -0,0 +1,236 @@
...
+            TerminalHelper.print_conditional(
+                debug_on,
+                f"{TerminalColors.OKCYAN}"
+                "Processing Contact: "
+                f"{contact.email},"
+                f" {contact.first_name},"
+                f" {contact.last_name}"
+                f"{TerminalColors.ENDC}",  # noqa
```
zandercymatics, line 163: Why is there a noqa here?

src/registrar/tests/common.py
```diff
@@ -405,8 +406,8 @@ def mock_user():
     """A simple user."""
     user_kwargs = dict(
         id=4,
-        first_name="Rachid",
-        last_name="Mrad",
+        first_name="Jeff",
+        last_name="Lebowski",
```
zandercymatics, line 410: Nice

src/registrar/tests/test_admin.py
zandercymatics: Totally fine to keep them commented out for now. The long term solution would be to reduce the engineering complexity of the tests, but I think I have an idea for how to fix those test cases

zandercymatics (approved):
This PR was a lot of fun to read, I appreciate the bits of humor sprinkled everywhere. Great job on this. The only thing code wise is just some minor cleanup. Mostly optional, though some old comments snuck in there from the copy+paste (I am guilty of the same from these files)

zandercymatics:
@rachidatecs Code looks good but you have a test case failure

PR #1494, opened 2023-12-14
(NOT ON A SANDBOX) Ticket #1448: Filter by Investigator

### LINKED ISSUE #1448: Django admin: Filter list of names (and alphabetize) the "by investigator" filter
### Issue description

Django admin: When viewing the "domain application" section there's a filter for "by investigator" on the right side of the page. Currently, this lists **all** users,  but it should only show users who have admin access. Also need to alphabetize the list

### Acceptance criteria

- [ ] "By investigator" filter only shows users who have Staff permission (is_staff )
- [ ] "By investigator" filter lists values in alphabetical order.

### Additional context

_No response_

### Links to other issues

_No response_

zandercymatics:
## Ticket

Resolves #1448 

## Changes

- Adds a custom filter `InvestigatorFilter` inside of `DomainApplicationAdmin` (admin.py) and links it up to that
- Adds a get_queryset re-implementation to order by domain name
- Test cases 

## Context for reviewers
This ticket is intended to narrow down the investigator filter list on the DomainApplication admin table page. To that affect, all users without the permission "is_staff" are excluded. Additionally, the output should be sorted alphabetically.

## Setup
1. Navigate to `/admin`
2. Click `Users`
3. Click any user from the table, and under `permissions`  disable `staff status` (the is_staff field). By default, fixtures gives all users the staff permission, we want to remove this permission on one user for testing purposes.
4. Click `Domain Application`
5. Note the sorting order, is it alphabetical?
6. On the righthand side, you should see the `By investigator` set of filters. Try to find the user that you modified. If you can't, that means this step was successful
7. Click any DomainApplication record from the table. Add an investigator.
8. Go back to the DomainApplicatoin table, and try sorting by that investigator you just added 

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


src/registrar/admin.py
```diff
@@ -547,6 +549,25 @@ class DomainApplicationAdmin(ListHeaderAdmin):
 
     """Custom domain applications admin class."""
 
+    class InvestigatorFilter(admin.SimpleListFilter):
+        """Custom investigator filter that only displays users with the manager role"""
+        title = "investigator"
+        parameter_name = "investigator"
```
zandercymatics, line 555: Should either of these be changed?

src/registrar/admin.py
```diff
@@ -547,6 +549,23 @@ class DomainApplicationAdmin(ListHeaderAdmin):
...
+            """Lookup reimplementation, gets users of is_staff.
+            Returns a list of tuples consisting of (user.id, user)
+            """
+            privileged_users = User.objects.filter(is_staff=True)
+            return [(user.id, user) for user in privileged_users]
+
+        def queryset(self, request, queryset):
+            """Custom queryset implementation, filters by investigator"""
+            return queryset.filter(investigator=self.value())
+
```
zandercymatics, line 568: This reduces what we can filter by, but it implies that we don't expect investigators to be users that aren't is_staff. This may be a model level change, but I'm taking the conservative approach and am not changing the model definition as that was not explicit in the ticket.

Would this be better to do it from the model level? I tested it and validation can be done on save() to check for is_staff

src/registrar/admin.py
```diff
@@ -641,6 +660,18 @@ def formfield_for_manytomany(self, db_field, request, **kwargs):
             kwargs["queryset"] = models.Website.objects.all().order_by("website")  # Sort websites
         return super().formfield_for_manytomany(db_field, request, **kwargs)
 
+    def formfield_for_foreignkey(self, db_field, request, **kwargs):
+        # Removes invalid investigator options from the investigator dropdown
+        if db_field.name == "investigator":
+            kwargs["queryset"] = User.objects.filter(is_staff=True)
+            return db_field.formfield(**kwargs)
+        return super().formfield_for_foreignkey(db_field, request, **kwargs)
```
zandercymatics, line 668: Given that we have to do this, it implies that investigators are "invalid" if they don't have the role is_staff. Should we also make a model level change as well to throw an error on create/save, if investigator is not is_staff?

I took the conservative approach for this PR and did not include that change (as it wasn't specified in the ticket).

zandercymatics:
The tests on here started breaking for some reason, despite no substantial code change. I think its a test interference thing with a recent merge.

Will investigate, but the PR is testable otherwise

src/registrar/admin.py
```diff
@@ -641,6 +663,18 @@ def formfield_for_manytomany(self, db_field, request, **kwargs):
...
+
+    def get_queryset(self, request):
```
rachidatecs, line 673: This is throwing so many red flags. All columns should be sortable by default when you click on the column header. It seems that this functionality is working on some tables (domains) but not on others (DAs and DIs). Maybe we introduced a bug at some point?

In either case, I don't this this should be needed.

src/registrar/tests/test_admin.py
```diff
@@ -842,12 +837,115 @@ def custom_is_active(self):
...
+        request = self.factory.get("/")
+        request.user = self.superuser
+
+        # Grab the current list of table filters
+        readonly_fields = self.admin.get_list_filter(request)
+        expected_fields = ("status", "organization_type", DomainApplicationAdmin.InvestigatorFilter)
+
+        self.assertEqual(readonly_fields, expected_fields)
+
+    def test_table_sorted_alphabetically(self):
```
rachidatecs, line 851: Good test, maybe belongs on another ticket. See my comment above on ordering.

src/registrar/tests/test_admin.py
```diff
@@ -842,12 +837,115 @@ def custom_is_active(self):
...
+        # Get the returned queryset
+        queryset = self.admin.get_queryset(request)
+
+        # Check the order
+        self.assertEqual(
+            list(queryset),
+            list(expected_order),
+        )
+
+    def test_displays_investigator_filter(self):
```
rachidatecs, line 871: I don't get what this test is doing that's different than test_has_correct_filters and test_investigator_filter_filters_correctly. I do not mind redundancy in tests so non-blocking, but the 3 tests could use more detailed string descriptions.

rachidatecs (commented):
Missing unit test and unnecessary method that can be replaced with a DjA setting.

src/registrar/admin.py
rachidatecs: I think this is fine. Right now, if you remove is_staff from a user the application that she's investigator on will not throw errors. They'll show blank for investigator because you are excluding non-staff from the dropdown but in the DB the investigator is still assigned. 

I think that's OK, just need to make sure the team knows about this.

src/registrar/tests/test_admin.py
zandercymatics: Will do, thanks. Mostly redundancy in the event the filter is still there but it doesn't sort correctly

rachidatecs:
The list of investigators in the filter list needs to be sorted alphabetically.

therealslimhsiehdy (approved):
Approving on caveat that I agree with Rachids comments, we may want to have a unit test for `formfield_for_foreignkey` -- otherwise looks good to me :) 

Before I removed Gaby as a user:
Screenshot 2023-12-19 at 3 00 25 PM
Removed Gaby as a user, and you'll see that Gaby only shows up once:
Screenshot 2023-12-19 at 3 01 34 PM
Added Gaby back in (it looks like Gaby shows 3 times, but it's 2 -- the 2nd one is her analyst + email thats really long hence why it looks like 3):
Screenshot 2023-12-19 at 3 02 17 PM

Katherine-Osos (approved):

dave-kennedy-ecs (approved):
code looks good

PR #1499, opened 2023-12-14
(on getgov-za) Ticket #1450:  No name for invitation email bug

### LINKED ISSUE #1450: Email for "You've been added to a .gov domain" doesn't name the person who initiated the invitation
### Current Behavior

When a new domain manager is added, users receive an email that looks odd and reads suspiciously, like something is missing. It doesn't actually say who initiated the invitation:

image

### Expected Behavior

When a non-CISA-admin domain manager adds a new domain manager, the email text should not try to add a first/last name of the creator and should instead show the email of whoever clicked the add user button (no check on creator)

When a cisa-admin adds a new domain manager, then the email shown should be help@get.gov. (this includes anyone with is_staff access)

### Steps to Reproduce

1. Add a new domain manager at an email address you control.
2. See that the email text is wrong.

### Environment

Stable: manage.get.gov

### Additional Context

[Slack thread](https://cisa-corp.slack.com/archives/C05BGB4L5NF/p1701893276703299?thread_ts=1701892663.512319&cid=C05BGB4L5NF)

> Found the bug, it is set to print the creator of the domain, not the person who sent the invitation. And ALL transition domains have no creator first name/ last name

### Issue Links

_No response_

zandercymatics:
## Ticket

Resolves #1450

## Changes

- Modify domain.py such that it will pull the first and last name from the domain requester, rather than the domain creator.
- Modify an associated test case to check email content

## Context for reviewers
The invitation email incorrectly assumed that the domain creator would always be the same user inviting you to a domain. This ticket pulls that information from the user making the request, and will now show the users email address instead of a first/last name pair.

**NOTE:** This PR is reliant on the bug fix for [#1551](https://github.com/cisagov/manage.get.gov/pull/1551). It contains content for that fix and cannot merge until that is in

## Setup
To test this, invite one of your email addresses to a domain. Do this for one where you are the creator, and one where you are not. 

For all users that have the role 'is_staff', a default email will be shown instead. To test if your own email is sent across, modify your analyst account and uncheck 'is_staff' on the User page. Then, login with that account and send an email out.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


src/registrar/tests/test_views.py
```diff
@@ -1415,11 +1415,34 @@ def test_domain_invitation_email_sent(self):
             add_page.form["email"] = email_address
             self.app.set_cookie(settings.SESSION_COOKIE_NAME, session_id)
             add_page.form.submit()
+
+        expected_email_content = {
```
neilmb, line 1419: **comment (blocking)**: This does not test the changes that you made, and it is fragile because this test will fail if we change anything about the email. I think that you should change the mock of the `send_email` method so that you can make assertions about the `Content` of the sent message. Then you should set up all of the different cases that your code works over: requester as request.user, requester as creator, and empty names so it uses emails for both. 

If you manually tested this code to make sure that it did the right thing, then you should turn each of those manual tests into unit tests so that we can verify on an ongoing basis that this code is doing the right thing.

neilmb (changes requested):
Unit testing needs improvement.

src/registrar/tests/test_views.py
zandercymatics: Will do, thanks

src/registrar/views/domain.py
```diff
@@ -653,12 +653,30 @@ def _send_domain_invitation_email(self, email: str, add_success=True):
         # created a new invitation in the database, so send an email
         domainInfoResults = DomainInformation.objects.filter(domain=self.object)
         domainInfo = domainInfoResults.first()
-        first = ""
-        last = ""
-        if domainInfo is not None:
+        first: str = ""
+        last: str = ""
+        if requester is not None:
+            first = requester.first_name
```
abroddrick, line 659: **Blocking**: this violates the expected behavior and thus the AC of that ticket "When a non-CISA-admin domain manager adds a new domain manager, the email text should not try to add a first/last name of the creator and should instead show the email of whoever clicked the add user button (no check on creator)

When a cisa-admin adds a new domain manager, then the email shown should be [help@get.gov](mailto:help@get.gov). (this includes anyone with is_staff access)"

No names should be shown at all, just email address.

abroddrick (changes requested):
See blocking comment

src/registrar/views/domain.py
zandercymatics: Fixed!

src/registrar/views/domain.py
```diff
@@ -644,21 +643,20 @@ def _domain_abs_url(self):
...
-        if domainInfo is not None:
-            first = domainInfo.creator.first_name
-            last = domainInfo.creator.last_name
-        full_name = f"{first} {last}"
+
+        # Check if the email requester has a valid email address
+        if requester.email is not None and requester.email.strip() != "":
+            requester_email = requester.email
+        else:
+            # This edgecase would be unusual if encountered. We don't want to handle this here. Rather, we would
```
abroddrick, line 657: **blocking** I disagree. with raising a valueError that produces a 500 server error which we should avoid in favor of a human readable message. I would instead look at how we handle the emailSendingError  in the below Except. There we add a message to the user. I think I would do this with messages.error  and be sure to log this with logger.info with something identifiable so we can find which user it is that is producing this issue (first/last name or UUID would be reasonable here).

src/registrar/tests/test_views.py
```diff
@@ -1415,13 +1414,153 @@ def test_domain_invitation_email_sent(self):
...
+                self.app.set_cookie(settings.SESSION_COOKIE_NAME, session_id)
+                add_page.form.submit().follow()
+
+        expected_message_content = "Can't send invitation email. No email is associated with your account."
+
+        # Grab the message content
+        returned_error_message = mock_error_message.call_args[0][1]
+
+        # Check that the message content is what we expect
+        self.assertEqual(expected_message_content, returned_error_message)
```
zandercymatics, line 1530: Doing an assert_has_calls was not working for me in this context, as it expects the request as well as the expected message. I wasn't able to figure out a way to get these two requests to match. They appear to be the same but something isn't matching on them.

I think the test above is OK, as the patch implies what kind of error we are receiving and what we'd like to receive. Other unit tests already test the redirect on this URL. 

```
Expected: error(<WSGIRequest: POST '/domain/1/users/add'>, "Can't send invitation email. No email is associated with your account.")
Actual: error(<WSGIRequest: POST '/domain/1/users/add'>, "Can't send invitation email. No email is associated with your account.")
```

neilmb:
I still don't see anything in here where "help@get.gov" will be used for staff users who send invitations instead of their individual Login.gov email address?

src/registrar/tests/test_views.py
```diff
@@ -1415,13 +1414,153 @@ def test_domain_invitation_email_sent(self):
...
+        # Create a fake user object
+        email_address = "mayor@igorville.gov"
+        User.objects.get_or_create(email=email_address, username="fakeuser@fakeymail.com")
+
+        self.domain_information, _ = DomainInformation.objects.get_or_create(creator=self.user, domain=self.domain)
+
+        mock_client = MagicMock()
+        mock_client_instance = mock_client.return_value
+
+        with boto3_mocking.clients.handler_for("sesv2", mock_client):
```
neilmb, line 1475: **question (non-blocking)**: These test cases are starting to get very repetitious when they use the same code and mocks to send these emails. Could we use a helper method called `_send_mock_emails()` that returns `mock_client` so that it can be asserted about? Maybe the same thing for `_send_mock_emails_error()` that would return `mock_error_message` too?

zandercymatics: Agreed, though after tinkering with it - I think this particular test class should be refactored before I can do that. Moving additional functionality into helpers seems to cause some weird test interference issue, even when its really barebones. There are also quite a few functions in here as you mention which use this pattern

Katherine-Osos:
@kristinacyin will be reviewing this PR for design input

kristinacyin (changes requested):
For some reason, no matter what email I invite (is_staff or not), the email always said that "help@get.gov" added me
tempsnip

kristinacyin (approved):
LGTM! Thank you for helping to explain it!

dave-kennedy-ecs (approved):
Ran code locally.  Tested all use cases successfully.  Reviewed code and tests.  Approved.

PR #1520, opened 2023-12-15
(On getgov-za) Ticket #1456: Table sorting not working correctly

### LINKED ISSUE #1456: Django Admin: Column sorting by clicking on arrows is not working properly
### Current Behavior

On Domain Information and Domains, the following column(s) do not sort correctly: 
- Submitter  (not alphabetical)
~- Domain (not alphabetical)~
**EDIT:** This appears to be sorting correctly for Domains, but not DomainInformation - Zander

On Domain Applications, the following column(s) do not sort correctly: 
-  Requested domain (non alphabetical)
-  Submitter  (non alphabetical)
- Investigator  (non alphabetical)

On Contacts,  the following column(s) do not sort correctly:
~- Contact (non alphabetical)~
**EDIT:** This appears to be sorting correctly - Zander

On User domain roles,  the following column(s) do not sort correctly: 
- User (non alphabetical)

[Slack thread](https://app.slack.com/client/T02QH7E1MHA/C05BGB4L5NF/thread/C05BGB4L5NF-1701967623.189499)

### Expected Behavior

Column sorting by clicking on arrows work as intended on the following:
- Domain applications-Requested domain, submitter and investigator fields should sort alphabetically
- Domain information and Domains - Submitter and Domain should sort alphabetically
- Contacts - Contacts  should sort alphabetically
- User domain roles - Users should sort alphabetically

### Additional AC
1. Review of ALL tables to make sure the sort is working properly.
2. Identify and fix those that are not working properly.

### Steps to Reproduce

1. On Django admin, columns do not sort alphabetically when you click the arrows, they should sort alphabetically. (see current behaviors for specific columns).

### Environment

_No response_

### Additional Context

This is being caused by the default django behavior to sort by id, instead we need to sort by that text value in the column. This works on Domain and that can be used as an example (it is a setting change).

[slack thread ](https://cisa-corp.slack.com/archives/C05BGB4L5NF/p1701967623189499)for context
### Issue Links

_No response_

zandercymatics:
## Ticket

Resolves #1456

## Changes

- Adds a custom ChangeList implementation called `MultiFieldSortableChangeList` in admin.py to allow it to return a list of order fields, rather than just one item as before
- Modified ListHeaderAdmin to include a custom get_changelist implementation, and to include the new OrderableFieldsMixin
- Adds a new array on each applicable Admin class, `orderable_fk_fields` (does anyone know a better name?) that specifies what fk fields we want to sort by and how
- Add unit tests
- Adds `OrderableFieldsMixin` to mixins.py. This mixin allows us to automatically generate the previously pretty convoluted admin_order_field syntax, and apply those changes to the class automatically
- Remove a redundant/finicky unit test I introduced in a prior PR

## Context for reviewers
Whew. This ticket turned out to be more complicated than I initially thought it had to be.

So, as it turns out, there are two quirks with Django Admin that made this solution necessary:
1. By default, all fk fields get sorted by their _id_ when clicking the table sort button, rather than anything else
2. By default, to override this for these particular fields, you need to obey this "interesting" syntax:  
```
# Assuming you want to filter on domain...
def get_domain(self, obj):
    return obj.domain
# Allows column order sorting
get_domain.admin_order_field = "DomainInformation__domain"
# Sets column's header 
get_domain.short_description = "domain"  
```

Then, you need to modify the list_display list to use your new function:
```
    list_display = [
        "get_domain",
        ....
    ]
```

Now, this obviously becomes a problem when you want to make this change on multiple fields. Its unwieldy and doesn't make sense to follow. But it also introduces a second problem: `admin_order_field`, by design, only supports one field at a time, rather than multiple (e.g. ["first_name", "last_name"]). This means that for name fields, like submitter, we can't actually sort on both first_name and last_name simultaneously. We have to choose one, or the other... Or separate it out into separate fields. 

To solve these issues, I had to do two things: 
1. Create a mixin that would handle all that ugly syntax building for us, and...
2. Modify the original ChangeList definition such that it doesn't error out when receiving a list as an input (while maintaining support for strings)

Because I'd rather not go messing with the logic for list_display directly, this is controlled by adding a new field, `orderable_fk_fields`, which takes a list of tuples that specifies the field to display on, and what to sort it by. The mixin will then automatically apply any changes to `list_display`, and our sorting works on any number of fields. 

## Setup

**PR review instructions for designers**
This change affects how the table sort button (when you click the columns) works on the following pages in django admin:
User Domain Role, Domain Information, and Domain Application. For each of those pages, test the sorting behavior of each column. Ensure that: a) they sort correctly and b) nothing else is wonky with them 

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


zandercymatics:
The two test case failures are due to the contact/user bug. That needs to be solved externally before those pass, but a temporary workaround is to comment out the `save` in `contact.py`

src/registrar/tests/common.py
```diff
@@ -93,6 +93,73 @@ def less_console_noise(output_stream=None):
...
+            factory (RequestFactory, optional): An instance of Django's RequestFactory.
+        """
+        super().__init__()
+        self.factory = factory
+        self.user = user
+        self.admin = admin
+        self.model = model
+        self.url = url
+
+    def assert_table_sorted(self, o_index, sort_fields):
```
zandercymatics, line 115: This doesn't have to be here or exist at all, it just dramatically simplifies the test cases and I think it provides a good backbone for future expansion, if we want it. Its intentionally pretty bare bones.

I know that generally we try to avoid architectural complexity for our test cases, so if this violates that principle I can take this out and migrate these changes to each test manually.

src/registrar/tests/test_admin.py
```diff
@@ -1163,10 +1371,11 @@ def test_alphabetically_sorted_fk_fields_domain_application(self):
             )
 
     def test_alphabetically_sorted_fk_fields_domain_information(self):
+        self.maxDiff = None
         tested_fields = [
             DomainInformation.authorizing_official.field,
             DomainInformation.submitter.field,
-            # DomainInformation.creator.field,
+            #DomainInformation.creator.field,
```
zandercymatics, line 1378: This particular field is failing due to a bug in the underlying class in common. There is another ticket to reduce some engineering complexity in common.py, so that is better suited for there

abroddrick:
@zandercymatics this is user facing so it does require a designer as well to approve.

dave-kennedy-ecs (approved):
I pulled and ran locally and tested locally.  This is a much more complex problem than it looked on the surface.  Well done.

zandercymatics:
Would a designer be able to take a peek at this ticket? @cisagov/gov-designers

Katherine-Osos (approved):

PR #1557, opened 2023-12-21
Ticket #1128: Reduce email bouncebacks

### LINKED ISSUE #1128: help@get.gov is receiving lots of bouncebacks for failed test emails
Our new program inbox will be help@get.gov, hosted in M365. We are also sending automated registrar emails from this address via SES. SES is doing what it does and actually attempting to deliver test emails we're sending to fake addresses, which bounces-back into our real inbox.

Though we can filter out bounced emails from amazonses.com, let's dig into this and see if we can limit the bouncebacks or try another approach.

- [ ] Find the cause of the emails (dig into sending logic, SES logs)
- [ ] Decide on a course of action (create actual test inboxes?)
- [ ] Make a change and confirm impact
- [ ] fixtures and tests should not result in an email being sent

Check out our [Slack thread](https://cisa-corp.slack.com/archives/C05BJTNCY12/p1695764077201119) about it.

zandercymatics:
## Ticket

Resolves #1128 

## Changes

- Modifies `fixtures_application.py` to use a new boolean `do_fake_send_email` (better name pending™)
- Modifies `_send_status_update_email` in `domain_application.py` to incorporate the said boolean
- Adds an unfathomable number of boto3_mocking handlers
- Adds less_console_noise onto many of the mock emails to reduce excessive logging

## Context for reviewers
When running the application, our fixtures and test cases had the tendency to actually send out emails using our AWS credentials. This was resulting in mass email spam, as AWS will send a bounceback after the email address can't be found.

## Setup
1. Spin up the PR locally. After fixtures finish loading, ctrl+f for this text: "An email was sent!". If you get any results, then that means we unintentionally sent out an email.
2. Run `docker-compose exec app ./manage.py test registrar.tests`
3. In the console logs, look for `An email was sent!`. You shouldn't get any results that would actually send an email out, or from emails that aren't "mayor@igorville.gov" (These are from mocks)

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


src/registrar/utility/email.py
```diff
@@ -20,7 +22,7 @@ def send_templated_email(template_name: str, subject_template_name: str, to_addr
     context as Django's HTML templates. context gives additional information
     that the template may use.
     """
-
+    logger.info(f"An email was sent! Template name: {template_name} to {to_address}")
```
zandercymatics, line 25: This line is temporary and will be removed. This is for testing purposes only so we know when an email is sent. If this logs in any other context outside of tests on startup, then we are sending emails when we shouldnt be

erinysong:
Able to run the PR locally without sending any emails! When testing, the instances where I see the log `An email was sent!` are either to mayor@igorville.gov or do not successfully send

image

src/registrar/models/domain_application.py
```diff
@@ -561,8 +561,8 @@ def domain_is_not_active(self):
             return not self.approved_domain.is_active()
         return True
 
-    def _send_status_update_email(self, new_status, email_template, email_template_subject):
-        """Send a atatus update email to the submitter.
+    def _send_status_update_email(self, new_status, email_template, email_template_subject, do_fake_send_email=False):
+        """Send a status update email to the submitter.
```
erinysong, line 565: Good catch!

src/registrar/models/domain_application.py
```diff
@@ -571,7 +571,12 @@ def _send_status_update_email(self, new_status, email_template, email_template_s
 
         if self.submitter is None or self.submitter.email is None:
             logger.warning(f"Cannot send {new_status} email, no submitter email address.")
-            return
+            return None
+
+        if do_fake_send_email:
```
erinysong, line 576: I'm not confident that I understand `do_fake_send_email` means so I had 2 questions!

1. Does the flag indicate that an email can be sent for testing purposes but doesn't actually send the email?
2. Is it only used in the context of locally testing email delivery?

src/registrar/models/domain_application.py
```diff
@@ -571,7 +571,12 @@ def _send_status_update_email(self, new_status, email_template, email_template_s
...
+            logger.info(f"Email was not sent. Would send {new_status} email: {self.submitter.email}")
```
erinysong, line 577: Optional suggestion: Think it'd be good to explain why the email was not sent (similar to the above Email not sent message) so the user understands that the email not being sent is expected
` logger.warning(f"Cannot send {new_status} email, no submitter email address.")`

src/registrar/models/domain_application.py
```diff
@@ -671,7 +676,7 @@ def action_needed(self):
         ],
         target=ApplicationStatus.APPROVED,
     )
-    def approve(self):
+    def approve(self, do_fake_send_email=False):
```
erinysong, line 679: I'm wondering if instead of having a flag for fake sending email that doesn't get sent, we make the flag indicate that an email **is** sent -- that way when we make tests approving emails, we can set all tests to set the flag as false and indicate that we are setting send_email as false because we don't want our tests to actually send emails? Or in our tests, we can have a function like `test_approve` that performs an email approval with the send_email flag set to false that we can call in our tests

erinysong: I'm not familiar with how emails work, so feel free to not go with this suggestion if it doesn't align with how the registrar sends and tests emails!

src/registrar/tests/test_transition_domain_migrations.py
```diff
@@ -18,7 +18,8 @@
 
 from registrar.models.contact import Contact
 
-from .common import MockEppLib, less_console_noise
+from .common import MockEppLib, MockSESClient, less_console_noise
+import boto3_mocking  # type: ignore
```
erinysong, line 22: Nonblocking info question - What does the `type: ignore` comment mean? Noticed that it's in our original code as well so was curious

erinysong (commented):
Was able to successfully test! Had a few questions on how to set up the fake sending email flag but other than that LGTM!

src/registrar/models/domain_application.py
zandercymatics: Probably because of confusing variable naming on my part! "send_email" might be less confusing
1. Yup. When we first start the app and run fixtures (generating fake data), it will "approve" a predetermined number of domains. Since we send an email on approval, this means that we are trying to send out 20-40 real emails everytime we docker-compose up. This variable is used exclusively for fixtures, and instead of actually sending them, it will just make a log entry instead
2. No - the intend for the change is that will be used only in our fixtures file, but not anywhere else

src/registrar/models/domain_application.py
zandercymatics: Great suggestion, thank you

src/registrar/models/domain_application.py
zandercymatics: Yeah I have no problem making that change, it would be more clear. I agree that setting it to `send_email=True` makes more contextual sense

src/registrar/tests/test_transition_domain_migrations.py
zandercymatics: Its part of [mypy](https://mypy-lang.org/) which we use for typing/linting. Basically its a linter override, so it just ignores that line. That import is a little wonky for some reason, not sure why but I've noticed that used on other boto3_mocking instances so I just kept with the pattern

Here is a good resource:
https://stackoverflow.com/questions/49220022/how-can-mypy-ignore-a-single-line-in-a-source-file

src/registrar/tests/test_admin.py
```diff
@@ -456,6 +460,7 @@ def test_save_model_sends_approved_email(self):
         # Perform assertions on the mock call itself
         mock_client_instance.send_email.assert_called_once()
 
+    @boto3_mocking.patching
```
rachidatecs, line 463: Why these additions throughout the tests?

src/registrar/utility/email.py
rachidatecs: What do you mean by tests on startup?

rachidatecs (commented):
Non-blocking unless you want it to be: Why not set the param `send_email=False` as a default? There is a handful of transitions we want to trigger emails, so we'll call using `send_email=True` from those.

This leaves us free to forget about fixtures and tests, and would allow us to clean up all the with_less_console_noise throughout the suite.

Also, linter.

src/registrar/tests/test_admin.py
zandercymatics: These seem to be redundant with `with boto3_mocking.clients.handler_for("sesv2", mock_client):`. I think its a code leftover, thanks for catching that

src/registrar/utility/email.py
zandercymatics: @rachidatecs Tests **or** startup

src/registrar/tests/test_admin.py
zandercymatics: Its related to the boto3_mocking client, basically when we call: `with boto3_mocking.clients`

abroddrick (commented):
see blocking question

zandercymatics:
> blocking question I don't quite understand why it's okay that this shows for some tests that are using mayor. Can you explain? I am trying to review this quick so I am not reading every line
It has to do with mocking. The MockSESEmail function just directly overrides the send_email command (The AWS bit) but it doesn't directly override that function so it still logs it. That said, the 7 that are logging were using MagicMock, rather than MockSES - so I will fix that. Did not notice that until now

abroddrick (approved):
I'll admit not my most thorough review, but tested and looked at the code and it seems to work! We should know for sure once everyone loads this into their sandboxes and updates with main.

PR #1560, opened 2023-12-21
Issue #1352 - Show saved nameservers if registry is down

### LINKED ISSUE #1352: Show saved nameservers if registry is down
### Story

As a registrant
I want to see my nameservers even if the registry is down
so that I am not worried about missing nameservers and so that I don't see the "add nameservers" button on the overview page

### Acceptance Criteria

- [ ] nameservers and their ip address are saved in our database when received from epp
- [ ] anytime nameservers/ips is recieved from epp, update our database object accordingly including removing them
- [ ] if epp is down and a nameserver is saved for that domain, that is shown instead of nothing  (this should be done on the overview page and nameserver page)

### Additional Context

note we already have a nameserver and host_ip tables that could be used, please evaluate them and make any changes as needed.

### Issue Links

Relates to #1336 (Consider this in design decisions for handling this condition)

dave-kennedy-ecs:
## Ticket

Resolves #1352 

## Changes

- DNS Nameservers (host names and IPs) are stored in the registrar database when they are fetched from registry through EPP
- When registry is unavailable through EPP, DNS Nameservers are retrieved from the registrar database rather than from the registry through EPP

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


src/registrar/models/domain.py
```diff
@@ -1666,6 +1670,51 @@ def _update_hosts_and_contacts(self, cleaned, fetch_hosts, fetch_contacts):
...
+            cleaned: dict containing hosts.  Hosts are provided as a list of dicts, e.g.
+                [{"name": "ns1.example.com",}, {"name": "ns1.example.gov"}, "addrs": ["0.0.0.0"])]
+            fetch_hosts: boolean indicating whether or not fetch_hosts was called
+        """
+        if fetch_hosts:
+            cleaned_hosts = cleaned["hosts"]
+            # Get all existing hosts from the database for this domain
+            existing_hosts_in_db = Host.objects.filter(domain=self)
+            # Identify hosts to delete
+            cleaned_host_names = set(cleaned_host["name"] for cleaned_host in cleaned_hosts)
```
therealslimhsiehdy, line 1687: **Question:** Would you mind giving an example of when we would have duplicates of a host name? (Just trying to think through all edge cases)

therealslimhsiehdy (commented):
**On local:** Bc EPP doesn't work on local I was able to test out the scenario of retrieving from the registrar DB (via Django Admin > Host > Add new host) rather than from the registry through EPP
Screenshot 2023-12-28 at 4 02 19 PM

**On sandbox:** When I add in Host(s) via Django Admin and then make changes via the UI to these Hosts + IPs, I see the changes reflected in the Django Admin -- this means that the database is being updated alongside
Screenshot 2023-12-28 at 4 35 44 PM
Screenshot 2023-12-28 at 4 35 54 PM

**Question:** Still milling about the edge case of adding 1 in via Hosts in Django Admin and another via UI and it having the section bork where I can't even add in new Nameservers and IPs even when I deleted it from Hosts. 

Even if we set it it as readonly on Django Admin, wouldn't we still need to apply certain permission settings to add in Hosts (ie make sure they put 2 Nameservers in for 1 domain, but I feel like this may be tricky) to safeguard against unwanted data errors even for an Admin permission level user?

rachidatecs:
The doctstrings for the Host and HostIp classes need updating. They still say: "This model exists ONLY to allow a new registrant to draft DNS entries before their application is approved."

src/registrar/models/domain.py
dave-kennedy-ecs: There should not be a case of duplicates.  It is not allowed by the logic in the name server form; throws an error 'Remove duplicate entry'.  And it is not allowed by the registry.

dave-kennedy-ecs:
> The doctstrings for the Host and HostIp classes need updating. They still say: "This model exists ONLY to allow a new registrant to draft DNS entries before their application is approved."

Updated the docstrings

dave-kennedy-ecs:
> **On local:** Bc EPP doesn't work on local I was able to test out the scenario of retrieving from the registrar DB (via Django Admin > Host > Add new host) rather than from the registry through EPP Screenshot 2023-12-28 at 4 02 19 PM
> 
> **On sandbox:** When I add in Host(s) via Django Admin and then make changes via the UI to these Hosts + IPs, I see the changes reflected in the Django Admin -- this means that the database is being updated alongside Screenshot 2023-12-28 at 4 35 44 PM Screenshot 2023-12-28 at 4 35 54 PM
> 
> **Question:** Still milling about the edge case of adding 1 in via Hosts in Django Admin and another via UI and it having the section bork where I can't even add in new Nameservers and IPs even when I deleted it from Hosts.
> 
> Even if we set it it as readonly on Django Admin, wouldn't we still need to apply certain permission settings to add in Hosts (ie make sure they put 2 Nameservers in for 1 domain, but I feel like this may be tricky) to safeguard against unwanted data errors even for an Admin permission level user?

I have removed Hosts from django admin.  For two reasons:
1 - any changes made in django admin will never be propagated to registry, so allowing admins to update this data in django admin would be misleading
2 - the logic which enforces rules about valid name servers is not applied in django admin and allows for data which can break functionality for domain management in the app

src/registrar/models/domain.py
```diff
@@ -1666,6 +1670,51 @@ def _update_hosts_and_contacts(self, cleaned, fetch_hosts, fetch_contacts):
...
+    def _update_hosts_and_ips_in_db(self, cleaned, fetch_hosts):
+        """Update hosts and host_ips in database if retrieved from registry.
+
+        Parameters:
+            self: the domain to be updated with hosts and ips from cleaned
+            cleaned: dict containing hosts.  Hosts are provided as a list of dicts, e.g.
+                [{"name": "ns1.example.com",}, {"name": "ns1.example.gov"}, "addrs": ["0.0.0.0"])]
+            fetch_hosts: boolean indicating whether or not fetch_hosts was called
+        """
+        if fetch_hosts:
```
rachidatecs, line 1682: What do you think about putting this guy outside the method?

src/registrar/models/domain.py
```diff
@@ -1651,7 +1655,7 @@ def _get_dnssec_data(self, response_extensions):
         return dnssec_data
 
     def _update_hosts_and_contacts(self, cleaned, fetch_hosts, fetch_contacts):
-        """Capture and store old hosts and contacts from cache if they don't exist"""
+        """Capture and cache old hosts and contacts from cache if they don't exist in cleaned"""
```
rachidatecs, line 1658: """Update cache from old cache"""

Is this what the method's doing?

rachidatecs (approved):

src/registrar/models/domain.py
therealslimhsiehdy: +1 on this, but definitely non-blocking!

therealslimhsiehdy (approved):

src/registrar/models/domain.py
dave-kennedy-ecs: You're right, I am updating the comment before merging

src/registrar/models/domain.py
dave-kennedy-ecs: I agree.  Done

PR #1590, opened 2024-01-04
(Not on a sandbox) Ticket #1501: Delete button for DomainApplication records

### LINKED ISSUE #1501: Allow users to delete domain requests they've started or withdrawn
### Issue description

To respect user's privacy, we should enable the ability for users to delete domain applications before they have submitted them or after they have withdrawn them.

### Acceptance criteria

- [ ] domain managers can delete domain applications before they have been submitted
- [ ] domain managers can delete domain applications that were withdrawn
- [ ] Implement design as defined in the [Figma mockup](https://www.figma.com/file/aJbuDLJNNikqQObTuCNKQa/GetGov_Designs_ECS-Truss?type=design&node-id=2955%3A12612&mode=design&t=Xf9KhN638FAIhaiC-1) Click the "present" button at the top right to see the prototype. 
- [ ] ~Number new domain requests that do not have a requested domain (see #1482)~ this was deemed too difficult, the creation date will be shown instead

### Additional context

_No response_

### Links to other issues

Design ticket: #735 
Related to: #1482

zandercymatics:
## Ticket

Resolves #1501

## Changes

- Added some styles in _tables.scss for positioning, increased width of card on the home page
- Modified home.html to include the delete button
- Added unit tests in test_views.py
- Added a new delete view in application.py for DomainApplication w/ a has_permission override to forbid certain domains from being deleted
- Modified index.py to add a context for the modal button and if there are any applications which can be deleted
- Added an abstract permission view for DomainApplication
- Adds some logic for displaying multiple unfinished applications

## Context for reviewers
This ticket adds a delete button for DomainApplications on the home page, and some logic where when you start multiple new domain applications, they will be numbered rather than having duplicate names. 

Regarding the delete button, it is only visible if the domain is in the state of started or withdrawn, otherwise it cannot be deleted. 

## Setup
**For designers**
Everything you need should already exists on my sandbox - one domainapplication in `STARTED` and one in `WITHDRAWN`. If this is not the case (i.e. there is not delete button for any of the DomainApplications), then you will need to navigate to /admin and add yourself as a submitter to a DomainApplication with either of those states.

Afterwards, just follow the workflow defined in the figma.

**For developers**
**IMPORTANT:** Since this PR adds a scss file and makes some of those changes, you will likely need to do collectstatic if that local cache is not refreshing for you. 

This PR should be fairly simple to review. When testing locally, all you need to do is boot up the application and you will have an assortment to test with.

Afterwards, start two domain applications, but do not finish either of them. Note that they will be numbered...

On sandboxes, you should already have this, but if you do not - grab a few applications and set yourself as a submitter. Then, delete them.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [x] Checked behavior

  - [x] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

## Screenshots
image


src/registrar/assets/sass/_theme/_usa-modal.scss
```diff
@@ -0,0 +1,13 @@
...
+.usa-modal.wider-modal {
+    width: 40% !important;
+    max-width: 40% !important;
+}
+
+.usa-modal {
+    .usa-modal__heading.blue-header{
+        color: $dhs-blue;
+    }
+}
```
zandercymatics, line 13: This is the same color as the figma but its not consistent with the other modals in our app (consistency difference between that and the figma). Thoughts?

src/registrar/templates/home.html
```diff
@@ -41,7 +41,16 @@ <h2>Domains</h2>
...
+          <th 
+            scope="col" 
+            role="columnheader" 
+            {% if has_deletable_applications %} class="action-col-custom-width" {% else %} class="width-15" {% endif %}
+          >
+            <span class="usa-sr-only">Action</span>
+          </th>
+          {% if has_deletable_applications %}
+          <th></th>
+          {% endif %}
```
zandercymatics, line 53: Technically we don't need to add this, just up the width -- but this seems to work better with scaling. Its the most reliable solution I found for it so far, but I'd be open to other suggestions

src/registrar/tests/test_views.py
```diff
@@ -88,11 +88,92 @@ def test_home_lists_domain_applications(self):
...
+        draft_domain = DraftDomain.objects.create(name="igorville.gov")
+        for status in DomainApplication.ApplicationStatus:
+            if status not in [
+                DomainApplication.ApplicationStatus.STARTED,
+                DomainApplication.ApplicationStatus.WITHDRAWN,
+            ]:
+                with self.subTest(status=status):
+                    application = DomainApplication.objects.create(
+                        creator=self.user, requested_domain=draft_domain, status=status
+                    )
```
zandercymatics, line 159: This can be re-implemented to exclude the for loop but it doesn't make sense to do here in this context. This is based off of other unit tests using subTest which test on status as well (FSM checks)

src/registrar/assets/sass/_theme/_usa-modal.scss
gabydisarli: @zandercymatics sometimes I forget to update the color when pulling in the USWDS base modal component because the color is so similar to our .gov palette. Go ahead and match the color we have for our styles on this and any modals going forward.

src/registrar/templates/home.html
```diff
@@ -121,22 +142,56 @@ <h2>Domain requests</h2>
...
+                id="toggle-delete-domain-alert-{{ forloop.counter }}"
+                aria-labelledby="Are you sure you want to continue?"
+                aria-describedby="User will be removed"
+                data-force-action
+              >
+                <form method="POST" action="{% url "application-delete" pk=application.id %}">
+                  {% if application.requested_domain %}
+                    {% include 'includes/modal.html' with modal_heading="Are you sure you want to delete "|add:application.requested_domain.name|add:"?"  modal_description="This will remove the domain request from the .gov registrar. This action cannot be undone." modal_button=modal_button|safe %}
+                  {% else %}
+                    {% include 'includes/modal.html' with modal_heading="Are you sure you want to delete your domain request?" modal_description="This will remove the domain request from the .gov registrar. This action cannot be undone." modal_button=modal_button|safe %}
```
gabydisarli, line 188: @zandercymatics is this the language that'd show when the domain request has no domain name on it yet. Like in the case of a request name being "New domain request"?

zandercymatics: @gabydisarli Correct! You can test this yourself by starting a new domain request, then leaving midway through. Rather than what it is now (`Are you sure you want to delete your domain request?`) Would you like this to conform to the previous pattern defined in the figma? I.e.  `Are you sure you want to delete New domain request`? Or something like this with corrected grammar?

gabydisarli:
Some accessibility findings. I saw these all on Voiceover, I can walk through it with you if it's unclear.

- The delete button is called out as a button, where as the others are called out as a link. Not sure how we want to handle this but this is another area where I'm fuzzy in what is best practice, might be something else for @rachidatecs to consider. Should they be read out as buttons or links?
- For the other actions, Manage, Edit, VO reads out the corresponding request name, like "Edit <domainrequestname>". The Delete button does not. Can we add this in for delete, so it matches the others?
- When opening the modal with VO on, I'm read the entire contents of the modal twice.

src/registrar/templates/home.html
gabydisarli: @zandercymatics yes let's follow that pattern. 

Are you sure you want to delete `New domain request #`?

gabydisarli: @zandercymatics with the numbering of unnamed requests. Let's set the 1st one to be called "New domain request 1" instead of "New domain request, New domain request 2, ..."

src/registrar/assets/sass/_theme/_usa-modal.scss
gabydisarli: I double checked and I sometimes mistakenly use medium-50 when it should be dark-60. My bad!

zandercymatics: Not a problem!

gabydisarli (approved):
LGTM, thanks Zander!

dave-kennedy-ecs (commented):
I pulled branch and tested locally.  Functions as expected.  Nice feature that I will make use of for sure!  Seems in line with how to perform this in django.
I don't have a comment on the css, it generally looks good to me, although I typically defer to others on the team with regards to css.

Non-blocking comment: I tested for contacts which are attached to the domain application.  When an application is deleted, contacts can / will become orphaned.  Is this the expected behavior?  Or do we want to delete them if they are not attached to any other entities in the system?

Comment: If I have a list of several New domain requests, and I delete the one labeled, New domain request 2, the page reloads and still shows New domain request 2 (since the domain requests get numbered at page load).  Is this the expected behavior?  Similarly, if you edit an existing request with a number on it, got through some screens, and save and continue, when you come back to the list and edit the existing request with that same number, you are not guaranteed to get the same application.  The numbering is confusing me.

zandercymatics:
@dave-kennedy-ecs 
> When an application is deleted, contacts can / will become orphaned. Is this the expected behavior?

Great observation. I think this depends on the level of data integrity we want to maintain. If the contact is associated with a User we would not want to delete that, but otherwise you could make a solid argument for deleting this. This would be a good architectural discussion as it relates to the discussion we're having with the user/contact linkage

zandercymatics:
>  The delete order is weird and names aren't consistent. Is this the expected behavior?  The numbering is confusing me.

Both of the issues you mentioned are a result of this information being generated on page load, rather than saving that change in the DB. This is generally unavoidable given the structure of those records.

To remedy this, a new DraftDomain object would need to be created each time a "New domain request" is started, which would mean the DB would be populated with numerous DraftDomains of the name "New domain request 1" (etc...). This makes me think that maybe we should add another field to DraftDomain, and some sort of mixin or something that would iterate this other field based on how many "None" exist

src/registrar/assets/sass/_theme/_usa-modal.scss
```diff
@@ -0,0 +1,7 @@
+@use "cisa_colors" as *;
+@use "uswds-core" as *;
+
+.usa-modal.wider-modal {
```
rachidatecs, line 4: Did you try the USWDS large modal `usa-modal--lg`? 

I think that could work. Related - if you ever actually need to extend a base CSS class with a variant, I like the -- convention (usa-modal--wider)

src/registrar/assets/sass/_theme/_tables.scss
```diff
@@ -25,6 +25,17 @@
     color: color('primary-darker');
     padding-bottom: units(2px);
   }
+
+  th.action-col-custom-width {
+    width: 24.25%;
+  }
+
+  // Fix margins in mobile view
+  @media screen {
```
rachidatecs, line 34: Looks like this media query is missing a value. In any case, you should use the USWDS mixin (eg: `@include at-media('tablet')`)

src/registrar/assets/sass/_theme/_base.scss
```diff
@@ -125,3 +125,7 @@ abbr[title] {
 .flex-end {
   align-items: flex-end;
 }
+
+.remove-underline {
```
rachidatecs, line 129: Missing ending blank line. Also there is a USWDS .text-no-underline class (https://designsystem.digital.gov/utilities/text-styles/). You should use it. If it doesn't work because of a cascade conflict, override with eg a.text-no-underline

src/registrar/templates/home.html
rachidatecs: Why add this at all?

src/registrar/templates/home.html
```diff
@@ -41,7 +41,16 @@ <h2>Domains</h2>
           <th data-sortable scope="col" role="columnheader">Domain name</th>
           <th data-sortable scope="col" role="columnheader">Expires</th>
           <th data-sortable scope="col" role="columnheader">Status</th>
-          <th scope="col" role="columnheader"><span class="usa-sr-only">Action</span></th>
+          <th 
+            scope="col" 
+            role="columnheader" 
+            {% if has_deletable_applications %} class="action-col-custom-width" {% else %} class="width-15" {% endif %}
```
rachidatecs, line 47: Why is this needed?

rachidatecs (dismissed):
See CSS and markup comments

src/registrar/assets/sass/_theme/_usa-modal.scss
rachidatecs: Looking at this again, the content should work fine in a default modal. We're using those for similarity sized content

zandercymatics: usa-modal--lg doesn't match the figma in terms of content-height and some other padding things. The reasoning for the implementation here is that most domain names tend to wrap their content, making it harder to read 
(stuff like this-plea-tea.gov becomes this-
plea-tea.gov). I think this should be kept for consistency with the design, but I am open to discussion on it

That said, I like the `usa-modal--wider` naming convention quite a lot

src/registrar/templates/home.html
zandercymatics: By default, the "manage" button doesn't align with the manage button in the other table. This is especially apparent if the user doesn't have anything that they can delete, so there is just an empty column. ` has_deletable_applications` is a flag that switches between the default table display (as it was before this change w/o that column) and the new one. These two scenarios need different widths on the manage button, especially because it wasn't correctly aligned before this PR

src/registrar/assets/sass/_theme/_usa-modal.scss
```diff
@@ -0,0 +1,7 @@
...
+    width: 40% !important;
+    max-width: 40% !important;
```
therealslimhsiehdy, line 6: **Question:** I see that `!important` is used for overriding, and am curious what component you're overriding + reasoning for it? (Just trying to understand, not a "this is bad!" comment)

src/registrar/templates/home.html
```diff
@@ -80,6 +89,9 @@ <h2>Domains</h2>
               {% endif %}
             </a>
           </td>
+          {% if has_deletable_applications %}
+          <td></td>
+          {% endif %}
```
therealslimhsiehdy, line 94: **Question:** Is this just used to empty spacing?

src/registrar/views/index.py
```diff
@@ -1,21 +1,46 @@
...
         # the active applications table
-        context["domain_applications"] = applications.exclude(status="approved")
+        applications = DomainApplication.objects.filter(creator=request.user).exclude(status="approved")
+
+        # Adds display logic for empty domain requests
+        counter = 1
+        for application in applications:
+            if not application.requested_domain or not application.requested_domain.name:
+                application.requested_domain = DraftDomain(name=f"New domain request {counter}")
+                counter += 1
```
therealslimhsiehdy, line 21: **Question:** Is it worth testing this scenario of "user exits without putting in a URL yet and want to make sure naming shows up correctly"? Don't know if it's worth (I also don't see in the Figma a sketch), but thought it'd be worth a mention

therealslimhsiehdy (commented):
Screenshot 2024-01-10 at 12 59 48 PM
Screenshot 2024-01-10 at 2 18 39 PM

Left a few questions -- looks great in the mobile and desktop view (you can see draft domain naming of "New domain request 1" + delete button)

rachidatecs:
> I pulled branch and tested locally. Functions as expected. Nice feature that I will make use of for sure! Seems in line with how to perform this in django. I don't have a comment on the css, it generally looks good to me, although I typically defer to others on the team with regards to css.
> 
> Non-blocking comment: I tested for contacts which are attached to the domain application. When an application is deleted, contacts can / will become orphaned. Is this the expected behavior? Or do we want to delete them if they are not attached to any other entities in the system?
> 
> Comment: If I have a list of several New domain requests, and I delete the one labeled, New domain request 2, the page reloads and still shows New domain request 2 (since the domain requests get numbered at page load). Is this the expected behavior? Similarly, if you edit an existing request with a number on it, got through some screens, and save and continue, when you come back to the list and edit the existing request with that same number, you are not guaranteed to get the same application. The numbering is confusing me.

+ 1 on numbering

On cleaning up orphan users: Dave and I did that on the other contacts formset delete. We should at the very least keep track of this side effect for a future ticket and / or for when we eventually refactor contacts.

src/registrar/assets/sass/_theme/_usa-modal.scss
zandercymatics: This is for the modal width just to increase how much wiggle room we get on that. This ended up being removed after a conversation with Rachid yesterday

src/registrar/views/index.py
zandercymatics: Yeah that's a good point. I am going to refactor how this bit works so I'll include that there

src/registrar/models/draft_domain.py
```diff
@@ -20,3 +20,5 @@ def __str__(self) -> str:
         default=None,  # prevent saving without a value
         help_text="Fully qualified domain name",
     )
+
+    is_complete = models.BooleanField(default=True, help_text="Determines if this Draft is complete or not")
```
rachidatecs, line 24: I think we might me able to do without is_complete. Trick would be to refactor what we're passing to the modal and to give it a 'New domain request {{ application.created_at }}' - which may be tricky, but worth it.

src/registrar/templates/includes/modal.html
```diff
@@ -4,6 +4,10 @@
     <div class="usa-modal__main">
         <h2 class="usa-modal__heading" id="modal-1-heading">
             {{ modal_heading }}
+            {% if heading_value is not None %}
+            <br>
```
abroddrick, line 8: same comment here about using a br instead of handling it with styling

src/registrar/views/index.py
```diff
@@ -7,15 +7,53 @@ def index(request):
...
+def _get_applications(request):
+    """Given the current request,
+    get all DomainApplications that are associated with the UserDomainRole object.
+
+    Returns a tuple of all applications, and those that are deletable by the user.
+    """
+    # Let's exclude the approved applications since our
+    # domain_applications context will be used to populate
+    # the active applications table
+    applications = DomainApplication.objects.filter(creator=request.user).exclude(status="approved")
```
abroddrick, line 45:
```suggestion
    applications = DomainApplication.objects.filter(creator=request.user).exclude(status=DomainApplication.ApplicationStatus.APPROVED
```
please change this before merging to use the contant value

abroddrick (approved):
See some minor comments before merging

src/registrar/templates/home.html
```diff
@@ -122,22 +141,71 @@ <h2>Domain requests</h2>
           </td>
           <td data-label="Status">{{ application.get_status_display }}</td>
           <td>
-             {% if application.status == "started" or application.status == "action needed" or application.status == "withdrawn"  %}
+            {% if application.status == "started" or application.status == "action needed" or application.status == "withdrawn"  %}
```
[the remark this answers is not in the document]
zandercymatics: Great suggestion!

src/registrar/models/domain_application.py
```diff
@@ -631,6 +631,7 @@ def submit(self):
 
         # Update submission_date to today
         self.submission_date = timezone.now().date()
+
```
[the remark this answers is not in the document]
zandercymatics: Nope, nice find

src/registrar/templates/home.html
```diff
@@ -41,7 +41,17 @@ <h2>Domains</h2>
...
           <th data-sortable scope="col" role="columnheader">Expires</th>
           <th data-sortable scope="col" role="columnheader">Status</th>
-          <th scope="col" role="columnheader"><span class="usa-sr-only">Action</span></th>
+          <th 
+            scope="col" 
+            role="columnheader" 
+          >
+            <span class="usa-sr-only">Action</span>
+          </th>
+          {% comment %}
```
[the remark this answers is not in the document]
zandercymatics: Will remove this as it is likely it won't be needed, rather it'd probably be better to just do sizing directly

abroddrick (changes requested):
Sorry for approving before, I forgot to check that contacts were being deleted. That needs to happen here so we don't create orphaned contacts on our production environment.

gabydisarli (approved):
Looks great!

src/registrar/views/application.py
```diff
@@ -611,3 +619,87 @@ def get(self, *args, **kwargs):
...
+
+        # Check if the desired item still exists in the DB
+        if check_db:
+            ao = self._get_contacts_by_id([ao.id]).first()
+            submitter = self._get_contacts_by_id([submitter.id]).first()
+            other_contacts = self._get_contacts_by_id(other_contact_ids)
+
+        # Pair each contact with its related name
+        checked_contacts = [(ao, "authorizing_official"), (submitter, "submitted_applications")]
+        checked_contacts.extend((contact, "contact_applications") for contact in other_contacts)
```
zandercymatics, line 684: Not the most elegant thing ever but it works. Let me know if this needs changing

src/registrar/views/application.py
```diff
@@ -611,3 +619,87 @@ def get(self, *args, **kwargs):
...
+        Contact.objects.filter(id__in=contacts_to_delete, user=None).delete()
+
+        # After a delete occurs, do a second sweep on any returned duplicates.
+        # This determines if any of these three fields share a contact, which is used for
+        # the edge case where the same user may be an AO, and a submitter, for example.
+        if len(duplicates) > 0:
+            duplicates_to_delete, _ = self._get_orphaned_contacts(application)
+            Contact.objects.filter(id__in=duplicates_to_delete, user=None).delete()
+
+        return response
```
zandercymatics, line 664: This is being done at the form level, as the ticket which handled the joins was done at this level too. From my understanding, we are trying to avoid modifying model-level behavior where possible, and this could be labelled as falling under that triage.

For a future ticket (probably during some sort of refactor or reconciliation), I would recommend some sort of helper, utility, class, etc that handles orphaned records more generically. This can be done either at the model or form level, but that on its own is another beast (metaprogramming) - but it can be done. I do not think this should be done now, though. Only if we have a need for it

src/registrar/views/application.py
```diff
@@ -611,3 +619,87 @@ def get(self, *args, **kwargs):
...
+        other_contacts = list(application.other_contacts.all())
+        other_contact_ids = application.other_contacts.all().values_list("id", flat=True)
+
+        # Check if the desired item still exists in the DB
+        if check_db:
+            ao = self._get_contacts_by_id([ao.id]).first() if ao is not None else None
+            submitter = self._get_contacts_by_id([submitter.id]).first() if submitter is not None else None
+            other_contacts = self._get_contacts_by_id(other_contact_ids)
+
+        # Pair each contact with its related name
```
abroddrick, line 682:
```suggestion
        # Pair each contact with its db related name for use in checking if it has joins
```

src/registrar/views/application.py
```diff
@@ -611,3 +619,87 @@ def get(self, *args, **kwargs):
...
+        checked_contacts = [(ao, "authorizing_official"), (submitter, "submitted_applications")]
+        checked_contacts.extend((contact, "contact_applications") for contact in other_contacts)
```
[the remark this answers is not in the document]
abroddrick: Agreed but makes sense here, just would update the comment to make it easier to read why this is happening

src/registrar/views/application.py
```diff
@@ -611,3 +619,87 @@ def get(self, *args, **kwargs):
...
+        # This determines if any of these three fields share a contact, which is used for
+        # the edge case where the same user may be an AO, and a submitter, for example.
+        if len(duplicates) > 0:
+            duplicates_to_delete, _ = self._get_orphaned_contacts(application)
+            Contact.objects.filter(id__in=duplicates_to_delete, user=None).delete()
+
+        return response
+
+    def _get_orphaned_contacts(self, application: DomainApplication, check_db=True):
+        """Collects all orphaned contacts"""
```
abroddrick, line 667: **change before merging**: add info on what this returns since it's not initially clear

src/registrar/views/application.py
```diff
@@ -611,3 +619,87 @@ def get(self, *args, **kwargs):
...
+
+        return (contacts_to_delete, self._get_duplicates(checked_contacts))
+
+    def _get_contacts_by_id(self, contact_ids):
+        """Given a list of ids, grab contacts if it exists"""
+        contacts = Contact.objects.filter(id__in=contact_ids)
+        return contacts
+
+    def _get_duplicates(self, objects):
+        """Given a list of objects, return a list of which items were duplicates"""
```
abroddrick, line 698: **fix before merging**: this is only ever used for contacts, I would be more specific here this is for contacts, especially since you use the variable "contact" below. Also, why not just convert objects to a set and then back to a list? That will automatically remove duplicates in a list normally, though I have never tried with a query result so curious if you ran into an issue there.

src/registrar/tests/test_views.py
```diff
@@ -89,17 +89,229 @@ def setUp(self):
...
+        application_2.other_contacts.set([contact_shared])
+
+        # Ensure that igorville.gov exists on the page
+        home_page = self.client.get("/")
+        self.assertContains(home_page, "igorville.gov")
+
+        # Trigger the delete logic
+        response = self.client.post(reverse("application-delete", kwargs={"pk": application.pk}), follow=True)
+
+        self.assertNotContains(response, "igorville.gov")
```
abroddrick, line 239:
```suggestion
        # igorville is now deleted
        self.assertNotContains(response, "igorville.gov")
```

abroddrick (approved):
Looks good, thanks for the change. A couple minor suggestions and for sure remove the availability check change before merging. That really should be linked to the associated ticket in it's own PR

src/registrar/views/application.py
zandercymatics: The reason why I am not converting it to a set is because I actually want to get a list of duplicates. This is because duplicates let me detect an edge case:

A contact (which is not joined on any other field) is attached to both the authorizing_official field and the submitter field. When the delete is ran, it "looks" like this field is not an orphan (because it technically isn't) - but that isn't correct, because it will be after the delete.

This is basically a check to see if we run into that case, i.e. if the contact is shared among any of the fields on Application. If it is, there is a good chance that we hit this edge case - so we use this as a flag to determine if we should do a second join check or not

PR #1650, opened 2024-01-18
Ticket #1643 - Remove stale form errors on the check availablity page

### LINKED ISSUE #1643: Remove stale form error messages on the .gov domain request page
### Current Behavior

There is a specific situation where error messages would be inconsistent. For example, you can start off by triggering an empty field error message by hitting "Save", then trigger another error message by typing in an unavailable domain and clicking on "Check availability". This way both error messages will show up at the same time at the form-field level (see screenshot). This also happens in the alternative domains section as well. 

inconsistentErrors
alt_inconsistentErrors

### Expected Behavior

Instead of having two error messages right next to each other, it makes more sense to only have one at the form-field level or one at the error level. As explained by @neilmb 

> The fundamental technical issue here is that we are mixing together server-generated error messages that come from field validation on the “Save” request ALONGSIDE error messages generated in the browser by handling the click on the “Check Availability” page (and its background request to the availability API).

So we need to 
1. only choose to use the server-generated error message or the browser-generated error message, and not both
2. make sure the error message that is shown corresponds to the content that is _**currently**_ in the field, not what happened before

### Steps to Reproduce

1.Trigger one type of error by clicking on "Save and continue" button
2. Trigger a different type of error by clicking on "Check availability" button (in the Alternative Domains section you just need to type in a different type of invalid response)

### Environment

_No response_

### Additional Context

Slack thread: https://cisa-corp.slack.com/archives/C05BDEA3C11/p1705516070653769

### Issue Links

Related to #1523

zandercymatics:
## Ticket

Resolves #1643

## Changes

- Modifies the `get-gov.js` file to remove stale form level error messages when a text change occurs

## Context for reviewers
An edge-case was found where the form error and the fail message could drift out of sync because the form error only gets updated when the form is submitted, vs the check availability button (and api call) which update on a user event.

This PR simply removes stale form errors that are no longer in sync with the new content changes.

## Setup
1. Start a new domain request, and go to dotgov_domain
2. Enter an invalid value, like "-", for domain and alternate domain. Click the check availability button for good measure
3. Click "Save and continue"
4. Enter a dot for domain and alternate domain. Click the check availability button. As you do so, the form errors should disappear, but the error message should be retained
5. Try to save again. The form should return a valid error

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### DIFF
--- a/src/registrar/assets/js/get-gov.js
+++ b/src/registrar/assets/js/get-gov.js
@@ -245,13 +245,10 @@ function handleValidationClick(e) {
   const alternateDomainsInputs = document.querySelectorAll('[auto-validate]');
   if (alternateDomainsInputs) {
     for (const domainInput of alternateDomainsInputs){
-      // Only apply this logic to alternate domains input
-      if (domainInput.classList.contains('alternate-domain-input')){
-          domainInput.addEventListener('input', function() {
-              removeFormErrors(domainInput, true);
-            }
-          );
-      }
+      domainInput.addEventListener('input', function() {
+          removeFormErrors(domainInput, true);
+        }
+      );
     }
     }
 })();
--- a/src/registrar/assets/js/get-gov.js
+++ b/src/registrar/assets/js/get-gov.js
@@ -264,7 +264,6 @@ function removeFormErrors(input){
   let errorMessage = document.getElementById(`${input.id}__error-message`);
   if (errorMessage) {
     errorMessage.remove();
-    console.log("Error message removed")
   }else{
     return
   }
@@ -274,6 +273,7 @@ function removeFormErrors(input){
     input.classList.remove('usa-input--error');
   }
 
+  // Get the form label
   let label = document.querySelector(`label[for="${input.id}"]`);
   if (label) {
     label.classList.remove('usa-label--error');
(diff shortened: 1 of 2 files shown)


dave-kennedy-ecs (commented):
Pulled locally.  I am still getting a mixture of the two error messages
image

zandercymatics:
@dave-kennedy-ecs Hey Dave, looks like a caching issue. Usually `docker compose run app python manage.py collectstatic --no-input` works. Alternatively I can push it to a sandbox since it does that plus the gulp command which resets the context.

> If I understand the ACs correctly, we only want to use form error messages above the inputs, correct? 

I think this is worth clarifying with @kristinacyin, my interpretation was that we don't want to display both simultaneously - but I may be misunderstanding that entirely. We can do either way

dave-kennedy-ecs (approved):

therealslimhsiehdy (approved):

rachidatecs:
Looks like the code you wrote is integrated in 1500 (cc. @therealslimhsiehdy) where I suggested a refactor to simplify the code after we removed on input validation.

@zandercymatics take a look at my review there. If you agree that 1500 merges this code cleanly, this PR can be closed.

PR #1669, opened 2024-01-22
(On getgov-za) Ticket #1511 - CSV export takes a long time

### LINKED ISSUE #1511: Exporting any domain report (full metadata, current-full, -federal) takes a long time, 401s, and only generates on the second attempt
### Current Behavior

See title.

### Expected Behavior

Generating one of these reports shouldn't take multiple minutes, throw an error, or require a retry.

### Steps to Reproduce

1. In the domains table on manage.get.gov/admin, click on "Export all domain metadata", "Export current-full.csv". or "Export current-federal.csv" 
2. Wait ~3 mins for the browser to throw ERR_TOO_MANY_REDIRECTS, redirect to a 401 page
3. Navigate back to the page to re-request the report, which should arrive in roughly a minute

### Environment

Stable

### Additional Context

* The [current-full.csv I was able to retrieve](https://docs.google.com/spreadsheets/d/1STyDpPj3kx1kJFJ9gQgmLe9OUamNNs1h7c_n71Q3x_s/edit#gid=787357430) the second time. Much of the data is off.
* See [Slack thread](https://cisa-corp.slack.com/archives/C05BGB4L5NF/p1702609627618349).

### Issue Links

_No response_

zandercymatics:
## Ticket

Resolves #1511

## Changes

- Modified the `get_security_email` function (and related functions) with the boolean `skip_epp_call` for scenarios (like the exports) where we want to just pull from either cache or the db
- Modified `csv_export.py` with multiple performance improvements
  - Changed the query on `get_domain_infos` to first do a "select_related" call on domain and ao, then did .annotate to concatenate first and last name (since, for some reason, this was one of the most computationally expensive operations)
   - Changed `write_row` to `parse_row` and simplified some of the internal logic. Of note, security email is now using `get_security_email` to pull from cache
   - Moved many values into variables
- Modified `write_body` to use pagination and the "writerows" function. The intent of this is to reduce total memory usage which should decrease (or ideally eliminate) crashing 
- Change 2

## Context for reviewers
**NOTE:** This PR isn't currently on a sandbox
This PR aims to substantially improve the performance of the csv export buttons. This will affect all of those, as well as the domain growth report. These will all behave the same.

The observed slowdown for three reasons: 1. redundant DB calls, 2. inefficient DB calls, and 3. computationally heavy parsing logic.

I have addressed all three of those here. On testing on my sandbox (getgov-za), the load time dropped from 2-3 minutes down to ~2-4 seconds.

## Setup
1. Visit the admin dashboard on a sandbox with data, or use the migration scripts to load a lot of data
2. Click the export buttons and note the load time

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### DIFF
--- a/src/registrar/management/commands/utility/extra_transition_domain_helper.py
+++ b/src/registrar/management/commands/utility/extra_transition_domain_helper.py
@@ -182,8 +182,6 @@ class LoadExtraTransitionDomain:
                 # STEP 5: Parse creation and expiration data
                 updated_transition_domain = self.parse_creation_expiration_data(domain_name, transition_domain)
 
-                # Check if the instance has changed before saving
-                updated_transition_domain.save()
                 updated_transition_domains.append(updated_transition_domain)
                 logger.info(f"{TerminalColors.OKCYAN}" f"Successfully updated {domain_name}" f"{TerminalColors.ENDC}")
 
@@ -198,6 +196,28 @@ class LoadExtraTransitionDomain:
                     f"{TerminalColors.ENDC}"
                 )
                 failed_transition_domains.append(domain_name)
+        
+        updated_fields = [
+            "organization_name", 
+            "organization_type", 
+            "federal_type", 
+            "federal_agency", 
+            "first_name", 
+            "middle_name", 
+            "last_name", 
+            "email", 
+            "phone", 
+            "epp_creation_date", 
+            "epp_expiration_date",
+        ]
+
+        batch_size = 1000
+        # Create a Paginator object. Bulk_update on the full dataset
+        # is too memory intensive for our current app config, so we can chunk this data instead.
+        paginator = Paginator(updated_transition_domains, batch_size)
+        for page_num in paginator.page_range:
+            page = paginator.page(page_num)
+            TransitionDomain.objects.bulk_update(page.object_list, updated_fields)
 
         failed_count = len(failed_transition_domains)
         if failed_count == 0:
--- a/src/registrar/management/commands/utility/extra_transition_domain_helper.py
+++ b/src/registrar/management/commands/utility/extra_transition_domain_helper.py
@@ -196,18 +196,18 @@ class LoadExtraTransitionDomain:
                     f"{TerminalColors.ENDC}"
                 )
                 failed_transition_domains.append(domain_name)
-        
+
         updated_fields = [
-            "organization_name", 
-            "organization_type", 
-            "federal_type", 
-            "federal_agency", 
-            "first_name", 
-            "middle_name", 
-            "last_name", 
-            "email", 
-            "phone", 
-            "epp_creation_date", 
+            "organization_name",
+            "organization_type",
+            "federal_type",
+            "federal_agency",
+            "first_name",
+            "middle_name",
+            "last_name",
+            "email",
+            "phone",
+            "epp_creation_date",
             "epp_expiration_date",
         ]
 
--- a/src/registrar/models/domain.py
+++ b/src/registrar/models/domain.py
@@ -909,17 +909,11 @@ class Domain(TimeStampedModel, DomainHelper):
         """Time to renew. Not implemented."""
         raise NotImplementedError()
 
-    def get_security_email(self, skip_epp_call=False):
+    def get_security_email(self):
         logger.info("get_security_email-> getting the contact")
 
-        # If specified, skip the epp call outright.
-        # Otherwise, proceed as normal.
-        if skip_epp_call:
-            logger.info("get_security_email-> skipping epp call")
-            security = PublicContact.ContactTypeChoices.SECURITY
-            security_contact = self.generic_contact_getter(security, skip_epp_call)
-        else:
-            security_contact = self.security_contact
+        security = PublicContact.ContactTypeChoices.SECURITY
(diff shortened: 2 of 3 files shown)


src/registrar/models/domain.py
```diff
@@ -909,11 +909,22 @@ def renew(self):
...
+
+        # If specified, skip the epp call outright.
+        # Otherwise, proceed as normal.
+        if skip_epp_call:
+            logger.info("get_security_email-> skipping epp call")
+            security = PublicContact.ContactTypeChoices.SECURITY
+            security_contact = self.generic_contact_getter(security, skip_epp_call)
+        else:
+            security_contact = self.security_contact
+
```
zandercymatics, line 923: Reasoning for this change:

In the original csv_export ticket, a workaround was used to query the value of the security email, but not trigger the EPP call. This was inefficient as it meant we had to do a DB call everytime we wanted this value, rather than relying on a single one and pulling from that cached data.

Rather than doing that, we can just modify the behavior of this function to specify this. This allows us to selectively toggle that behavior for scenarios where we might not want it (for example, a csv export or a script). By default, it will always pull from EPP

src/registrar/management/commands/utility/extra_transition_domain_helper.py
```diff
@@ -198,6 +196,28 @@ def update_transition_domain_models(self):
...
+            "epp_expiration_date",
+        ]
+
+        batch_size = 1000
+        # Create a Paginator object. Bulk_update on the full dataset
+        # is too memory intensive for our current app config, so we can chunk this data instead.
+        paginator = Paginator(updated_transition_domains, batch_size)
+        for page_num in paginator.page_range:
+            page = paginator.page(page_num)
+            TransitionDomain.objects.bulk_update(page.object_list, updated_fields)
```
zandercymatics, line 220: Including this change in here since we often need to run the transfer_transition_domain script for datasets with these sorts of tickets. This just improves the load time of it. Not by a lot but it makes it less-so. Works the same as before.

therealslimhsiehdy (approved):
For 9K+ results I was able to download `current-full.csv` in 5-6 seconds! 

Left a few comments + I do see you have a linting check that didn't pass, but otherwise LGTM :)

src/registrar/management/commands/utility/extra_transition_domain_helper.py
```diff
@@ -182,8 +182,6 @@ def update_transition_domain_models(self):
                 # STEP 5: Parse creation and expiration data
                 updated_transition_domain = self.parse_creation_expiration_data(domain_name, transition_domain)
 
-                # Check if the instance has changed before saving
-                updated_transition_domain.save()
```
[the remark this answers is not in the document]
zandercymatics: .save() basically triggers a db call everytime its invoked. Rather than doing it a bunch of times individually, we can do a "bulk update" so it does all of that one time, rather than in 9k little pieces:
TransitionDomain.objects.bulk_update(page.object_list, updated_fields)

This improves performance by a lot because there is latency to contact the db, update the record, wait for a response, etc. Bundling all of that one time means we skip those steps, so it improves load time. The comment there is a bit misleading as .save() doesn't really check to see if its been changed, it just updates the record

src/registrar/models/domain.py
zandercymatics: It is gone!

src/registrar/management/commands/utility/extra_transition_domain_helper.py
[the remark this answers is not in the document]
abroddrick: why are we updating a transition domain item here at all though... we shouldn't really be touching the transition domain scripts now that we aren't transitioning.

src/registrar/models/domain.py
```diff
@@ -910,10 +910,15 @@ def renew(self):
...
-        if secContact is not None:
-            return secContact.email
+        logger.info("get_security_email-> getting the contact")
+
+        security = PublicContact.ContactTypeChoices.SECURITY
+        security_contact = self.generic_contact_getter(security)
+
+        # If we get a valid value for security_contact, pull its email
+        # Otherwise, just return nothing
+        if security_contact is not None and isinstance(security_contact, PublicContact):
```
abroddrick, line 920: **comment/non blocking** interesting why was the isinstance needed?

src/registrar/management/commands/utility/extra_transition_domain_helper.py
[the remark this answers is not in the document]
zandercymatics: @abroddrick I've worked on a lot of tickets recently that are essentially hotfixes for that data. The slowest part about developing them is the transition script, since it takes a long time to load. I can definitely remove this portion of it though

src/registrar/models/domain.py
zandercymatics: This PR originally had some changes to this particular function. As I was making those, I found that it was possible for this function to pass in the wrong type here. I don't think this is a problem we could encounter outside that though, its just a nice to have check

PR #1860, opened 2024-03-01
Issue #1859: Hot fix for metadata

### LINKED ISSUE #1859: The meta data report is not able to download on stable
### Current Behavior

The meta data report is not able to download on stable and results in an error

### Expected Behavior

The meta data report should download

### Steps to Reproduce

1.go to manage.get.gov/admin/registrar/domain
2. click the meta data report
3. see it takes for ever to load and throws an error

### Environment

stable

works fine on others

### Additional Context

Adding additional columns on the meta data report has resulted in a lot more computation time when it comes to our full data set. Look at the logic around columns and you will see some duplication that could be reduced. 

### Issue Links

Relates to #1840 and #1556

rachidatecs:
## Ticket

Resolves #1859

## Changes

- Refactor how we get the max domain managers count (on the domain that has the most)

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### DIFF
--- a/src/registrar/utility/csv_export.py
+++ b/src/registrar/utility/csv_export.py
@@ -45,7 +45,9 @@ def get_domain_infos(filter_condition, sort_fields):
     return domain_infos_cleaned
 
 
-def parse_row(columns, domain_info: DomainInformation, max_dm_count, security_emails_dict=None, get_domain_managers=False):
+def parse_row(
+    columns, domain_info: DomainInformation, max_dm_count, security_emails_dict=None, get_domain_managers=False
+):
     """Given a set of columns, generate a new row from cleaned column data"""
 
     # Domain should never be none when parsing this information
@@ -131,6 +133,7 @@ def _get_security_emails(sec_contact_ids):
             logger.warning("csv_export -> Domain was none for PublicContact")
 
     return security_emails_dict
+
 
 def write_csv(
     writer,
--- a/src/registrar/utility/csv_export.py
+++ b/src/registrar/utility/csv_export.py
@@ -187,7 +187,8 @@ def write_csv(
         total_body_rows.append(rows)
 
     update_columns_with_domain_managers(columns, max_dm_count)
-    write_header(writer, columns)
+    if should_write_header:
+        write_header(writer, columns)
     writer.writerows(total_body_rows)
 
 
@@ -220,7 +221,7 @@ def export_data_type_to_csv(csv_file):
     ]
     filter_condition = {
         "domain__state__in": [
-            Domain.State.UNKNOWN,
+            Domain.State.READY,
             Domain.State.DNS_NEEDED,
             Domain.State.ON_HOLD,
         ],
--- a/src/registrar/utility/csv_export.py
+++ b/src/registrar/utility/csv_export.py
@@ -184,9 +184,10 @@ def write_csv(
                 # It indicates that DomainInformation.domain is None.
                 logger.error("csv_export -> Error when parsing row, domain was None")
                 continue
-        total_body_rows.append(rows)
+        total_body_rows.extend(rows)
 
-    update_columns_with_domain_managers(columns, max_dm_count)
+    if get_domain_managers:
+        update_columns_with_domain_managers(columns, max_dm_count)
     if should_write_header:
         write_header(writer, columns)
     writer.writerows(total_body_rows)
--- a/src/registrar/utility/csv_export.py
+++ b/src/registrar/utility/csv_export.py
@@ -25,7 +25,7 @@ def write_header(writer, columns):
 
 def get_domain_infos(filter_condition, sort_fields):
     domain_infos = (
-        DomainInformation.objects.select_related("domain", "authorizing_official")
+        DomainInformation.objects.select_related("domain", "domain__permissions", "authorizing_official")
         .filter(**filter_condition)
         .order_by(*sort_fields)
     )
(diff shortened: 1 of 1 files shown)


src/registrar/utility/csv_export.py
```diff
@@ -161,19 +162,20 @@ def write_csv(
...
+    # We get the max so we can set the column header accurately
+    max_dm_count = 0
+    total_body_rows = []
 
     for page_num in paginator.page_range:
         rows = []
         page = paginator.page(page_num)
         for domain_info in page.object_list:
+            # Get count of all the domain managers for an account
+            dm_count = domain_info.domain.permissions.count()
```
therealslimhsiehdy, line 170: **Question:** Just to confirm -- because we're not doing the domain_info call outside, and bc we already have it and it's within the loop that helps the time intensiveness, correct?

src/registrar/utility/csv_export.py
```diff
@@ -182,8 +184,13 @@ def write_csv(
                 # It indicates that DomainInformation.domain is None.
                 logger.error("csv_export -> Error when parsing row, domain was None")
                 continue
+        total_body_rows.extend(rows)
```
therealslimhsiehdy, line 182: **Question:** To confirm -- we're also getting the row count and extending it once versus writing the row each time here, correct?

src/registrar/utility/csv_export.py
rachidatecs: That's the hope

src/registrar/utility/csv_export.py
```diff
@@ -215,7 +218,7 @@ def export_data_type_to_csv(csv_file):
     ]
     filter_condition = {
         "domain__state__in": [
-            Domain.State.READY,
+            Domain.State.UNKNOWN,
```
therealslimhsiehdy, line 221: Don't forget to remove this when you merge 😆

src/registrar/utility/csv_export.py
rachidatecs: Yeah. That's not a good thing. That's needed because we can't write body rows before we write the header. We can't write the header before we know max-dm. We can't know max_dm before we iterate over all DIs. We get the row data while iterating over DIs. So, we hold off on writing and store them all in a mega-list until the header is done.

src/registrar/utility/csv_export.py
rachidatecs: DUDE!

src/registrar/utility/csv_export.py
```diff
@@ -202,7 +205,7 @@ def export_data_type_to_csv(csv_file):
         "State",
         "AO",
         "AO email",
-        "Security contact email",
+        # "Security contact email",
```
therealslimhsiehdy, line 208: I'M SORRY ITS ME AGAIN I think for testing purposes you may have commented it out, but I think we want to keep the security contact email for this data collection (I might be missing context so lmk if I'm incorrect) 😅

src/registrar/utility/csv_export.py
```diff
@@ -44,7 +45,9 @@ def get_domain_infos(filter_condition, sort_fields):
     return domain_infos_cleaned
 
 
-def parse_row(columns, domain_info: DomainInformation, security_emails_dict=None, get_domain_managers=False):
+def parse_row(
+    columns, domain_info: DomainInformation, max_dm_count, security_emails_dict=None, get_domain_managers=False
+):
```
abroddrick, line 50: **non blocking suggestion**: this is for a section not part of your changes
This part below will just cause an unhelpful 500 error to show on screen. Fine for now as hopefully this should never happen but bad practice, we should at least be logging the error before throwing errors. 

Mind adding a log there? Not
```
 if domain_info.domain is None:
        raise ValueError("Domain is none")
```

src/registrar/utility/csv_export.py
```diff
@@ -94,6 +97,11 @@ def parse_row(columns, domain_info: DomainInformation, security_emails_dict=None
     }
 
     if get_domain_managers:
+        for i in range(1, max_dm_count + 1):
```
abroddrick, line 100: **non blockingsuggestion**
Hey so won't we ONLY need to add to this list if this check in write_csv actually enters and updates the max dm count values " if dm_count > max_dm_count:
                max_dm_count = dm_count"

Instead of always doing this call I would think just passing an update_domain_count bool could would save on looping through the columns. line 102 has O(n) complexity. Not ideal, but also this shouldn't hurt a ton given that list shouldn't be too big.

src/registrar/utility/csv_export.py
abroddrick: I think I made a very similiar suggestion to zander's here, but on the actual loop in question. and I think moving it here would save on having a bool passed to the other function.

src/registrar/utility/csv_export.py
```diff
@@ -161,29 +160,33 @@ def write_csv(
...
+
             try:
-                row = parse_row(columns, domain_info, security_emails_dict, get_domain_managers)
+                row = parse_row(columns, domain_info, max_dm_count, security_emails_dict, get_domain_managers)
                 rows.append(row)
             except ValueError:
                 # This should not happen. If it does, just skip this row.
                 # It indicates that DomainInformation.domain is None.
                 logger.error("csv_export -> Error when parsing row, domain was None")
                 continue
```
abroddrick, line 184: **comment** don't get why there is a continue here, python continues on excepted errors anyways

src/registrar/utility/csv_export.py
abroddrick: So there is a difference on writre row vs rows but it's dependent on the amount of data we have. Memory wise, yes storing all rows in memory is costly on memory but not on computation time. Whereas as writerow on large data sets is more expensive in time but less expensive in memory. Time is kind of our killer here and is really dependent on the large data set. A timer on the methods would require a large data set to get an accurate view, but the documentation online on row vs rows is a bit more enlightening. Writerows seems like the better call based on online searches

abroddrick (commented):
Almost done reviewing no red flags so far just want to share in case I don't finish before our next meeting

src/registrar/utility/csv_export.py
abroddrick: **must change before merging** I agree looking at staging the security email is supposed to be here. Add it back in before merging

abroddrick (approved):
This seems to work, I made several comments, only one of which is critical and needs to address before merging. Approving with the understanding that one will at least be changed OR you will tell me I am wrong. Consider the other ones, but not going to block a hotfix for them.

PR #2144, opened 2024-05-07
Ticket #1988: Investigate indexes

### LINKED ISSUE #1988: Investigate using Django indexes to speed up queries on tables commonly used for reports
### Issue description

Reports and accessing django admin pages is fairly slow. Django is designed with indexes in mind to make fast database queries, but we are not using them where we should and in some places not using them at all. These indexes are a way of making a quick reference based off of a model field for quick access. 

### Acceptance criteria

- [ ] add indexes in the meta section of models that we use often especially for reports. (at the very least domain name in domain request and domain should be indexed)
- [ ] remove unneeded indexes (currently they exist on zipcodes for sure, but check for others that are added inappropriately)
- [ ] test this on a sandbox with data that matches stable, run the migration and note in your PR any lag that occurred. This is important as if it brings down the system for a considerable time reduce the number of indexes. Convey your finding to the leads so any user impact is well known in advance. 

note with the last ac: if there is an impact we may need to make sure this is released to stable after working hours and apply a maintenance banner at the same time. This may be a non issue if the time to run the migration is non noticeable.

### Additional context

What are indexes in django?
Basically if it's a field you use a lot for filtering, like domain "name", then you should make an index for it. This considerable reduces the time it takes to find items when this particular filter is used. For reference, without the index it will run at a time complexity O(n), where n is the number of items in the table, but when indexed it will run at O(logn) which is noticeably faster.  

Why is this faster? 
django stores these values in a b-tree sort for fast retrieval

Why is there a downside then, can't everything be an index?
When creating the initial index b-tree, django locks the database during the migration. This is considered negligible for small datasets, but we have a large data set on stable. By adding indexes now we do risk unforseen issue when pushing to staging& stable. Hence, we just need to do this carefully. Indexes are also expensive if you're waisting them on non filtered fields, so we should only add them to things we filter by

What could use an index:
I would look to see what fields we are often filtering by in reports and for sure add an index for "name" column in domain request and domain. Note that these are two of our biggest data sets.

What shouldn't have an index:
any fields we don't usually filter by. Go through the code and find any that have db_index currently set and remove ones that are not filtered by often. Note as of now,  I see zipcode on domain request and domain information is a db_index, but we never filter by this value! We should remove this.

How:
django [documentation](https://docs.djangoproject.com/en/5.0/ref/models/fields/#db-index) says to avoid using db_index where possible and to use the model meta to set these values,[ see this documentation](https://docs.djangoproject.com/en/5.0/ref/models/options/#indexes)

Note: marked as blocked by 1989 so that we can be sure to test this on non-staging sandboxes

### Links to other issues

blocked by #1989

zandercymatics:
## Ticket

Resolves #1988

## Changes

- Removes db_index on all model properties and adds indexes to multiple of them

## Context for reviewers
This change adds indexes more strategically than our previous implementation. The intent is to improve performance during queries. However, doing so can increase the time it takes to complete migrations.

## Setup
On a sandbox with a lot of data, pull in this code and run migrations. It should run fine without any issues.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


src/registrar/admin.py
```diff
@@ -33,6 +33,7 @@
 
 from django.utils.translation import gettext_lazy as _
 
+
```
abroddrick, line 36: accident?

src/registrar/utility/csv_export.py
```diff
@@ -14,6 +14,7 @@
 from registrar.models.user_domain_role import UserDomainRole
 from registrar.utility.enums import DefaultEmail
 
+
```
abroddrick, line 17: accident?

rachidatecs:
Indices?

abroddrick (approved):

PR #2160, opened 2024-05-10
Small updates to our onboarding issue templates

h-m-f-t:
We sometimes have members of the public who _love_ .gov and attempt to go through our onboarding issues thinking it's something they need to do before we can contribute. This small change to the description of our role-centric issue templates aims to help random internet people know this isn't something they need to do.


abroddrick (approved):
Works for me!

PR #2219, opened 2024-05-24
Ticket #2154: enable/disable email sending

### LINKED ISSUE #2154: Add a way to turn off email sending in non-production envirornments
### Issue description

Email sending should always be on by default, however, there are times when we are running through tests in shared envirornments where users may perform activites which will result in unnecessary email sending. This is especially the case when a fixture user's email is the creator on a domain request and another user approves the  domain request. The risk of sending an accidental email to a coworker would be nice to avoid. 

We will also be providing a way to do and export from stable and into a test environment but we don't want emails accidentally sending to real users.

Thus in all non-production settings we should have the ability to turn sending emails on and off

### Acceptance criteria

- [ ] create an environment variable or feature flag for turning 
- [ ] add a few lines to documentation (readmes in the developer folder) to let other devs know how to turn this on/off and see if it's on (be brief)
- [ ] ensure this cannot be on in production (you can just check your feature flag/env variable along side our IS_PRODUCTION variable)

### Additional context

Pro of using feature flag is that a user just needs to check django admin to see if it's on and can use that to flip it on and off. This is pretty convenient on sandboxes that are shared as otherwise you would have to check what environment variables are turned on for that sandbox which is a slightly slower process. The choice is up to the developer here.

### Links to other issues

_No response_

zandercymatics:
## Ticket

Resolves #2154

## Changes

- Adds a switch called "disable_email_sending"
- Configure/use the built in "Switch" component from django-waffle

## Context for reviewers
This PR adds a WaffleSwitch (much like a waffle flag) that disables email sending when activated. The rationale for using a switch rather than a flag or an environment variable is [noted here](https://github.com/cisagov/manage.get.gov/issues/2154#issuecomment-2129716017). **However**, this can easily change if need be.

## Setup
To test this, first try sending an email normally. This can be done in many ways, but the fastest is usually by sending a domain invitation. This should work as it did before.

Afterwards, go to `/admin` (as a superuser) and click `WaffleSwitch`. Find the record listed as `disable_email_sending` and click it. On the record, click the enable checkbox.

Now, try to send out emails. All of them should fail, but still display an error message for the end user that it has failed.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### DIFF
--- a/src/registrar/utility/email.py
+++ b/src/registrar/utility/email.py
@@ -37,7 +37,7 @@ def send_templated_email(
     Raises EmailSendingError if SES client could not be accessed
     """
 
-    if flag_is_active(None, "disable_email_sending") and not settings.IS_PRODUCTION:  # noqa
+    if flag_is_active(None, "disable_email_sending") and not settings.IS_PRODUCTION:  # type: ignore
         message = "Could not send email. Email sending is disabled due to flag 'disable_email_sending'."
         raise EmailSendingError(message)
 


src/registrar/config/settings.py
```diff
@@ -329,10 +329,17 @@
 # If Waffle encounters a reference to a flag that is not in the database, should Waffle create the flag?
 WAFFLE_CREATE_MISSING_FLAGS = True
 
+# If Waffle encounters a reference to a switch that is not in the database, should Waffle create the switch?
```
abroddrick, line 332: remove the question mark here and on 329 and make these statements please.

src/registrar/utility/email.py
```diff
@@ -35,6 +36,10 @@ def send_templated_email(
 
     Raises EmailSendingError if SES client could not be accessed
     """
+    if switch_is_active("disable_email_sending") and not settings.IS_PRODUCTION:
+        message = "Could not send email. Email sending is disabled due to switch 'disable_email_sending'."
+        raise EmailSendingError(message)
```
abroddrick, line 41: Why are we raising an error here? we should just no-op here after logging that email sending is turned off, right?

abroddrick:
@zandercymatics I didn't want to use a switch here as we are increasing our tool base where we could still use a flag to do this same thing. This increases the training that people have to now know the difference between the two, when to use one vs the other, and now we have another table that will need content etc. While it may seem like the perfect case to use it, it adds additional overhead for future developers.

docs/developer/README.md
```diff
@@ -405,3 +405,11 @@ This function is triggered by the post_save event on the User model, designed to
 1. For New Users: Upon the creation of a new user, it checks for an existing `Contact` by email. If no matching contact is found, it creates a new Contact using the user's details from Login.gov. If a matching contact is found, it associates this contact with the user. In cases where multiple contacts with the same email exist, it logs a warning and associates the first contact found.
 
 2. For Existing Users: For users logging in subsequent times, the function ensures that any updates from Login.gov are applied to the associated User record. However, it does not alter any existing Contact records.
+
```
abroddrick, line 408: If we are adding in switches, and now increasing dev tools to use both, please add documentation on when to use one vs the other, it can link to other documentation from the source if you want. If this needs to be a follow on ticket that's fine too but please make that ticket

abroddrick:
@zandercymatics For waffle flag needing a request object, looking at the code it won't use the http response in every case, so I would first try just sending none as the request and then try an empty httprequest obejct without needing to pass anything. see how it's used in [this function](https://github.com/jazzband/django-waffle/blob/e63a13f87b328e9938e7459be3fd8b488226035d/waffle/models.py#L261). 
I don't think adding a none or an empty http response would really create tech debt, if those didn't work and you had to grab the current request that may be another story, but the documentation and stack overflow seem to indicate that sending empty http requests should work fine.

 I would say to at least attempt to use the current waffle tool before jumping to switches, but again I am not blocking on this as it's for development only. I would rather use the admin interface no matter what than an env variable.

CocoByte (approved):
LGTM!  Nice work on the testing portion.  I do like how you raise an error with a custom message if e-mail sending is disabled via the waffle flag.  Seems more complete and debug-friendly than not raising an error.

Approved.

PR #2344, opened 2024-06-20
Update domain_invitation_description.html to give better direction to analysts

h-m-f-t:
The table description can be improved to give analysts better context in-platform. This change makes an attempt.

### DIFF
--- a/src/registrar/templates/django/admin/includes/descriptions/domain_invitation_description.html
+++ b/src/registrar/templates/django/admin/includes/descriptions/domain_invitation_description.html
@@ -5,7 +5,7 @@ accept and become a domain manager.
 </p>
 
 <p>
-An “invited” status indicates that the recipient has not logged in to the registrar since the invitation was sent. Deleting an invitation with a "invited" status will prevent the user from signing in.
+An “invited” status indicates that the recipient has not logged in to the registrar since the invitation was sent. Deleting an invitation with an "invited" status will prevent the user from signing in.
 A “received” status indicates that the recipient has logged in. Deleting an invitation with a "received" status will stop a user's email address from showing in the domain managers list, but it will not revoke their access from the domain. To remove a user who has already signed in, go to <a class="text-underline" href="{% url 'admin:registrar_userdomainrole_changelist' %}">User domain roles</a> and delete the role for the correct  domain/manager combination.
 </p>
 
--- a/src/registrar/templates/django/admin/includes/descriptions/domain_invitation_description.html
+++ b/src/registrar/templates/django/admin/includes/descriptions/domain_invitation_description.html
@@ -5,8 +5,8 @@ accept and become a domain manager.
 </p>
 
 <p>
-An “invited” status indicates that the recipient has not logged in to the registrar since the invitation was sent. 
-A “received” status indicates that the recipient has logged in.
+An “invited” status indicates that the recipient has not logged in to the registrar since the invitation was sent. Deleting an invitation with a "invited" status will prevent the user from signing in.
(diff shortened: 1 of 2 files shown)


h-m-f-t:
cc: @Katherine-Osos, feedback welcome. When this lands, let's update the Google Doc.

h-m-f-t:
Thanks, Zander! I was definitely taking a stab in the dark and appreciate your feedback.

src/registrar/templates/django/admin/includes/descriptions/domain_invitation_description.html
```diff
@@ -5,8 +5,8 @@
 </p>
 
 <p>
-An “invited” status indicates that the recipient has not logged in to the registrar since the invitation was sent. 
-A “received” status indicates that the recipient has logged in.
+An “invited” status indicates that the recipient has not logged in to the registrar since the invitation was sent. Deleting an invitation with a "invited" status will prevent the user from signing in.
```
Katherine-Osos, line 8:
```suggestion
An “invited” status indicates that the recipient has not logged in to the registrar since the invitation was sent. Deleting an invitation with an "invited" status will prevent the user from signing in.
```

src/registrar/templates/django/admin/includes/descriptions/domain_invitation_description.html
```diff
@@ -5,8 +5,8 @@
...
+A “received” status indicates that the recipient has logged in. Deleting an invitation with a "received" status will stop a user's email address from showing in the domain managers list, but it will not revoke their access from the domain. To remove a user who has already signed in, go to <a class="text-underline" href="{% url 'admin:registrar_userdomainrole_changelist' %}">User domain roles</a> and delete the role for the correct  domain/manager combination.
```
Katherine-Osos, line 9:
```suggestion
A “received” status indicates that the recipient has logged in. Deleting an invitation with a "received" status will not revoke that user's access from the domain. To remove a user who has already signed in, go to <a class="text-underline" href="{% url 'admin:registrar_userdomainrole_changelist' %}">User domain roles</a> and delete the role for the correct  domain/manager combination.
```

Katherine-Osos: I just tested this out. If you delete an invitation that has been retrieved via the admin (which is only possible by superusers), the only visual change in the registrar is that—on the "domain managers" **page**—the user's email will no longer appear in the "Invitations" table. But that email will still appear in the "Domain managers" table. (There are two tables on the page once invitations have been sent.) 

That might be confusing to include within this description. So I suggest omitting that visual distinction from this description.

h-m-f-t: @Katherine-Osos I think the deleted text is still worthwhile. It's reasonable for someone to expect that removing an invitation would do both things (removing from the domain manager list _and_ revoking access). Revoking access is definitely more important, so it's good to be clearer there. Can you say more about why we should be silent on removing from the list?

Katherine-Osos: @h-m-f-t Sorry, my description was probably kind of confusing. If you delete a retrieved invitation via the admin ... that user will still show on the "domain manager" page within the registrar. The deleted text says it "_will stop a user's email address from showing in the domain managers list_" - which is not true. Were you actually saying it will be removed from the "domain invitations" table in the admin? If so, we should use that language, instead.

abroddrick: Agree with Katherine here. It could even call out too that "if you delete an invitation, be sure to remove their user domain role for this domain as well"

PR #2385, opened 2024-06-27
Update issue-default.yml to use * in links to other issues section

h-m-f-t:
Using a `*` before an issue makes GitHub display the issue details and whether it's open or closed. We like this! This change encourages its use in the issue template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue-default.yml
+++ b/.github/ISSUE_TEMPLATE/issue-default.yml
@@ -31,8 +31,8 @@ body:
     attributes:
       label: Links to other issues
       description: |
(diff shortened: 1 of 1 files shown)


abroddrick (approved):
I found checkboxes work too, and that's what I did with a few. It was also easier since I could highlight the numbers and just click the checkbox button. (Although, yes, only mildly easier). Any objection to this being checkboxes? 

Approving as I am not opposed to keep this in!

katypies:
I think any kind of list works, so probably can add the markdown, or use the buttons for any of the lists (I have no preference! I personally like having it expand, so whichever of the lists makes sense for a given issue is 👍 )

h-m-f-t:
@abroddrick just to confirm
-  [ ] <--- this?

PR #2393, opened 2024-06-28
Fix docker-compose race condition: ./manage.py load is ran before a cachetable can be created

zandercymatics:
## Ticket

None

## Changes

- This PR changes the order of our docker-compose.yaml commands locally, such that we run cachetable first, THEN we run load

## Context for reviewers
Recently, after updating my docker, I have been running into a race condition locally wherein the "load" command starts running before a cache table is created - causing a migration issue.

This tends to happen sporadically, but it doesn't hurt to fix the load order of these anyway.

## Screenshots
image

### DIFF
--- a/src/docker-compose.yml
+++ b/src/docker-compose.yml
@@ -67,8 +67,8 @@ services:
     # command: "python"
     command: >
       bash -c " python manage.py migrate &&
-      python manage.py load &&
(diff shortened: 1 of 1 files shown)


dave-kennedy-ecs (approved):
This is a good change.  LGTM

PR #2419, opened 2024-07-03
Ticket #1901: Allow analysts to customize action needed emails

### LINKED ISSUE #1901: Admin: Send customized "Action needed" emails
### Issue description

We've previously removed the functionality for sending out emails when a request status changes to "Action Needed". This was because the automated email template did not provide any helpful information on what "action(s)" the requester needs to do, such that the analysts have to reach out directly to the requester anyway.

Ideally, we want to still be able to send automated "Action Needed" emails through the admin portal, but customize the emails sent such that they can point to the exact places in the request form that need adjustment, and allow room for the analyst to make additional comments if necessary.

### Acceptance criteria

- [ ] Create automated email that are sent for requests with the "Action needed" status, making sure the email is customized based on the analysts' choice(s) in the list (implemented in #1900 

### Additional context

[Slack thread](https://cisa-corp.slack.com/archives/C05BDEA3C11/p1708033097789769) on topic

### Links to other issues

This is the last step in a three step process:

1. #1888 
2. #1900 
3. You are here 📍 (will need to wait until the above two steps are completed first)

zandercymatics:
## Ticket

Resolves #1901

## Changes

- Allows an analyst to edit the default action needed email that gets sent to the recipient
- Changed architecture minorly to accommodate for this

## Context for reviewers
This PR allows the end user to edit action needed emails after they've been populated in the auto generated email field.

An email will be sent out once an action needed reason is added or if the action needed reason is changed or if the email itself changes

## Setup
To test this, go to any domain request and change its state to action needed. You will see the "action needed email" field appear. From there, try changing the state normally: you should get the default email.

After that, try changing the action needed email and modifying the email.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### DIFF
--- a/src/registrar/assets/js/get-gov-admin.js
+++ b/src/registrar/assets/js/get-gov-admin.js
@@ -583,12 +583,20 @@ function initializeWidgetOnList(list, parentId) {
     function updateActionNeededEmailDisplay(reason) {
         let emailHasBeenSentBefore = sessionStorage.getItem(emailSentSessionVariableName) !== null;
         let collapseableDiv = readonlyView.querySelector(".collapse--dgsimple");
+        let showMoreButton = document.querySelector("#action_needed_reason_email__show_details");
         if ((reason && reason != "other") && !emailHasBeenSentBefore) {
             showElement(actionNeededEmail.parentElement)
             hideElement(readonlyView)
+            hideElement(showMoreButton)
         } else {
             if (!reason || reason === "other") {
                 collapseableDiv.innerHTML = reason ? "No email will be sent." : "-";
+                hideElement(showMoreButton)
+                if (collapseableDiv.classList.contains("collapsed")) {
+                    showMoreButton.click()
+                }
+            }else {
+                showElement(showMoreButton)
             }
             hideElement(actionNeededEmail.parentElement)
             showElement(readonlyView)
--- a/src/registrar/templates/django/admin/includes/detail_table_fieldset.html
+++ b/src/registrar/templates/django/admin/includes/detail_table_fieldset.html
@@ -149,7 +149,7 @@ This is using a custom implementation fieldset.html (see admin/fieldset.html)
         <div class="margin-top-05 collapse--dgsimple collapsed">
         {{ field.field.value|linebreaks }}
         </div>
-        <button type="button" class="collapse-toggle--dgsimple usa-button usa-button--unstyled margin-top-0 margin-bottom-1 margin-left-1">
+        <button id="action_needed_reason_email__show_details" type="button" class="collapse-toggle--dgsimple usa-button usa-button--unstyled margin-top-0 margin-bottom-1 margin-left-1">
             <span>Show details</span>
             <svg class="usa-icon" aria-hidden="true" focusable="false" role="img" width="24" height="24">
                 <use xlink:href="/public/img/sprite.svg#expand_more"></use>


Katherine-Osos:
@zandercymatics When you first select "action needed" - before selecting a reason - the auto-generated field reads: 

> "No email will be sent." 

That message makes sense when "other" is selected as the reason. But misleading if no selection has been made yet ... In that scenario, can you update the message to show a series of dashes (similar to what is shown in the "action-needed reason" dropdown when no selection is made).

Screenshot 2024-07-11 at 4 24 53 PM

Katherine-Osos:
@zandercymatics Although this wasn't captured in the AC, it was captured in the comments for the original ticket ... after the email is sent, the email content should be "locked" from editing. We want to prevent an analyst from accidentally editing after it's been sent ... which would send another email for that status/reason combo. However, if the user switches to another status/reason combo, then the related content should be editable.

Assuming the above update is ok ... can we also add help text below the field depending on whether the content is editable or not? 

Editable:
> This email will be automatically sent when you save your changes for this request.

Not editable:
> This email has already been sent. The text is no longer editable.

src/registrar/assets/js/get-gov-admin.js
```diff
@@ -527,10 +527,22 @@ function initializeWidgetOnList(list, parentId) {
...
+    const noEmailText = "No email will be sent."
     if(actionNeededReasonDropdown && actionNeededEmail) {
         // Add a change listener to the action needed reason dropdown 
         handleChangeActionNeededEmail(actionNeededReasonDropdown, actionNeededEmail);
+
+        document.addEventListener('DOMContentLoaded', function() {
+            if (!actionNeededReasonDropdown.value) {
+                noEmailMessage.innerHTML = emptyReasonText
+                showNoEmailMessage(actionNeededEmail, noEmailMessage);
+            }else if (actionNeededReasonDropdown.value == "other") {
```
rachidatecs, line 542: Nitpick
```suggestion
            } else if (actionNeededReasonDropdown.value == "other") {
```

src/registrar/assets/js/get-gov-admin.js
```diff
@@ -541,8 +553,11 @@ function initializeWidgetOnList(list, parentId) {
             // You also cannot save the model in this state.
             // This flow occurs if you switch back to the empty picker state.
             if(!reason) {
-                showNoEmailMessage(actionNeededEmail);
+                noEmailMessage.innerHTML = emptyReasonText
+                showNoEmailMessage(actionNeededEmail, noEmailMessage);
                 return;
+            }else if (reason === "other") {
```
rachidatecs, line 559: Nitpick: same nitpick here and below }else -> } else

src/registrar/assets/js/get-gov-admin.js
```diff
@@ -551,27 +566,25 @@ function initializeWidgetOnList(list, parentId) {
                 let emailBody = emailData.email_body_text
                 if (emailBody) {
                     actionNeededEmail.value = emailBody
-                    showActionNeededEmail(actionNeededEmail);
```
rachidatecs, line 554: simplify this block to 
if (emailData && emailBody) {
...
} else {
...
}

src/registrar/assets/js/get-gov-admin.js
```diff
@@ -551,27 +566,25 @@ function initializeWidgetOnList(list, parentId) {
...
+                showNoEmailMessage(actionNeededEmail, noEmailMessage);
             }
 
         });
     }
 
     // Show the text field. Hide the "no email" message.
-    function showActionNeededEmail(actionNeededEmail){
-        let noEmailMessage = document.getElementById("no-email-message");
+    function showActionNeededEmail(actionNeededEmail, noEmailMessage){
```
rachidatecs, line 581: Nitpick: I think it's more readable if you got rid of these 2 simple functions and just called the show/hide

src/registrar/models/domain_request.py
```diff
@@ -614,7 +614,8 @@ def save(self, *args, **kwargs):
 
         # Handle the action needed email. We send one when moving to action_needed,
         # but we don't send one when we are _already_ in the state and change the reason.
-        self.sync_action_needed_reason()
+        if self.status == self.DomainRequestStatus.ACTION_NEEDED and self.action_needed_reason:
```
rachidatecs, line 617: The comment above this is no longer accurate

src/registrar/models/domain_request.py
```diff
@@ -681,7 +683,7 @@ def delete_and_clean_up_domain(self, called_from):
             logger.error(f"Can't query an approved domain while attempting {called_from}")
 
     def _send_status_update_email(
-        self, new_status, email_template, email_template_subject, send_email=True, bcc_address="", wrap_email=False
+        self, new_status, email_template, email_template_subject, bcc_address="", context=None, **kwargs
```
rachidatecs, line 686: Nitpick:
I really think the extra layers of abstraction make your code harder to follow. I'd get rid of kwargs and just list the args

src/registrar/templates/django/admin/includes/detail_table_fieldset.html
```diff
@@ -143,6 +143,15 @@
     {% endwith %}
 {% endblock field_readonly %}
 
+{% block field_other %}
```
rachidatecs, line 146: Nitpick

```suggestion
{% block field_other %}
    {% if field.field.name == "action_needed_reason_email" %}
      {{ field.field }}
      <p id="no-email-message" class="{% if original_object.action_needed_reason %}display-none{% endif %}">No email will be sent.</p>
    {% else %}
      {{ field.field }}
    {% endif %}
{% endblock field_other %}
```

src/registrar/tests/test_admin.py
```diff
@@ -1479,6 +1484,29 @@ def test_action_needed_sends_reason_email_prod_bcc(self):
         # Should be unchanged from before
         self.assertEqual(len(self.mock_client.EMAILS_SENT), 4)
 
+        # Tests if an analyst can override existing email content
```
rachidatecs, line 1487: Pretty good, but since you have added a bunch of logic on WHEN to send emails, you should make sure the unit tests cover this (action_needed->action_needed and default email, other_status->action_needed and default email, action_needed->action_needed and custom email, other_status->action_needed and custom email)

rachidatecs (dismissed):

rachidatecs:
> > Assuming the above update is ok ... can we also add help text below the field depending on whether the content is editable or not?
> 
> As an alternative, what if we show the analysts a confirmation modal when they encounter that scenario you describe: when they try changing action needed email without changing the status first? That way they still have the option of sending a second email automatically if they choose, but a way to back out in that particular scenario.
> 
> This modal would not exist normally, only in that edge case.
> 
> Would this meet your criteria? This would be difficult to achieve without adding a db field - which we would want to avoid doing here.

Should be doable in JS methinks, just use cache?

rachidatecs:
> > Should be doable in JS methinks, just use cache?
> 
> Cache can get removed, no? Its the same issue with session variables.
> 
> I guess it depends, @Katherine-Osos are you OK with this field appearing as editable (when it shouldn't be) after browser cache gets cleared _or even when using another browser_? While cache exists and/or doesn't expire, it will appear as editable

Rachid Mrad (ECS, DC Area)
  11:18 AM
Cache can work
11:18
Maybe
11:18
Let's think it through
11:19
A user first lands on the page. He should be able to edit the fiield no matter what

Zander Adkinson (ECS, UT)
  [11:19](https://cisa-corp.slack.com/archives/D05MGHG1VNW/p1720797564852949) AM
Correct

Rachid Mrad (ECS, DC Area)
  11:20 AM
After the fiield is edited, we update cache, the filed becomes locked on refresh and on subsequent page visits

Zander Adkinson (ECS, UT)
  11:20 AM
This is true, but what if they use another browser or just clear their cache?

Rachid Mrad (ECS, DC Area)
  11:21 AM
Not an issue bro

Zander Adkinson (ECS, UT)
  11:21 AM
That doesn't persist though, right?

Rachid Mrad (ECS, DC Area)
  11:21 AM
Its an edge case, and the worst case scenario is that an email gets sent
11:21
It does not persist
11:22
So in a NEW session, the field is no longer locked... which is desireable, no?

src/registrar/assets/js/get-gov-admin.js
```diff
@@ -526,54 +526,72 @@ function initializeWidgetOnList(list, parentId) {
...
-        // Add a change listener to the action needed reason dropdown 
-        handleChangeActionNeededEmail(actionNeededReasonDropdown, actionNeededEmail);
-    }
-
-    function handleChangeActionNeededEmail(actionNeededReasonDropdown, actionNeededEmail) {
-        actionNeededReasonDropdown.addEventListener("change", function() {
+    // Since this is an iife, these vars will be removed from memory afterwards
+    var actionNeededReasonDropdown = document.querySelector("#id_action_needed_reason");
+    var actionNeededEmail = document.querySelector("#id_action_needed_reason_email");
+    var readonlyView = document.querySelector("#action-needed-reason-email-readonly");
```
rachidatecs, line 532: Non-blocking: Why toggle between 2 elements instead of setting /resetting value and disabled attr?

src/registrar/tests/test_admin.py
```diff
@@ -2243,6 +2289,7 @@ def test_user_sets_restricted_status_modal(self):
             self.assertContains(response, "Yes, select ineligible status")
 
     def test_readonly_when_restricted_creator(self):
+        self.maxDiff = None
```
rachidatecs, line 2292: Remove

src/registrar/tests/test_admin.py
```diff
@@ -2243,6 +2289,7 @@ def test_user_sets_restricted_status_modal(self):
...
         with less_console_noise():
```
rachidatecs, line 2293: Nitpick: Use the decorator

rachidatecs:
Non-blocking design issue:
Screenshot 2024-07-16 at 1 51 41 PM

rachidatecs (approved):

Katherine-Osos (approved):
LGTM!

PR #2538, opened 2024-08-05
#2322 Stop Election Office from unlocking early

### LINKED ISSUE #2322: Tribal gov step in domain request wizard gets skipped (shows as locked)
### Current Behavior

Tribal gov step in domain request wizard gets skipped (shows as locked) when it's the last unfilled step in an application you come to edit.

### Expected Behavior

Tribal gov step should not be locked when it is unfilled.

### Steps to Reproduce

1- Start an application
2- Select Tribal
3- Go to home page
4- Click edit on the application
5- See that Tribal Government has been skipped

### Environment

Stable
Staging

### Additional Context

Per Rachid: My guess is this is some bug in the logic that populates context for unlocked steps.

[Slack discussion](https://cisa-corp.slack.com/archives/C05BGB4L5NF/p1718068191072389)

### AC

- [ ] Solve this bug
- [ ] evaluate if any other conditional steps in our request flow are effected as well and ensure they are also fixed (break into a separate ticket if this balloons beyond the estimate for time to resolve)

### Issue Links

_No response_

Matt-Spence:
## Ticket

Resolves #2322 

## Changes
Removed a check in generic_helper that incorrectly assumes none values for is_election_board can only happen via manual tinkering. The combination of generic_org_type={Type which can be election org} + is_election_board=None can actually occur if only the generic org type was saved during the initial request creation (ie they backed out before saving the election yes/no). This normally doesn't matter much, it's effectively just a default no, but for instances like Tribal where there is a conditional form in between generic org and election org it makes election org become available early (because the None gets changed to False, which triggers the DB check that populates the sidebar with checks/locks).

- removed a check in generic_helper 
- Alter some tests that were built under the previous logic, and add new ones to make sure the new logic doesn't break things.

## Context for reviewers

Currently there is a bug that causes "Election Office" to unlock on the sidebar out of order. This is only really visible when selecting tribal for the generic org type, however. Steps to reproduce:

1. Create new request
2. Select Tribal
3. Save
4. Back out to the main page
5. Edit the request
6. Note that "Election Office" is unlocked on the sidebar even though it was not filled out.

Technically, election office is unlocking early for every org type (at least those that can have an election office). However, since it's the second step for most requests it doesn't look out of order. In practice it just looks like a default was populated and saved. However, it is noticeable on Tribal since that org type has a conditional form after saving the generic org.

Walk through the steps to reproduce, and note that election org no longer unlocks until all previous steps have been completed.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### DIFF
--- a/src/registrar/models/utility/generic_helper.py
+++ b/src/registrar/models/utility/generic_helper.py
@@ -245,7 +245,8 @@ class CreateOrUpdateOrganizationTypeHelper:
             if election_board_mismatch or org_type_mismatch:
                 message = (
                     "Cannot add organization_type and generic_org_type simultaneously "
-                    "when generic_org_type, is_election_board, and organization_type values do not match."
+                    "when generic_org_type ({}), is_election_board ({}), and organization_type ({}) values do not match."
+                    .format(generic_org_type, self.instance.is_election_board, organization_type)
                 )
                 logger.error("_validate_new_instance: %s", message)
                 raise ValueError(message)
--- a/src/registrar/models/utility/generic_helper.py
+++ b/src/registrar/models/utility/generic_helper.py
@@ -84,7 +84,6 @@ class CreateOrUpdateOrganizationTypeHelper:
         should_proceed = self._validate_new_instance()
         if not should_proceed:
             return None
-        
         # == Program flow will halt here if there is no reason to update == #
 
         # == Update the linked values == #
@@ -104,7 +103,6 @@ class CreateOrUpdateOrganizationTypeHelper:
     def _handle_existing_instance(self, force_update_when_no_changes_are_found=False):
         # == Init variables == #
         try:
-
             # Instance is already in the database, fetch its current state
             current_instance = self.sender.objects.get(id=self.instance.id)
 
--- a/src/registrar/models/utility/generic_helper.py
+++ b/src/registrar/models/utility/generic_helper.py
@@ -240,7 +240,7 @@ class CreateOrUpdateOrganizationTypeHelper:
             is_election_type = "_election" in organization_type
             can_have_election_board = organization_type in self.generic_org_to_org_map
 
-            election_board_mismatch = (is_election_type and (not self.instance.is_election_board or self.instance.is_election_board == None)) and can_have_election_board
+            election_board_mismatch = is_election_type and not self.instance.is_election_board and can_have_election_board
             org_type_mismatch = mapped_org_type is not None and (generic_org_type != mapped_org_type)
             if election_board_mismatch or org_type_mismatch:
                 message = (
--- a/src/registrar/models/utility/generic_helper.py
+++ b/src/registrar/models/utility/generic_helper.py
@@ -4,7 +4,6 @@ import time
 import logging
 from urllib.parse import urlparse, urlunparse, urlencode
 
-
 logger = logging.getLogger(__name__)
 
 
@@ -82,9 +81,13 @@ class CreateOrUpdateOrganizationTypeHelper:
 
     def _handle_new_instance(self):
         # == Check for invalid conditions before proceeding == #
-        should_proceed = self._validate_new_instance()
-        if not should_proceed:
+        try:
+            should_proceed = self._validate_new_instance()
+            if not should_proceed:
+                return None
+        except ValueError:
             return None
+        
         # == Program flow will halt here if there is no reason to update == #
 
         # == Update the linked values == #
@@ -104,6 +107,7 @@ class CreateOrUpdateOrganizationTypeHelper:
     def _handle_existing_instance(self, force_update_when_no_changes_are_found=False):
         # == Init variables == #
         try:
+
             # Instance is already in the database, fetch its current state
             current_instance = self.sender.objects.get(id=self.instance.id)
 
@@ -174,8 +178,8 @@ class CreateOrUpdateOrganizationTypeHelper:
             self.instance.organization_type = generic_org_type
         else:
             # This can only happen with manual data tinkering, which causes these to be out of sync.
-            if self.instance.is_election_board is None:
-                self.instance.is_election_board = False
+            # if self.instance.is_election_board is None:
+            #     self.instance.is_election_board = False
 
             if self.instance.is_election_board:
                 self.instance.organization_type = self.generic_org_to_org_map[generic_org_type]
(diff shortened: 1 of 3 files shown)


src/registrar/models/utility/generic_helper.py
```diff
@@ -249,6 +247,7 @@ def _validate_new_instance(self):
                     "Cannot add organization_type and generic_org_type simultaneously "
                     "when generic_org_type, is_election_board, and organization_type values do not match."
                 )
+                logger.error("_validate_new_instance: %s", message)
```
Matt-Spence, line 250: These aren't related to the issue, I just added them to help myself understand the code and thought it was worth leaving in.

src/registrar/models/utility/generic_helper.py
```diff
@@ -173,10 +172,6 @@ def _update_org_type_from_generic_org_and_election(self):
                 self.instance.is_election_board = None
             self.instance.organization_type = generic_org_type
         else:
-            # This can only happen with manual data tinkering, which causes these to be out of sync.
-            if self.instance.is_election_board is None:
-                self.instance.is_election_board = False
-
```
Matt-Spence, line 179: This turns out not to be true. If you back out after only filling out the generic org form, is_election_board will be None.

src/registrar/models/utility/generic_helper.py
abroddrick: For ease of troubleshooting if this occurs, I would print the values of these fields. Save a future dev the search time.

src/registrar/tests/test_models.py
```diff
@@ -1654,11 +1669,28 @@ def test_create_or_update_organization_type_existing_instance_updates_election_b
         self.assertEqual(domain_information.is_election_board, False)
         self.assertEqual(domain_information.organization_type, DomainRequest.OrgChoicesElectionOffice.CITY)
 
-        # Try reverting setting an invalid value for election board (should revert to False)
+    @less_console_noise_decorator
+    def test_create_or_update_organization_type_existing_instance_updates_election_board_to_none(self):
```
abroddrick, line 1673: These test names are very long, we may be competing with Java here. Can we shorten the name and add more into the comment. Great unit test comments follow the given/when/then story concept and while we don't enforce that as a must, those main points should ideally be covered in the comment if it's not immediately apparent. Here I don't think I would know exactly how election board would get updated to none using our system, if I hadn't seen the other comment you made.

src/registrar/models/utility/generic_helper.py
Matt-Spence: Good idea. Let me know if you like this formatting.

src/registrar/tests/test_models.py
```diff
@@ -1868,10 +1899,11 @@ def test_is_tribal_complete(self):
...
         self.domain_request.save()
         self.assertTrue(self.domain_request._is_tribal_complete())
-        self.domain_request.tribe_name = None
         self.domain_request.is_election_board = None
         self.domain_request.save()
-        # is_election_board will overwrite to False bc of _update_org_type_from_generic_org_and_election
+        self.assertFalse(self.domain_request._is_tribal_complete())
+        self.domain_request.tribe_name = None
+        self.domain_request.save()
         self.assertFalse(self.domain_request._is_tribal_complete())
```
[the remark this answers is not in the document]
Matt-Spence: Is_tribal_complete will fail if either tribe_name or is_election_board are None, but I didn't test the case you described because it's not possible with the normal request flow.

src/registrar/tests/test_models.py
```diff
@@ -1495,11 +1495,26 @@ def test_create_or_update_organization_type_existing_instance_updates_election_b
...
+    def test_create_or_update_organization_type_existing_instance_updates_election_board_to_none(self):
+        """Test create_or_update_organization_type for an existing instance."""
+        domain_request = completed_domain_request(
+            status=DomainRequest.DomainRequestStatus.STARTED,
+            name="started.gov",
+            generic_org_type=DomainRequest.OrganizationChoices.CITY,
+            is_election_board=False,
+        )
+        domain_request.is_election_board = True
+        domain_request.save()
```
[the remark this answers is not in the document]
Matt-Spence: That sounds like it would be testing the completed_domain_request function itself rather than the model. What would be the point of assigning is_election_board to false if the function would spit back a domain request where it updated to true before we even did anything?

Matt-Spence:
> Not to steal the thunder, I defer to Alysia on these things - but just thought I'd leave a few comments over a coffee break

All reviews are appreciated!

src/registrar/tests/test_models.py
Matt-Spence: Done. I also changed the initial is_election_office value to false, since that more accurately describes the situation where a request was only completed to the point of generic_org_type.

Matt-Spence: I made a few changes in addition to this after realizing that setting the initial value on creating a domain information object to False made it fail validation.

src/registrar/models/utility/generic_helper.py
```diff
@@ -219,12 +214,15 @@ def _update_generic_org_and_election_from_org_type(self):
...
-    def _validate_new_instance(self):
+    def _validate_new_instance(self) -> bool:
         """
         Validates whether a new instance of DomainRequest or DomainInformation can proceed with the update
         based on the consistency between organization_type, generic_org_type, and is_election_board.
 
         Returns a boolean determining if execution should proceed or not.
+
+        Raises: 
+            ValueError if there is a mismatch between organization_type, generic_org_type, and is_election_board
```
abroddrick, line 225: Love this!

src/registrar/models/utility/generic_helper.py
abroddrick: Unless I am missing it, I don't see you actually printing the values for those 3, just the message saying they don't match, did you mean to push a different change?

src/registrar/models/utility/generic_helper.py
```diff
@@ -242,13 +240,17 @@ def _validate_new_instance(self):
             is_election_type = "_election" in organization_type
             can_have_election_board = organization_type in self.generic_org_to_org_map
 
-            election_board_mismatch = (is_election_type != self.instance.is_election_board) and can_have_election_board
+            election_board_mismatch = (is_election_type and (not self.instance.is_election_board or self.instance.is_election_board == None)) and can_have_election_board
             org_type_mismatch = mapped_org_type is not None and (generic_org_type != mapped_org_type)
```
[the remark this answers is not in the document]
Matt-Spence: I did not know that, thanks for pointing it out!

src/registrar/models/utility/generic_helper.py
Matt-Spence: Check under the changes tab, it's showing up there but not under the conversation for whatever reason.

src/registrar/tests/test_models.py
Matt-Spence: @zandercymatics I accidentally typed False when I meant None here, I have updated the comment. Previously the assumption in validation was that is_election_board would always be a boolean value, hence is_election_type was compared directly to is_election_board. When the latter is None, that causes the check to always fail even though there are now valid states where is_election_board=None (even when the generic org type can technically have an election office).

PR #2609, opened 2024-08-16
Ticket #2586: Allow for multiple portfolios

### LINKED ISSUE #2586: Allow for multiple portfolios in the future
### Issue description

We will for sure be adding in the ability for users to have multiple portfolios sometime in the future. This will be post MVP, but it could be very soon post mvp. With all of the routing changes we just did for domain management pages, we should go ahead and make sure when we go to multiple portfolios we don't have to do a massive refactor (especially on a product that will be live at that point). 

When this feature is added fully into our system, users will have a way to toggle between between portfolios like toggling between slack spaces, or toggling between google accounts in google chrome. This portfolio selection will most likely be it's own page and, in fact, the first page that org users see when logging in. That said we will not be building those pages yet as design will be looking at how they want that later.

This means we need to roughly anticipate that there will be a method, after a user logs in, where they select a portfolio. Selecting the portfolio will bring them to what we currently have as the org model landing page. We then need to ensure that the User object can have multiple portfolios; yet at the same time still only allow for a single portfolio per user (as we do not yet have a toggle system in place). On the completion of this ticket, we shouldn't notice a visual change on the org model pages, nor should we see any change to the user flow there. We should however see the portfolio on a user be a list, and still block someone from adding more than one (we will remove the blocking logic later). 

### Acceptance criteria

- [ ] The user object now has portfolios not "portfolio", portfolios is now a many to many field connecting to the portfolio table
- [ ] a feature flag is added for "multiple_portfolios" and defaults to being turned off (that is just standard)
- [ ] On the User model object we still only allow for **saving** one portfolio total onto a user (suggest just handling in the validate function and check the feature flag first). Note, you should still be able to deselect a portfolio on the User and choose a different one, that should be unchanged.
- [ ] no change to where the org model pages land upon login, nor to the current user flow between pages
- [ ] all instances currently expecting portfolio to be singular are updated accordingly (this includes checks to see if a user has a portfolio or if retrieving the portfolio for a user)
- [ ] while the feature flag is on the portfolio is always the first one in the list. 

if going with the suggested approach below in Additional context ensure the following:

- [ ] last selected portfolio is add to the User ( suggest this->last_selected_portfolio = models.ForeignKey(Portfolio, on_delete=models.SET_NULL, null=True, blank=True, related_name='last_used_by') )
- [ ] on login be sure to check the waffle flag for multiple portfolios, if it is **off** and the user has at least one portfolio set  last selected portfolio to be the first portfolio value. If the portfolio flag is **on** then this value should get reset to None (to allow for the toggle feature)

### Additional context

One easy solution to this is to still keep the portfolio logic on the User and have a value on the user "last selected portfolio". This would then get reset on login to nothing. Once we have the toggle switch a user would log in, this value would automatically clear, then they would be directed to the toggle page at which point selecting a portfolio will set this value. Toggling again will change this value. For now though, we could still use a field "last selected portfolio" on User, default it to None, and then upon successful login (likely in the on each login function), we check a "multiple_portfolios" feature flag. if this flag is on **and** the user has at least one portfolio, we reset the "last selected portfolio" value to None,  if the flag is off, we set it to be the first portfolio. 

There is an alternative approach you could take where you expect the pk for the selected portfolio to be passed in once we have the toggle feature, and then you use this waffle flag to temporarily default this pk to the first value in their portfolios list. This gets around having to do another field but then gets complicated with ensuring that value is global or having to then pass this pk between ALL org views (which is a bigger lift). I don't recommend this one, but if a dev makes a case for this being better or has an alternative approach that is cleaner, changing how this is done is fair game. 

### Links to other issues

_No response_

zandercymatics:
## Ticket

Resolves #2586

Note: This is a collaboration between @dave-kennedy-ecs and @zandercymatics 

## Changes

- New data model, UserPortfolioPermission, for allowing multiple portfolios to be assigned to a User
- Current portfolio is stored in session
- When multiple portfolios, first portfolio is chosen for the user when they log in. (If this behavior is to change with future tickets, there are notes in this ticket where to change this.)
- When multiple_portfolios flag is not true for the User, only one portfolio can be added. If analyst attempts to add more than one, an error will be thrown in django admin

## Context for reviewers

Will want to test the new mechanisms in django admin for adding portfolio and roles/permissions to a User
Will want to test that multiple portfolios are not allowed to be added to a User when multiple_portfolios waffle flag is set to false
Will want to test that multiple portfolios are allowed to be added to a User when multiple_portfolios waffle flag is set to true
Will want to test that portfolio views work properly when organization waffle flag is set to true, and also that the views work properly when multiple_portfolios waffle flag is set to true

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### DIFF
--- a/src/registrar/models/user_portfolio_permission.py
+++ b/src/registrar/models/user_portfolio_permission.py
@@ -42,7 +42,7 @@ class UserPortfolioPermission(TimeStampedModel):
     user = models.ForeignKey(
         "registrar.User",
         null=False,
-        # when a portfolio is deleted, permissions are too
+        # when a user is deleted, permissions are too
         on_delete=models.CASCADE,
         related_name="portfolio_permissions",
     )


src/djangooidc/views.py
```diff
@@ -111,6 +111,10 @@ def login_callback(request):
             if not user.verification_type or is_fixture_user:
                 user.set_user_verification_type()
                 user.save()
+            
+            if not user.last_selected_portfolio:
```
abroddrick, line 115: Missing the waffle flag check here

src/registrar/models/user_portfolio_permission.py
```diff
@@ -0,0 +1,95 @@
...
+
+    portfolio = models.ForeignKey(
+        "registrar.Portfolio",
+        null=False,
+        # when a portfolio is deleted, permissions are too
+        on_delete=models.CASCADE,
+        related_name="portfolio_users",
+    )
+
+    portfolio_roles = ArrayField(
```
abroddrick, line 55: we don't need to repeat the name portfolio here and below since the table has moved

src/registrar/models/portfolio_invitation.py
```diff
@@ -87,9 +87,10 @@ def retrieve(self):
             raise RuntimeError("Cannot find the user to retrieve this portfolio invitation.")
 
         # and create a role for that user on this portfolio
-        user.portfolio = self.portfolio
+        user_portfolio = user.last_selected_portfolio
+        user_portfolio = self.portfolio
```
abroddrick, line 91: we do not want the last selected portfolio to change just because they accepted an invite, we want this to add a row in the UserPortfolioPermissions table

src/registrar/migrations/0119_remove_user_portfolio_and_more.py
```diff
@@ -0,0 +1,97 @@
...
+            name="portfolio",
+        ),
+        migrations.RemoveField(
+            model_name="user",
+            name="portfolio_additional_permissions",
+        ),
+        migrations.RemoveField(
+            model_name="user",
+            name="portfolio_roles",
+        ),
```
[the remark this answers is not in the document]
abroddrick: Thanks!

src/registrar/models/user_portfolio_permission.py
```diff
@@ -0,0 +1,119 @@
...
+    def clean(self):
+        """Extends clean method to perform additional validation, which can raise errors in django admin."""
+        super().clean()
+
+        # Check if a user is set without accessing the related object.
+        has_user = bool(self.user_id)
+        # Have to create a bogus request to set the user and pass to flag_is_active
+        request = HttpRequest()
+        request.user = self.user
+        if not flag_is_active(request, "multiple_portfolios") and self.pk is None and has_user:
```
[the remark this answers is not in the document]
abroddrick: Always support comments that say "bogus"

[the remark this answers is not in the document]
dave-kennedy-ecs: I may be dating myself with that comment!

src/registrar/admin.py
```diff
@@ -742,9 +743,6 @@ def overridden_email_field(self, obj):
                 "fields": (
                     "is_active",
                     "groups",
-                    "portfolio",
```
Matt-Spence, line 745: All of the admin changes work great, but thinking about it from an analyst perspective it seems inconvenient to no longer be able to view/edit anything about portfolios from within the user admin page. With these changes, if I want to see all portfolios a user is a part of the best I can do (far as I can tell at least) is go to the user_portfolio_permissions page, sort by user, and ctrl+F for the user I want. That's not the worst thing ever, but is there any easy way to make this more accessible?

src/registrar/models/user.py
```diff
@@ -230,68 +200,51 @@ def get_formatted_name(self):
...
+        if not portfolio:
             return False
 
-        portfolio_permissions = self._get_portfolio_permissions()
+        portfolio_perms = self.portfolio_permissions.filter(portfolio=portfolio, user=self).first()
+        if not portfolio_perms:
+            return False
 
+        portfolio_permissions = portfolio_perms._get_portfolio_permissions()
         return portfolio_permission in portfolio_permissions
```
Matt-Spence, line 214:
```suggestion
        user_portfolio_perms = self.portfolio_permissions.filter(portfolio=portfolio, user=self).first()
        if not user_portfolio_perms:
            return False

        return portfolio_permission in user_portfolio_perms._get_portfolio_permissions()
```
Nit-picky suggestion. I found these lines confusing since the names were so similar between portfolio_perms and portfolio_permissions.

Matt-Spence (approved):
A few very minor questions/suggestions, but all are optional so I'm approving.

src/registrar/admin.py
dave-kennedy-ecs: cc: @abroddrick I think this would need to be another ticket as it would increase the scope of this ticket quite a bit.  I think it is a good comment though worth considering.
We just did another ticket which lists the users in the portfolio page.  This would essentially be the opposite, listing the portfolios in the user page, right?

abroddrick: Yes agreed that 100 percent would need to be it's own ticket. On a related not, when looking at this message I noticed there wasn't a direct relationship between user and the portfolio permission table, such as "portfolios-(models.ManyToManyField("registrar.Portfolio",through="registrar.PortfolioPermission",
        related_name="portfolio_users",)" Which would allow us to get the portfolios for a user with just doing user.portfolios as opposed to any other filtering (and is slightly faster django wise). Was there a reason this was excluded. (I also looked very fast so forgive me if I missed it or something similar to it.

abroddrick: So time wise, the conversion to values list is slower for getting portfolios vs having a relation. Funny you point out UserDomainRoles though as that one is also connected on User with "domains". This was done for the same reason to allow for fast queries to get the relation and avoid the code of having to query the table and then get the series fitting "domains".
```
domains = models.ManyToManyField(
        "registrar.Domain",
        through="registrar.UserDomainRole",
        related_name="users",
    )
```

src/registrar/models/user_portfolio_permission.py
```diff
@@ -0,0 +1,119 @@
...
+        ],
+        UserPortfolioRoleChoices.ORGANIZATION_MEMBER: [
+            UserPortfolioPermissionChoices.VIEW_PORTFOLIO,
+        ],
+    }
+
+    user = models.ForeignKey(
+        "registrar.User",
+        null=False,
+        # when a portfolio is deleted, permissions are too
```
abroddrick, line 45: copy and paste error
```suggestion
        # when a user is deleted, permissions are too
```

src/registrar/admin.py
abroddrick: Okay side note though, I just saw that you have portfolio_permissions as the related name, but then also have a function with the same name and another function called get_portfolio_permissions. This made tracing the code rather rough, if I didn't know to look for one that was a relation, it would have taken a lot longer, I would highly encourage renaming the relation or renaming the functions so someone who is new to the codebase isn't confused on which of the multiple portfolio_permissions is being referred to.

abroddrick: Also to be clear I am not blocking on any of this, just asking nosey questions.

abroddrick: [ticket](https://github.com/cisagov/manage.get.gov/issues/2671) for adding portfolio back on the User. Keeping it simple for the first go, we can iterate on it after some more feedback from analysts

PR #2843, opened 2024-09-23
2589: RDAP API endpoint [rh]

### LINKED ISSUE #2589: Add RDAP data endpoint in manage.get.gov API
### Issue description

To enable users to view domain registration information, we are using WHOIS/RDAP to look up the registration data of a domain. However, GFE devices are blocked from accessing WHOIS due to being equipped with ZScaler, which blocks WHOIS. This prevents users on GFEs from accessing WHOIS data. 

Because manage.get.gov is serviced through cloud.gov, we will be adding a public API endpoint that acts as middleware to fetch WHOIS/RDAP data. The API endpoint can then be called on other .gov services without being blocked by ZScaler.

### Acceptance criteria

- [x] Create a public RDAP API endpoint  in the registrar that takes in a domain name and returns a JSON of its RDAP data
- [x] Append .gov if there is no TLD associated (going to interpret this as the input having no `.`) similar to how availability API appends .gov to inputs without TLD
~~- [ ] the api should provide an error in the json response if the domain ends in something other than .gov; however, it should except a domain with no suffix and it should append a .gov in this case (just like the availability api)~~
~~- [ ] the api  provide an error in the json response if the domain is not found~~
- [x] Add necessary tests for API endpoint functionality ~~make new class to mock basic RDAP responses~~
- [x] the api should be modeled much like the api for availability. suggest also utilizing  caching on the api endpoint such as with @ttl_cache(ttl=600) // Add a comment in ttl_cache that this may affect memory since it's storing this info in the system memory to keep in mind for memory issues down the road and we need to reevaluate cache time

### Additional context
Removed ACs above because RDAP returns JSON with error handling which is then parsed on get.gov's end

_No response_

### Links to other issues

blocks: [#316](https://github.com/cisagov/get.gov/issues/316)

erinysong:
## Ticket

Resolves #2589 

## Changes

- Adds public API endpoint `/rdap/?domain=` that fetches RDAP data of a domain from Cloudflare's RDAP API.
- Adds appropriate tests for RDAP API functionality.
- Remove `_domains` an artifact of old availability API that is no longer used

## Context for reviewers

Adding manage.get.gov API endpoint to fetch domain RDAP in order to mediate fetching Cloudflare data from other .gov projects.

## Setup

Go [https://getgov-rh.app.cloud.gov/api/v1/rdap/?domain=...](https://getgov-rh.app.cloud.gov/api/v1/rdap/?domain=whitehouse.gov) and replace ... with different domains. The API should respond to the following use cases:

- [ ] Valid RDAP domain names (eg `whitehouse.gov`) will output a JSON of the domain's RDAP data from Cloudflare.
- [ ] Valid RDAP domain names that exclude the .gov TLD (eg `whitehouse`) will output a JSOn of the domain's RDAP from Cloudflare.

Invalid domain names will be handled by Cloudflare's API handling and error codes which are then used on get.gov's end. 

- [ ] Domains that have a TLD (top-level-domain) that is not .gov will output JSON with the error message that the TLD does not exist under our registry.
- [ ] Domains that have more than one part above the TLD (manage.get.gov) will output JSON with the error message that domains must have exactly one part above their TLD. Open to discussing if we should follow Cloudflare's protocol here or if our use cases warrant handling it differently.
- [ ] Domain names that are not .gov domains (eg `hello.gov`) will output JSON with the error message that the .gov domain searched was not found.

Also confirmed that this API successfully displays RDAP JSON on GFEs and would like someone else to verify these steps work on their GFE.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### DIFF
--- a/src/api/tests/test_rdap.py
+++ b/src/api/tests/test_rdap.py
@@ -6,8 +6,9 @@ from django.contrib.auth import get_user_model
 from django.test import RequestFactory
 from django.test import TestCase
 
-from ..views import available, check_domain_available
+from ..views import available, check_domain_available, rdap
 from .common import less_console_noise
+from registrar.tests.common import MockRdapLib, MockEppLib
 from registrar.utility.errors import GenericError, GenericErrorCodes
 from unittest.mock import call
 
@@ -17,10 +18,42 @@ from epplibwrapper import (
 
 API_BASE_PATH = "/api/v1/rdap/?domain="
 
-class RdapAPITest(MockEppLib):
-     """Test that the RDAP API can be called as expected."""
 
-     def setUp(self):
+class RDapViewTest(MockRdapLib):
+    """Test that the RDAP view function works as expected"""
+
+    def setUp(self):
+        super().setUp()
+        self.user = get_user_model().objects.create(username="username")
+        self.factory = RequestFactory()
+
+    def test_rdap_get_no_tld(self):
+        """RDAP API successfully fetches RDAP for domain without a TLD"""
+        request = self.factory.get(API_BASE_PATH + "whitehouse")
+        request.user = self.user
+        response = rdap(request, domain="whitehouse")
+        # contains the right text
+        self.assertContains(response, "rdap")
+        # can be parsed into JSON with appropriate keys
+        response_object = json.loads(response.content)
+        self.assertIn("rdapConformance", response_object)
+
+    def test_rdap_invalid_domain(self):
+        """RDAP API accepts invalid domain queries and returns JSON response
+        with appropriate error codes"""
+        request = self.factory.get(API_BASE_PATH + "whitehouse.com")
+        request.user = self.user
+        response = rdap(request, domain="whitehouse.com")
+
+        self.assertContains(response, "errorCode")
+        response_object = json.loads(response.content)
+        self.assertIn("errorCode", response_object)
+
+
+class RdapAPITest(MockRdapLib):
+    """Test that the API can be called as expected."""
+
+    def setUp(self):
         super().setUp()
         username = "test_user"
         first_name = "First"
@@ -33,8 +66,10 @@ class RdapAPITest(MockEppLib):
         )
         
     def test_rdap_get(self):
+        """Can call RDAP API"""
         self.client.force_login(self.user)
         response = self.client.get(API_BASE_PATH + "whitehouse.gov")
-        self.assertContains(response, "RDAP")
+        self.assertContains(response, "rdap")
         response_object = json.loads(response.content)
-        self.assertIn("RDAP", response_object)
+        self.assertIn("rdapConformance", response_object)
+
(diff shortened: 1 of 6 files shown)


src/api/tests/test_rdap.py
```diff
@@ -0,0 +1,66 @@
...
+from django.contrib.auth import get_user_model
+from django.test import RequestFactory
+from django.test import TestCase
+
+from ..views import rdap
+
+API_BASE_PATH = "/api/v1/rdap/?domain="
+
+
+class RdapViewTest(TestCase):
```
erinysong, line 14: @abroddrick Since we don't call the RDAP API in any of manage.get.gov's views, I had our tests inherit from TestCase rather than create a mock RDAP library since it would go unused unless we find a reason to add RDAP views into the registrar. I can add a mock RDAP library class if we want to follow availability standards but it will likely go unused until we use RDAP in manage.get.gov and not just get.gov

src/api/views.py
```diff
@@ -19,6 +19,7 @@
 
 
 DOMAIN_FILE_URL = "https://raw.githubusercontent.com/cisagov/dotgov-data/main/current-full.csv"
```
abroddrick, line 21: Can you remove this and any uses of this while your in the code

src/api/tests/test_rdap.py
abroddrick: That makes sense

src/api/views.py
```diff
@@ -99,6 +100,22 @@ def available(request, domain=""):
     return json_response
 
 
+@require_http_methods(["GET"])
+@login_not_required
+# Since we cache domain RDAP data, cache time may need to be re-evaluated this if we encounter any memory issues
+@ttl_cache(ttl=600)
+def rdap(request, domain=""):
+    """Returns JSON dictionary of a domain's RDAP data from Cloudflare API"""
+    domain = request.GET.get("domain", "")
```
asaki222, line 109: overall lgtm! I was just wondering why the ttl is 600 seconds? Is that standard?

src/api/views.py
erinysong: @abroddrick sure! I'm not as familiar with where this is used but is current-full no longer referenced in manage.get.gov?

src/api/views.py
erinysong: I think generally 5 minutes is one of the standard defaults for caching which we also use in the availability endpoint! If there ends up being a cache time that makes more sense for this data we can also always update it if the need arises :")

asaki222: Okay, LGTM. 600 seconds is 10 minutes

asaki222 (approved):

src/api/views.py
erinysong: Ack you're right, did not math that math correctly :")

src/api/views.py
abroddrick: Nope it isn't

PR #2869, opened 2024-09-30
#2355: Custom rejection reason emails - [BACKUP] - MIGRATION

### LINKED ISSUE #2355: Dev: Display the text of the "Rejected" email associated with a sub-category before saving
### Issue description

In #1578, we allowed analysts to select a specific reason for rejection and for an email to be sent to a requestor. In this issue, let's display the text of the email template once the category is selected, letting analysts review the template (confirming it meets the need), and then send out an email.

### Acceptance criteria

- [ ] When analyst selects a "rejection reason," the text of the template is displayed inline (below the rejection reason) before the analyst clicks save to give them time to change it before the email sends

### Additional context

Reference the work done on this related ticket that exposed the content for "action needed" emails: https://github.com/cisagov/manage.get.gov/issues/2332

### Links to other issues

_No response_

zandercymatics:
## Ticket

Resolves #2355

## Changes

- Adds the ability to edit rejection reason emails, like what we're doing with action needed
- Consolidates the custom email logic to make it easier to change in the future + make additions

## Context for reviewers
This PR makes the rejection reason emails editable, like what we're doing with action needed. 
The only (current) difference between the two is that for rejection you can send emails out in the state "other". The only reason for this is that we had content for it. However, this can be changed if it is deemed that it needs changing.

As it turns out, the old logic for action needed email wasn't as flexible as we would have hoped. The infra had to be changed to accommodate for both rejection reason and action needed. This is the main reason for the line count

## Setup
1. Pick a domain request in django admin
2. Set the status to rejected
3. Send out an email, the email field should auto-update for each selection
4. After sending, note that the content has changed to denote this fact.
5. Pick another domain request and do the same thing - but this time edit the email

- [x] If any model was updated to modify/add/delete columns, makemigrations was ran and the associated migrations file has been commited.

- [x] (Rarely needed) Did dependency updates in Pipfile also get changed in requirements.txt?

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [x] Checked behavior

  - [ ] Chrome
  - [x] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### DIFF
--- a/src/registrar/config/urls.py
+++ b/src/registrar/config/urls.py
@@ -34,7 +34,7 @@ from registrar.views.utility.api_views import (
     get_rejection_email_for_user_json,
 )
 
-from registrar.views.domain_request import Step
+from registrar.views.domain_request import Step, PortfolioDomainRequestStep
 from registrar.views.transfer_user import TransferUserView
 from registrar.views.utility import always_404
 from api.views import available, rdap, get_current_federal, get_current_full
@@ -62,6 +62,9 @@ for step, view in [
     (Step.ADDITIONAL_DETAILS, views.AdditionalDetails),
     (Step.REQUIREMENTS, views.Requirements),
     (Step.REVIEW, views.Review),
+    # Portfolio steps
+    (PortfolioDomainRequestStep.REQUESTING_ENTITY, views.RequestingEntity),
+    (PortfolioDomainRequestStep.ADDITIONAL_DETAILS, views.PortfolioAdditionalDetails),
 ]:
     domain_request_urls.append(path(f"{step}/", view.as_view(), name=step))
 
@@ -190,7 +193,12 @@ urlpatterns = [
         name="export_data_type_requests",
     ),
     path(
-        "domain-request/<id>/edit/",
+        "reports/export_data_type_requests/",
+        ExportDataTypeRequests.as_view(),
+        name="export_data_type_requests",
+    ),
+    path(
+        "domain-request/<int:id>/edit/",
         views.DomainRequestWizard.as_view(),
         name=views.DomainRequestWizard.EDIT_URL_NAME,
     ),
--- a/src/registrar/forms/domain_request_wizard.py
+++ b/src/registrar/forms/domain_request_wizard.py
@@ -21,6 +21,13 @@ from registrar.utility.constants import BranchChoices
 logger = logging.getLogger(__name__)
 
 
+class RequestingEntityForm(RegistrarForm):
+    organization_name = forms.CharField(
+        label="Organization name",
+        error_messages={"required": "Enter the name of your organization."},
+    )
+
+
 class OrganizationTypeForm(RegistrarForm):
     generic_org_type = forms.ChoiceField(
         # use the long names in the domain request form
--- a/src/registrar/forms/utility/wizard_form_helper.py
+++ b/src/registrar/forms/utility/wizard_form_helper.py
@@ -279,11 +279,11 @@ class BaseYesNoForm(RegistrarForm):
         return initial_value
 
 
-def request_step_list(request_wizard):
+def request_step_list(request_wizard, step_enum):
     """Dynamically generated list of steps in the form wizard."""
     step_list = []
-    for step in request_wizard.StepEnum:
-        condition = request_wizard.WIZARD_CONDITIONS.get(step, True)
+    for step in step_enum:
+        condition = request_wizard.wizard_conditions.get(step, True)
         if callable(condition):
             condition = condition(request_wizard)
         if condition:
--- a/src/registrar/templates/domain_request_dotgov_domain.html
+++ b/src/registrar/templates/domain_request_dotgov_domain.html
@@ -12,8 +12,10 @@
 
   <p>Names that <em>uniquely apply to your organization</em> are likely to be approved over names that could also apply to other organizations. 
     {% if not is_federal %}In most instances, this requires including your state’s two-letter abbreviation.{% endif %}</p> 
-
+  
+  {% if not portfolio %}
   <p>Requests for your organization’s initials or an abbreviated name might not be approved, but we encourage you to request the name you want.</p> 
+  {% endif %}
 
   <p>Note that <strong>only federal agencies can request generic terms</strong> like
   vote.gov.</p>
--- a/src/registrar/templates/domain_request_intro.html
+++ b/src/registrar/templates/domain_request_intro.html
@@ -12,7 +12,11 @@
 
         <h1>You’re about to start your .gov domain request.</h1>
         <p>You don’t have to complete the process in one session. You can save what you enter and come back to it when you’re ready.</p>
+        {% if portfolio %}
+        <p>We’ll use the information you provide to verify your domain request meets our guidelines.</p>
+        {% else %}
         <p>We’ll use the information you provide to verify your organization’s eligibility for a .gov domain. We’ll also verify that the domain you request meets our guidelines.</p>
+        {% endif %}
         <h2>Time to complete the form</h2>
         <p>If you have <a href="{% public_site_url 'domains/before/#information-you%E2%80%99ll-need-to-complete-the-domain-request-form' %}" target="_blank" class="usa-link">all the information you need</a>,
           completing your domain request might take around 15 minutes.</p>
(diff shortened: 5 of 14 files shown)


src/registrar/models/domain_request.py
```diff
@@ -655,23 +661,62 @@ def save(self, *args, **kwargs):
...
+            return
+
+        # We should never send an email if no reason was specified.
+        # Additionally, Don't send out emails for reasons that shouldn't send them.
+        if status_info.get("reason") is None or status_info.get("reason") in status_info.get("excluded_reasons"):
+            return
+
+        # Only send out an email if the underlying reason itself changed or if no email was sent previously.
+        if status_info.get("cached_reason") != status_info.get("reason") or status_info.get("cached_reason") is None:
+            bcc_address = settings.DEFAULT_FROM_EMAIL if settings.IS_PRODUCTION else ""
```
rachidatecs, line 711: I feel like you missed an opportunity here

src/registrar/models/domain_request.py
```diff
@@ -655,23 +661,62 @@ def save(self, *args, **kwargs):
...
+            self._send_status_update_email(
+                new_status=status,
+                email_template="emails/includes/custom_email.txt",
+                email_template_subject="emails/status_change_subject.txt",
+                bcc_address=bcc_address,
+                custom_email_content=status_info.get("email"),
+                wrap_email=status_information.get("wrap_email"),
+            )
```
rachidatecs, line 719: OPTIONAL: Do we want an else block here that logs that send_custom_status_update_email met unexpected conditions?

src/registrar/models/domain_request.py
```diff
@@ -1040,17 +1065,12 @@ def reject(self):
         """Reject an domain request that has been submitted.
 
         As side effects this will delete the domain and domain_information
-        (will cascade), and send an email notification."""
+        (will cascade), and send an email notification using send_custom_status_update_email.
```
rachidatecs, line 1068: Match the commnent for action_needed, it's much clearer on where the email is being sent out from

src/registrar/models/domain_request.py
```diff
@@ -901,51 +946,31 @@ def in_review(self):
...
-        (will cascade) when they exist."""
+        (will cascade) when they exist.
+
+        Afterwards, we send out an email for action_needed in def save().
+        See the function send_custom_status_update_email.
+        """
 
         if self.status == self.DomainRequestStatus.APPROVED:
-            self.delete_and_clean_up_domain("reject_with_prejudice")
+            self.delete_and_clean_up_domain("action_needed")
```
rachidatecs, line 964: Good catch

src/registrar/templates/django/admin/includes/detail_table_fieldset.html
```diff
@@ -221,11 +221,99 @@ <h2 class="usa-modal__heading">
         </div>
         
         {% if original_object.action_needed_reason_email %}
-        <input id="last-sent-email-content" class="display-none" value="{{original_object.action_needed_reason_email}}">
+        <input id="last-sent-action-needed-email-content" class="display-none" value="{{original_object.action_needed_reason_email}}">
```
rachidatecs, line 224: NITPICK: Fix the tabbing

src/registrar/templates/django/admin/includes/detail_table_fieldset.html
```diff
@@ -221,11 +221,99 @@ <h2 class="usa-modal__heading">
...
+                data-close-modal
+                >
+                    <svg class="usa-icon" aria-hidden="true" focusable="false" role="img">
+                        <use xlink:href="{%static 'img/sprite.svg'%}#close"></use>
+                    </svg>
+                </button>
+            </div>
+        </div>
+        
+        {% if original_object.rejection_reason_email %}
```
rachidatecs, line 312: NITPICK: Tabbing

src/registrar/templates/django/admin/includes/detail_table_fieldset.html
```diff
@@ -221,11 +221,99 @@ <h2 class="usa-modal__heading">
...
+        <div class="margin-top-05 text-faded field-rejection_reason_email__placeholder">
+            &ndash;
+        </div>
+
+        {{ field.field }}
+
+        <button
+            aria-label="Edit email in textarea"
+            type="button"
+            class="usa-button usa-button--unstyled usa-button--dja-link-color usa-button__small-text margin-left-1 text-no-underline field-rejection_reason_email__edit flex-align-self-start"
```
rachidatecs, line 239: Since these buttons/ modals will never exist on the same page at the same time, I think there's an opportunity to significantly simplify the HTML and the JS by using generic handles (email-edit instead of field-rejection_reason_email__edit).

On the JS: You can now trim down config by moving a lot of the handles to the parent class

On the template: You can probably pull out the modals into an include that's a variant of modal.html, if not modal.html itself (that would be ideal)

It's not insignificant and your code has nothing wrong with it, so I won' block if you decide not to do it, but we should capture this plan for the next custom email ticket if it exists.

src/registrar/tests/test_admin_request.py
```diff
@@ -632,6 +637,7 @@ def assert_email_is_accurate(
         with less_console_noise():
             # Access the arguments passed to send_email
             call_args = self.mock_client.EMAILS_SENT
+            logger.info(f"what are the call args? {call_args}")
```
rachidatecs, line 640: BLOCKING: Remove

src/registrar/tests/test_admin_request.py
```diff
@@ -757,55 +773,53 @@ def test_action_needed_sends_reason_email_prod_bcc(self):
...
+        domain_request.save()
+        domain_request.refresh_from_db()
+
+    @override_settings(IS_PRODUCTION=True)
+    @less_console_noise_decorator
+    def test_rejected_sends_reason_email_prod_bcc(self):
+        """When a rejection reason is set, an email is sent out and help@get.gov
+        is BCC'd in production"""
+        # Create fake creator
+        EMAIL = "meoward.jones@igorville.gov"
```
rachidatecs, line 788: Q: Why caps? Good practice for const? I've rarely seen them in the project.

src/registrar/tests/test_models_requests.py
```diff
@@ -337,13 +337,6 @@ def test_withdraw_sends_email(self):
             domain_request, msg, "withdraw", 1, expected_content="withdrawn", expected_email=user.email
         )
 
-    @less_console_noise_decorator
```
rachidatecs, line 340: Q: This test should still work. Is that not the case? If so why?

src/registrar/utility/admin_helpers.py
```diff
@@ -2,35 +2,37 @@
...
+    )
+
+
+def get_rejection_reason_default_email(domain_request, rejection_reason):
+    """Returns the default email associated with the given rejection reason"""
+    return _get_default_email(
+        domain_request,
+        file_path="emails/status_change_rejected.txt",
+        reason=rejection_reason,
+        # excluded_reasons=[DomainRequest.RejectionReasons.OTHER]
```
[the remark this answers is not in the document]
rachidatecs: No opinion on this

rachidatecs:
500 Err on backup when I try to access a domain request

rachidatecs:
- Get to a in-review domain request as an admin
- Change the creator from eg Paul to Rachid
- Select a status with a custom email
- Note: The email text will still address Paul, but the email triggered on Save will go to Rachid. I do NOT think this shoud be addressed, just something to be aware of.

src/registrar/models/domain_request.py
rachidatecs: The function gets called. Inside the function, stuff happens only when certain conditions are right, otherwise it's just a ghost function silently running with no effect. I think we should log something to the effect of 'ran silently' which indicates an issue with our app or data

src/registrar/models/domain_request.py
rachidatecs: nvm

rachidatecs:
Found a bug: If you send out an action needed or rejected email, then switch out to the other state (eg, send action needed, then switch to rejected), session thinks you already sent out a rejection email and the edit button triggers a modal when it shouldn't.

rachidatecs (approved):

witha-c (approved):
Looks good. Adding a note here that I noticed a small part of the email content in the email footer was different from our content google docs and sent the list to @Katherine-Osos to decide whether to update the docs or the emails.

Issue #2775, opened 2024-09-18
Org Model: Update "view only" message on the domains page and fix tooltip


Katherine-Osos:
### Issue description

**Domain page**: We need to update the message that displays if a user can only view a domain, but not manage it. 

**Domains table**: We also need to fix the tooltip that appears when hovering over the "On Hold" status. The tooltip gets cut off when there are multiple domains in the table with different statuses. Note: the tooltip _doesn't_  seem to get cut off if the "On Hold" domain is the only one listed in the table.

### Acceptance criteria

- [x] On the domain page, update the message that appears if a user can only view a domain, but not manage it. Use the text in [this document](https://docs.google.com/document/d/1u-K1yQKb16H3AWj33mH2e1LJme6Ay9dNXMG8isiOXWk/edit#heading=h.tsr9qh4ofhbi).
- [x] In the domain table, fix the tooltip that appears when hovering over a domain in "on hold" status. The beginning of the tooltip gets cut off.

### Additional context

Beginning of tooltip gets cut off:
 
![Image](https://github.com/user-attachments/assets/7f70d59e-842f-4ec9-ad51-8ec522a36bdc)

rachidatecs:
@Katherine-Osos @abroddrick I just reviewed this ticket. The tooltip getting cut off is due to how the responsive table wrapper sets the overflow. This could prove to be really tricky to solve. A easier approach would be to shorten the tooltip text (could work, would need testing). Is that an option?

Katherine-Osos:
@rachidatecs @CocoByte All the words in that message are pretty important, but I think we only need to reduce a few characters, right? Can you try this (shortened by 3 characters)

> This domain is administratively paused. It can’t be edited and won’t resolve in DNS. Contact help@get.gov for details.

abroddrick:
@Katherine-Osos unfortunately, shortening it by 3 characters could still result in it being cut off on some screen sizes. To avoid it ever being cut off the ideal state would be for it to be just a few words, such as "needs name servers".

Katherine-Osos:
@abroddrick The tooltip resizes appropriately on most screensizes, it's just the version I posted above that gets cut off and it's two letters. Can a dev give it a try before we make any more adjustments?

vickyszuchin:
Moving this ticket from sprint 55 to the current sprint (sprint 56). cc: @abroddrick @lizpearl

PR #2894, opened 2024-10-03
#2275 - Org Model view only message and tooltip fixes - [AB]

CocoByte:
## Ticket [2275](https://github.com/cisagov/manage.get.gov/issues/2775)

Resolves #[2275](https://github.com/cisagov/manage.get.gov/issues/2775)

## Changes

- Updates the message that appears on the Domain page if a user can only view a domain, but not manage it. See the text in [this document](https://docs.google.com/document/d/1u-K1yQKb16H3AWj33mH2e1LJme6Ay9dNXMG8isiOXWk/edit#heading=h.tsr9qh4ofhbi).
- Fixes tooltips in the Domains table so that their content doesn't get clipped by parent containers (this is accomplished in two ways -- first, by adjusting wrapping logic so we can specify max widths to be less than the viewport width.  Secondly, by adjusting the positioning logic so the tooltip can show above its parent container(s).  Between these two mechanisms, we should be able to easily customize the tooltip to design needs)
- The above tooltip fix also adds a child element "usa-tooltip__content" to the tooltip.  This was needed for making the viewport adjustments for wrapping behavior.  But a handy side-effect of this is if design ever wants a fancy tooltip, we can absolutely make it as fancy as we want now by modifying this child element.

## Context for reviewers

NOTE @rachidatecs -- I remember you mentioning that it is preferable to use `<span>` within the tooltip.  However, using a `<span>` instead of a `<div>` (or other containers) causes conflicts in positioning.  `<span>` is getting handled differently and while I can see how using `<span>` would be the more minimal solution (especially for text content), `<div>` I think is more extensible/robust.  Perhaps I could add `<span>` within the `<div>` to keep text within appropriate containers...but this would *assume* purely text-based content (what if we wanted HTML instead?).  And since making assumptions in code is also breaking best practices, I opted to just use a `<div>` for simplicity and extensibility.

- [ ] Login as a user that can only view a domain, but not manage it.  Navigate to the domain page for that domain, and verify that the text is updated to "You don't have access to manage < domain >. If you need to make updates, contact one of the listed domain managers."
- [ ] Go back to the Domains table and verify that a various window widths, the tooltips do not have their content clipped.  (See image below, which shows a tooltip modified with lots of text to demonstrate wrapping)

image

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

- [ ] Checked behavior

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari


Katherine-Osos:
@CocoByte Since we won't have "member permissions" rolled out before people see this message ... can you remove the third sentence for now? (We'll need to adjust the message again after admins are incorporated into the system.) 

Therefore, it should read: 

> You don't have access to manage < domain >. If you need to make updates, contact one of the listed domain managers.

rachidatecs:
NOTE: You're linking to another PR instead of the ticket at the top of your PR description

src/registrar/assets/js/uswds-edited.js
```diff
@@ -6128,7 +6158,13 @@ const setUpAttributes = tooltipTrigger => {
   tooltipBody.setAttribute("aria-hidden", "true");
 
   // place the text in the tooltip
-  tooltipBody.textContent = tooltipContent;
+  // DOTGOV: nest the text element to allow us creater control over width and wrapping behavior
```
rachidatecs, line 6161:
```suggestion
  // DOTGOV: nest the text element to allow us greater control over width and wrapping behavior
```

src/registrar/assets/js/uswds-edited.js
```diff
@@ -5949,8 +5966,10 @@ const showToolTip = (tooltipBody, tooltipTrigger, position) => {
     const topMargin = calculateMarginOffset("top", e.offsetHeight, tooltipTrigger);
     const leftMargin = calculateMarginOffset("left", e.offsetWidth, tooltipTrigger);
     setPositionClass("top");
-    e.style.left = `50%`; // center the element
-    e.style.top = `-${TRIANGLE_SIZE}px`; // consider the pseudo element
+
+    // DOTGOV: updated logic for position:fixed
```
rachidatecs, line 5970: BLOCKING: For all changes in this file, I suggest you wrap them (even if it's one line) with DOTGOV EDIT and END DOTGOV EDIT comments. Leave the original USWDS code within these blocks, commented out.

We want to make it as easy as possible for devs who do not have context to update USWDS. To do so, they have to compile new releases and then go in and redo our edits onto the compiled code. I think my suggestion takes out some cognitive effort from this task.

src/registrar/assets/sass/_theme/_tooltips.scss
```diff
@@ -28,3 +28,46 @@
...
+    width: 50vw;
+    text-wrap: wrap;
+    text-align: center;
+    font-size: inherit; //inherit tooltip fontsize of .93rem
+    max-width: fit-content;
+    @include at-media('tablet') {
+      width: 70vw;
+    }
+  }
+}
```
rachidatecs, line 73: NITPICK: endline

src/registrar/assets/js/uswds-edited.js
```diff
@@ -6128,7 +6158,13 @@ const setUpAttributes = tooltipTrigger => {
   tooltipBody.setAttribute("aria-hidden", "true");
 
   // place the text in the tooltip
-  tooltipBody.textContent = tooltipContent;
+  // DOTGOV: nest the text element to allow us creater control over width and wrapping behavior
+   tooltipBody.innerHTML = `
+    <div class="usa-tooltip__content">
```
rachidatecs, line 6163: The div inside a span does not seem to trigger accessibility check errors, but it goes counter to HTML specifications. A div is a block level container, a span is an inline-level container span[https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span]

CSS-wise, you can easily make a span behave like a div in the layout by setting display: block on it. 

Q: Is there something else that's keeping you from using a span with a display block style on?

rachidatecs:
NOTE: Code is no longer on NL. Deploy it somewhere else for the design review.

src/registrar/assets/sass/_theme/_tooltips.scss
```diff
@@ -28,3 +28,46 @@
...
+  /*
+  Add width and wrapping to tooltip content in order to confine it to a smaller viewport window.
+  */
+  .usa-tooltip__content {
+    width: 50vw;
+    text-wrap: wrap;
+    text-align: center;
+    font-size: inherit; //inherit tooltip fontsize of .93rem
+    max-width: fit-content;
+    @include at-media('tablet') {
```
rachidatecs, line 69: Suggest changing this media query from tablet to desktop. It'll just be a bit more stable. See screenshot:

Screenshot 2024-10-09 at 7 31 07 PM

rachidatecs:
Fantastic work.

src/registrar/assets/js/uswds-edited.js
CocoByte: Totally agree.  I like this suggestion.

src/registrar/assets/js/uswds-edited.js
CocoByte: I see your point, and I get nesting divs inside spans is generally a no-no.   However, the element is set to display:block already when it becomes visible (.usa-tooltip__body.is-set), yet there is something else going on that positions the tooltip incorrectly above the tooltip icon when we use `<span>`.  We might want to huddle on this.

rachidatecs (approved):

PR #3196, opened 2024-12-06
Create sub-issue issue template

h-m-f-t:
This PR adds a "sub-issue" issue template, modeled on our current template for "parent" issues.

### DIFF
--- a/.github/ISSUE_TEMPLATE/sub-issue.yml
+++ b/.github/ISSUE_TEMPLATE/sub-issue.yml
@@ -1,0 +1,36 @@
+name: Sub-issue
+description: Describe an idea, problem, or feature that is related to a parent issue.
+
+body:
+  - type: markdown
+    id: title-help
+    attributes:
+      value: |
+        > Titles should be short, descriptive, and compelling. Use sentence case (don't capitalize unnecessarily).
+  - type: textarea
+    id: description
+    attributes:
+      label: Description
+      description: |
+        Describe the issue so that someone who wasn't present for its discovery can understand why it matters. Use full sentences, plain language, and [good formatting](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax).
+
+        For stories, use the user story format (e.g., As a user, I want, So that).
+    validations:
+      required: true
+  - type: textarea
+    id: acceptance-criteria
+    attributes:
+      label: Acceptance criteria
+      description: "If known, share 1-3 statements that would need to be true for this issue to be considered resolved. Use a [task list](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/about-task-lists#creating-task-lists) if appropriate."
+      placeholder: "- [ ]"
+  - type: textarea
+    id: additional-context
+    attributes:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/sub-issue.yml
```diff
@@ -0,0 +1,36 @@
+name: Sub-issue
+description: Describe an idea, problem, or feature that is related to a parent issue.
+
+body:
+  - type: markdown
+    id: title-help
+    attributes:
+      value: |
+        > Titles should be short, descriptive, and compelling. Use sentence case (don't capitalize unnecessarily).
```
gabydisarli, line 9: Can we adjust the label of the "Add a title" field. I'd suggest "Sub-issue title"

.github/ISSUE_TEMPLATE/sub-issue.yml
```diff
@@ -0,0 +1,36 @@
...
+  - type: textarea
+    id: description
+    attributes:
+      label: Description
```
gabydisarli, line 13: Suggesting: "Sub-issue description"

.github/ISSUE_TEMPLATE/sub-issue.yml
```diff
@@ -0,0 +1,36 @@
...
+        Describe the issue so that someone who wasn't present for its discovery can understand why it matters. Use full sentences, plain language, and [good formatting](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax).
+
+        For stories, use the user story format (e.g., As a user, I want, So that).
+    validations:
+      required: true
+  - type: textarea
+    id: acceptance-criteria
+    attributes:
+      label: Acceptance criteria
+      description: "If known, share 1-3 statements that would need to be true for this issue to be considered resolved. Use a [task list](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/about-task-lists#creating-task-lists) if appropriate."
```
gabydisarli, line 24: Are the quotations required here?

gabydisarli (approved):
Just a couple of optional suggestions here and there. Looks great!

.github/ISSUE_TEMPLATE/sub-issue.yml
h-m-f-t: Are you suggesting we change the `id` value or text in the `value`?

.github/ISSUE_TEMPLATE/sub-issue.yml
h-m-f-t: I think so; they are present in the Issue template (and the marks don't show in the actual issue). We'll find out!

.github/ISSUE_TEMPLATE/sub-issue.yml
gabydisarli: I'm not sure with it would be. It's the label for the field which currently says "Add a title" not sure if that's even editable though.

h-m-f-t: Ah. I looked and that doesn't appear to be part of the template; it's GitHub-controlled.

Issue #2960, opened 2024-10-18
Dev: Make updates to domain invitation emails


Katherine-Osos:
### Story

As a recipient of a domain invitation
I want the system to recognize whether I'm invited to multiple domains and if I'm already a user of the registrar
so that I'm only sent relevant information.

### Acceptance Criteria
Make the following updates to the domain invitation email (applies to org model AND non-org model):
- [x] Display dynamic text based on whether someone was invited to a single domain or multiple domains.
- [x] In the greeting, display the user's name if they're already a "user" in our system.  
  - [x] Simply display "Hi." if not a user in the system.
- [x] Display the section about needing a Login.gov account only if the invitee is not a “user” in our system yet.
- [x] Ensure the email reflects [this updated content](https://docs.google.com/document/d/18eQRA5QWu5Ct6ICK2-FCZJf9kbbk4LklekRRfb6g9tQ/edit?tab=t.0#heading=h.gjdgxs)

When invitations are sent for domains that are part of Org Model, double check that this AC from previous ticket #3019 is still intact
- [x]  If the invitee is not part of the portfolio yet, send the [portfolio invitation email ](https://github.com/cisagov/manage.get.gov/issues/3019)AND the domain invitation email.

- [x] The above ACs apply to domain invitations send from the following:
   - [x] Django admin
   - [x]  Domain Managers page (on a domain page)
   
  
- [x] On django admin, if a new domain invitation is created then send the same emails mentioned above (domain invitation and portfolio invitation if applicable) (the alert warning will be added to this table in #2370 )

Error handling for non-unique user email:

- [ ] ~~When making a domain invitation for an existing user, check for non-unique user email and trigger error msg/log on DJA and on the form View. Send email, do not proceed with creating the user domain role.~~ @dave-kennedy-ecs @abroddrick @rachidatecs discussed handling this edge case on the oidc side

Portfolio invites on existing user:

- [x] If someone is invited to a portfolio and they are already a registrar user, we should automatically create their user portfolio permission record (auto-retrieve the portfolio invite). This applies to straight portfolio invites, also applies to the 'side-effect' portfolio invites that get triggered on domain invites.

### Additional Context
this will be added to the members page in #3295 , you can do these two in the same PR if desirable, it was just seperated in case someone wants to do them in separate PRs (and since it is a different area in the code/unit tests)

### Issue Links

- ⛔  Blocked by #3019 
-  Blocks #3295 (or do together)
- bundled in #3296
[...]

PR #3304, opened 2025-01-06
#3228: DJA - new domain invitations are auto retrieved when user exists [MEOWARD]

### LINKED ISSUE #3228: Portfolio member domains list missing domain invites for existing members
### Current Behavior

The list of domains on portfolio domains management  pages gets different lists for members and an invites.

If it's an invite (ie email), we retrieve the domains from the domain invitations that intersect the email and portfolio. 

if it's a member (ie User), we get the domains from the user domain roles that intersect the user and the portfolio. In this case, we are missing any potential domains that user is invited to.

### Expected Behavior

The domain list for a member should include any domains that member is invited to.

### Steps to Reproduce

1. Go to Django admin domain invitation table
2. Add a domain invite to an existing portfolio User. 
3. Check the domains list for that user.
4. It does not include the domain added through the invite.

### Environment

_No response_

### Additional Context

Suggested solution: 
When going to the Domain assignment tab, if the user is an existing member, then we retrieve any “invited” domain invitations and generate the user domain roles. This is done prior to getting all user domain roles for the page, thus when we query for user domain roles we should then have all domains this person can manage.  

### Issue Links

_No response_

dave-kennedy-ecs:
## Ticket

Resolves #3228 

## Changes

- Modified django admin, DomainInvitation save so that new DomainInvitations created through DJA are auto-retrieved if a user already exists
- This solution "solves" the problem in ticket #3228 in that all of the views for existing portfolio users will not need to look at both UserDomainRoles and DomainInvitations when displaying information about domains a portfolio user has access to. Instead, the views will only need to look at UserDomainRoles 

- [x] Tag @dotgov-designers in this PR's Reviewers for design review. If code is not user-facing, delete design reviewer checklist

- [x] Verified code meets all checks above. Address any checks that are not satisfied
- [x] Verify migrations are valid and do not conflict with existing migrations

- [x] Meets all designs and user flows provided by design/product
- [x] (Rarely needed) Tested as both an analyst and applicant user

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References

### DIFF
--- a/src/registrar/admin.py
+++ b/src/registrar/admin.py
@@ -1439,18 +1439,12 @@ class DomainInvitationAdmin(ListHeaderAdmin):
         Override the save_model method.
 
         On creation of a new domain invitation, attempt to retrieve the invitation,
-        which will be successful if a user exists for that email; otherwise, will
-        raise a RuntimeError, and in this case can continue to create the invitation.
+        which will be successful if a single User exists for that email; otherwise, will
+        just continue to create the invitation.
         """
-        # NOTE: is a future ticket accounting for a 'not member of this org' scenario
-        # to mirror the logic in DomainAddUser view?
-        if not change:  # Domain Invitation creation
-            try:
-                User.objects.get(email=obj.email)
-                obj.retrieve()
-            except User.DoesNotExist:
-                # Proceed with invitation as new as exception indicates user does not exist
-                pass
+        if not change and User.objects.filter(email=obj.email).count() == 1:
+            # Domain Invitation creation for an existing User
+            obj.retrieve()
         # Call the parent save method to save the object
         super().save_model(request, obj, form, change)
 
--- a/src/registrar/admin.py
+++ b/src/registrar/admin.py
@@ -1434,6 +1434,26 @@ class DomainInvitationAdmin(ListHeaderAdmin):
         # Get the filtered values
         return super().changelist_view(request, extra_context=extra_context)
 
+    def save_model(self, request, obj, form, change):
+        """
+        Override the save_model method.
+
+        On creation of a new domain invitation, attempt to retrieve the invitation,
+        which will be successful if a user exists for that email; otherwise, will
+        raise a RuntimeError, and in this case can continue to create the invitation.
+        """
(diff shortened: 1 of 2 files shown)


src/registrar/admin.py
```diff
@@ -1434,6 +1434,26 @@ def changelist_view(self, request, extra_context=None):
...
+    def save_model(self, request, obj, form, change):
+        """
+        Override the save_model method.
+
+        On creation of a new domain invitation, attempt to retrieve the invitation,
+        which will be successful if a user exists for that email; otherwise, will
+        raise a RuntimeError, and in this case can continue to create the invitation.
+        """
+        # NOTE: is a future ticket accounting for a 'not member of this org' scenario
+        # to mirror the logic in DomainAddUser view?
```
abroddrick, line 1446: It should be! The follow on ticket, #2960 has this AC "If the invitee is not part of the portfolio yet, send the portfolio invitation email AND the domain invitation email."

src/registrar/admin.py
```diff
@@ -1434,6 +1434,26 @@ def changelist_view(self, request, extra_context=None):
...
+        if not change:  # Domain Invitation creation
```
abroddrick, line 1447: Why not just checkin the if statement,  if the email exists in the User table? That way we avoid just passing on an exception if we don't have to. Also, there is still the occasional issue where someone has multiple accounts with the same email. This scenario will cause a 500 error for the analyst that clicked the invitation, it also won't save the invitation. I would just add a check in here that it is a single email. 

Ugly suggestion just for the visual of what I am saying. This should then allow you to remove the try and except and catch the corner case of there being duplicate accounts.
```suggestion
        if not change and User.objects.filter(email=obj.email).count()==1:  # Domain Invitation creation
```

Yes the user will still get a 500 when they log in, but we do want them to contact us directly if this has happened.

dave-kennedy-ecs: Good suggestion! Implemented this suggestion and modified the comments accordingly. Much simpler looking code now!

abroddrick (approved):
LGTM

PR #3340, opened 2025-01-14
#3275: Slowness on Domain, Domain Info and Domain Requests - [NL]

### LINKED ISSUE #3275: Slowness on Domain, Domain Info and Domain Requests
### Current Behavior

Domain Requests, Domain and Domain info  tables on django admin are slow to navigate to and to apply filters on with large amounts data. In general they are slower than they were before even on smaller data sets (though small data sets make the difference harder to see in some cases). 

### Expected Behavior

It shouldn't take more than 2 seconds to load a page, really less than a second is what a user is expecting for going to a page or applying a filter. This ticket should bring down the load time as much as possible, ideally to < 1 second, but 2 seconds would also be acceptable in the meantime.

### Steps to Reproduce

1.go to django admin
2. got to Domains
3. see the long load time

*Note occasionally it seems to load at a reasonable speed, in which case clicking to the next page in the table usually results in a long load time again

### Environment

_No response_

### Additional Context

I loaded a decent amount of fixtures data to a sandbox and deploying main at the time of this ticket (basically staging-135) had a long load time. However, deploying staging-124 to the same sandbox resulted in less than a second load time for navigating to the tables and applying filters/pagination etc. 

### Issue Links

_No response_

CocoByte:
## Ticket 3275

Resolves #3275

## Changes

- Revised GenericOrgFilter and FederalTypeFilter to be more efficient
- Fixtures for DomainRequests was updated to allow us to specify number of entries to load.

## Context for reviewers

The admin tables for Domain Requests, Domains, and Domain Information were given a thorough check-up to tease out any remaining opportunities for optimization and to diagnose causes of slow load times.  

#### Findings for root cause of slowness
We already have taken measures to optimize performance of the admin tables, so the cause of slowness this time around was determined to be two filters that were recently added and not optimized for performance -- GenericOrgfilter and FederalTypeFilter.  These were added during the course of updating our model structure to use converted fields.  However, they were making expensive database calls.  They have been revised to be avoid making such calls.

#### Areas that were verified to be optimized
It is worth noting that the following structures/subroutines were examined and determined to NOT cause slowness.  (Code was commented out down to a bare-bones load, then different areas where systematically added back in to see performance impact.  Performance was tested for 100000 records.  We might want to re-examine some of these areas if we see performance issues with larger data sets)
- Pagination (This came up alot in the Django admin community as being a cause for performance slowdown, but I did not see any significant slowdown yet for 100000 records)
- Django's notoriously expensive .count() method (This one came as a surprise for me -- impact on performance was negligible if I disabled the count.  We might still want to disable it in the future for very large data sets but I left it alone for now.)
- Filters other than the two listed above (I inspected each filter for performance impact.  Only the two listed above created significant slowdown)
- Sorting
    - It appears that sorting on non-indexed data isn't having significant negative impact on performance so far.  We are doing OK on making sure we add indexes to our models.
    - The custom multisort structure had negligible impact on performance
- Pre-fetching is in place for foreign key data, so I did not see any obvious places to optimize there.

## Setup

- This might be easiest to test locally, since you need to load the database with large datasets and you need compare performance across both old and new code
- Load a LOT of data (I used 100000, but smaller values might work, too) into your local database for DomainRequests, Domains, and DomainInformation (for DomainRequests, you can do this by modifying the value of "total_domain_requests_to_make" in the fixture and refreshing your docker OR use whatever other method works for you)

Do the following for DomainRequest, Domain, and DomainInformation tables
- Compare load times of old code with load times of new code on large data sets. (Play around with filters, search, and sorting)
- [ ] Verify that load times on the new code is significantly faster.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References


src/registrar/tests/test_admin_domain.py
```diff
@@ -779,7 +779,8 @@ def test_short_org_name_in_domains_list(self):
         response = self.client.get("/admin/registrar/domain/")
         # There are 4 template references to Federal (4) plus four references in the table
         # for our actual domain_request
-        self.assertContains(response, "Federal", count=57)
+        self.assertContains(response, "Federal", count=54)
+        self.assertContains(response, "federal", count=225)
```
CocoByte, line 783: The updates to the filters means we no longer grab display values for Federal Type or Generic org type.  It's a small sacrifice that needs to be reflected in the unit tests, which check for camel case.

src/registrar/admin.py
```diff
@@ -1670,28 +1670,31 @@ class DomainInformationAdmin(ListHeaderAdmin, ImportExportModelAdmin):
     class GenericOrgFilter(admin.SimpleListFilter):
         """Custom Generic Organization filter that accomodates portfolio feature.
         If we have a portfolio, use the portfolio's organization.  If not, use the
-        organization in the Domain Information object."""
+        organization in the Domain Request object."""
```
asaki222, line 1674: Why was this updated to Domain Request object in this comment? It looks like this is under Domain Information

src/registrar/admin.py
```diff
@@ -3218,65 +3232,86 @@ def queryset(self, request, queryset):
...
 
         title = "generic organization"
```
asaki222, line 3237: It doesn't seem that we are grabbing info from the Domain Request object here.

asaki222 (dismissed):
I think this looks good overall. Just had a couple questions. 

On the sandbox, I got a 500 error when I hit federal type on the domain admin table, but generic org type works. 

I don't see the federal type filter on the right on the Domain Request table on the sandbox.

PR #3362, opened 2025-01-17
#3307: duplicate key bug for public contact [ZA]

### LINKED ISSUE #3307: Duplicate key issue in PublicContact

zandercymatics:
## Ticket

Resolves #3307 

## Changes

- Adds a try/catch and transaction.atomic() block in the get_or_create_public_contact function
- Adds a unit test

## Context for reviewers
As far as I can tell, this appears to be a race condition ([notes](https://docs.google.com/document/d/1PYHta6nsnx9ag00_V1rs3TxqYLpm84WeATV7PgCypek/edit?tab=t.0)). In both instances of this bug occurring, `get_contact_in_keys` was called twice with a very small time interval between. Right after this occurs - we get this fail condition. In addition, it seems like our app was undergoing heavy load at this same exact time (use the view surrounding documents button). 

The check in `_get_or_create_public_contact` is very concrete in that we just return the record if one exists but does not modify it. This is fine sequentially, but it's vulnerable to a common "check-then-act" type of race condition. Consider this sequence:
1. Thread A checks if contact exists (returns empty queryset)
2. Thread B checks if contact exists (returns empty queryset)
3. Thread A creates contact
4. Thread B attempts to create contact => IntegrityError

In other words, it appears as though the only code path that can cause an integrity error is if the _filter_ return an empty queryset at just the right time. I've included a unit test for this, and it reliably reproduces the bug without my fix and the test passes with it.

I've directly tested (modified code for) two alternative ideas:
1. The `get_id` on public_contact.py function returns the same id.
    **a.** Sequentially, the probability of collision is really low here -- roughly 10^25. random.choices() isn't uniformly distributed though like uuid, but the probability space is so high that I find this option less convincing.
    **b.**  In parallel (threads), `random` is thread-safe. [Pythons docs](https://docs.python.org/3/library/random.html) note concurrency issues when using a free-threaded build (so without the GIL), but we do not use that build.
2. The interaction between `_get_or_create_public_contact` and `_add_missing_contacts_if_unknown`  creates duplicates because .save() is called twice on the same record.
    **a.** It is worth noting that `_add_missing_contacts_if_unknown`  calls save, but doesn't check for duplicates like get_or_create does.
    **b.** This doesn't seem consistent with our log stack and is in line with what addAllDefaults does.
    **c.** Sequentially, I couldn't find a code path for an id conflict to occur because this error happens within _get_or_create_public_contact (which is called before add_missing_contacts) AND we have a check for its existence before attempting a save.

Ultimately though, we are missing just enough logging to prevent a more exhaustive search here. I've tried to add it to places that I think are important. 

## Setup
This is fairly tricky to test as this bug is somewhat rare in the first place. I've included a test case that is able to reliably reproduce the issue though  - and that might be a good place to test this. 

To cause this error, do the following in `_get_or_create_public_contact`:
Comment out this code:
```
            try:
                with transaction.atomic():
                    public_contact.save(skip_epp_save=True)
                    logger.info(f"Created a new PublicContact: {public_contact}")
            except IntegrityError as err:
                logger.error(
                    f"_get_or_create_public_contact() => tried to create a duplicate public contact: {err}",
                    exc_info=True,
                )
                return PublicContact.objects.get(
                    registry_id=public_contact.registry_id,
                    contact_type=public_contact.contact_type,
                    domain=self,
                )
 ```
 And replace it with this (or the old code):
 ```
 public_contact.save(skip_epp_save=True)
 ```
 
 Then run the included unit test. You should reliably see an integrity error as per the original issue description.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References

### DIFF
--- a/src/registrar/tests/common.py
+++ b/src/registrar/tests/common.py
@@ -578,6 +578,13 @@ class MockDb(TestCase):
             creator=cls.custom_superuser, federal_agency=cls.federal_agency_3, organization_type="federal"
         )
 
+        cls.suborganization_1, _ = Suborganization.objects.get_or_create(
+            name="SubOrg 1",
+            portfolio=cls.portfolio_1,
+            city="Nashville",
+            state_territory="TN",
+        )
+
         current_date = get_time_aware_date(datetime(2024, 4, 2))
         # Create start and end dates using timedelta
 
@@ -848,6 +855,7 @@ class MockDb(TestCase):
                 status=DomainRequest.DomainRequestStatus.IN_REVIEW,
                 name="city2.gov",
                 portfolio=cls.portfolio_1,
+                sub_organization=cls.suborganization_1,
             )
             cls.domain_request_3 = completed_domain_request(
                 status=DomainRequest.DomainRequestStatus.STARTED,
@@ -863,6 +871,9 @@ class MockDb(TestCase):
             cls.domain_request_5 = completed_domain_request(
                 status=DomainRequest.DomainRequestStatus.APPROVED,
                 name="city5.gov",
+                requested_suborganization="requested_suborg",
+                suborganization_city="SanFran",
+                suborganization_state_territory="CA",
             )
             cls.domain_request_6 = completed_domain_request(
                 status=DomainRequest.DomainRequestStatus.STARTED,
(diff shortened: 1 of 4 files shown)


CocoByte (approved):
Nice fix and good unit test.  This was a tricky one, and a beast to test.  I'm mainly going off of sound code logic and your thorough explanation (as well as that solid unit test)...overall, aside from a minor comment it looks good to me. .atomic() *should* take care of this.

PR #3430, opened 2025-01-30
#3145: Change usage of transaction.atomic() to be inside try/catch instead of outside of it [litterbox]

### LINKED ISSUE #3145: When using transaction.atomic(), exceptions should be caught outside of block
### Issue description
In our code, especially in fixtures, we're catching the exception inside the `atomic` block like this:
https://github.com/cisagov/manage.get.gov/blob/8a58070ac11067509ac6b19fcdbfb840c6f2c04f/src/registrar/fixtures/fixtures_portfolios.py#L90-L95

We should be catching the exception outside of the `atomic` block like this: 
  ```
  try:
        with transaction.atomic():
            generate_relationships()
    except IntegrityError:
        handle_exception()
```
(See here for details: https://docs.djangoproject.com/en/5.1/topics/db/transactions/#:~:text=Avoid%20catching%20exceptions%20inside%20atomic!)

### Acceptance criteria

- [ ] All instances of `transaction.atomic()` should catch exceptions outside of the `atomic()` block
- [ ] Do a sanity check that all the fixtures functions still work as intended

### Additional context

_No response_

### Links to other issues

 Relates to https://github.com/cisagov/manage.get.gov/pull/3089

### PR DESCRIPTION
**Pull request by zandercymatics**
## Ticket

Resolves #3145 

## Changes

- Moves transaction.atomic() inside of try/catch blocks in multiple places

## Context for reviewers
This mostly impacts fixtures, so the most efficient way to test this is locally. However, it does also touch the transfer_user logic so that is worth testing.

## Setup
Run this locally or on a sandbox, and test our fixtures. Additionally, try transferring a user. Both should work.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References

### DIFF
--- a/src/registrar/fixtures/fixtures_portfolios.py
+++ b/src/registrar/fixtures/fixtures_portfolios.py
@@ -111,7 +111,7 @@ class PortfolioFixture:
         except Exception as e:
             logger.warning(e)
 
-        # Bulk create domain requests
+        # Bulk create portfolios
         if len(portfolios_to_create) > 0:
             try:
                 Portfolio.objects.bulk_create(portfolios_to_create)

### CODE REVIEW

File: src/registrar/utility/db_helpers.py [hunk 1]
```diff
@@ -9,12 +9,13 @@ def ignore_unique_violation():
     Execute within an atomic transaction so that if a unique constraint violation occurs,
     the individual transaction is rolled back without invalidating any larger transaction.
     """
-    with transaction.atomic():
```
rachidatecs, line 12: Look at the method description. I think this was implemented with a try inside an atomic transaction on purpose, so that a fail would get skipped instead of block. A hack, in other words. Need confirmation from @Matt-Spence on this.

rachidatecs (changes requested):
The refactor blocks fixture re-runs from creating extra records - this is expected in hindsight, as a single conflict within a batch will now invalidate the whole batch instead of getting skipped. I think the more correct refactor would be to remove atomic transactions from fixtures completely.

src/registrar/utility/db_helpers.py
Matt-Spence: That is correct. In practice this is a nested atomic transaction, which basically functions as a "save point" that can be rolled back to without invalidating the larger transaction. If any exception is thrown outside the inner transaction, even if it is caught, the outer transaction will be invalidated. It's a bit hacky but it was far and away the simplest method I could find to handle uniqueness violations.

rachidatecs (approved):

PR #3513, opened 2025-02-12
#3440: Opt-out permissions [DK]

### LINKED ISSUE #3440: Audit permissions and refactor
### Issue description

Our current approach to permissions is to add mixins that restrict who can access a given view. This results in views being insecure by default and requires us to remember each time to add permissions to sensitive views. We should invert this structure so that views are secure by default, and mixins or other code structures are used to loosen restrictions on a case-by-case basis. This will change our failure case from insufficient restrictions to excessive restrictions, which is preferable. 

### Acceptance criteria
- [x] Audit all existing urls to check for vulnerabilities (communicate with leads if any are found)
- [x] Review https://github.com/cisagov/manage.get.gov/issues/659 in regards to refactoring our permissions
- [x] All views have high level permissions imposed by default
  - [x] This should mean either nobody can access them, or only our most trusted users can access them.
  - [x] Ensure that newly created views will have these permissions imposed without requiring any action or knowledge by the developer.
- [x] Mixins or other tools are available to allow for less restrictive access to specific views
- [x] All existing views are refactored to reflect these changes
- [x] Be sure to double check unconventional views such as api end points for permission handling

### Links to other issues

_No response_

dave-kennedy-ecs:
## Ticket

Resolves #3440 

## Changes

- Added middleware which restricts access by default, unless explicitly granted by a view or class decorator
- Added grant_access decorator and associated permissions previously defined by a combination of mixins, permission views, and views.
- Enumerated a list of permissions that can be assigned to views through grant_access to grant access to views.
- Removed mixins and permission_view
- Fixed a few views which did not have assigned permissions
- Result is an OPT-OUT approach to access restrictions. If permissions not associated with a view, the view will be restricted by default.

## Context for reviewers

The heart of the change is in two files:
request_middleware.py
decorators.py

All other changes stem from these two. A few notes on other changes.
In urls, url parameters were standardized in many places so that the decorators could predictably evaluate urls for permission. This standardization resulted in many subsequent changes in templates, views, and tests.
Permissions_views and mixins were removed, as they were replaced by the above two files.
For views which had additional permissions, based not on user but rather on the resource or some other logic, that logic was defined in has_permission within the view.

- [x] Tag @dotgov-designers in this PR's Reviewers for design review. If code is not user-facing, delete design reviewer checklist

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References


src/registrar/config/urls.py
```diff
@@ -392,6 +396,7 @@
 # This way, we can share a view for djangooidc, and other pages as we see fit.
 handler500 = "registrar.views.utility.error_views.custom_500_error_view"
 handler403 = "registrar.views.utility.error_views.custom_403_error_view"
+handler404 = "registrar.views.utility.error_views.custom_404_error_view"
```
dave-kennedy-ecs, line 399: Added this, but not certain if this is necessary, or if always_404 is working properly. Added because was seeing instances of generic 404 json responses. Now, not so sure we need the redundancy

src/registrar/config/urls.py
```diff
@@ -68,7 +68,7 @@
     (PortfolioDomainRequestStep.REQUESTING_ENTITY, views.RequestingEntity),
     (PortfolioDomainRequestStep.ADDITIONAL_DETAILS, views.PortfolioAdditionalDetails),
 ]:
-    domain_request_urls.append(path(f"<int:id>/{step}/", view.as_view(), name=step))
+    domain_request_urls.append(path(f"<int:domain_request_pk>/{step}/", view.as_view(), name=step))
```
dave-kennedy-ecs, line 71: Standardized from generic id or pk in many cases to domain_pk, domain_request_pk and domain_invitation_pk so that grant_access decorator can efficiently access domain, domain_request, and domain_invitation from the url kwargs to make authorization decisions.

src/registrar/views/utility/error_views.py
```diff
@@ -35,3 +35,10 @@ def custom_403_error_view(request, exception=None, context=None):
     if context is None:
         context = {}
     return render(request, "403.html", context=context, status=403)
+
+
+def custom_404_error_view(request, exception=None, context=None):
+    """Used to redirect 404 errors to a custom view"""
+    if context is None:
+        context = {}
+    return render(request, "404.html", context=context, status=404)
```
dave-kennedy-ecs, line 44: As noted earlier, added this view when I noticed 404 generic json responses. Then saw always_404, so there appears to be redundancy, but need further testing to see if one can be removed. This handling is in same place as 403 and 500 so may be the way to go.

src/registrar/decorators.py
```diff
@@ -0,0 +1,300 @@
...
+IS_PORTFOLIO_MEMBER_AND_DOMAIN_MANAGER = "is_portfolio_member_and_domain_manager"
+IS_DOMAIN_MANAGER_AND_NOT_PORTFOLIO_MEMBER = "is_domain_manager_and_not_portfolio_member"
+HAS_PORTFOLIO_DOMAINS_ANY_PERM = "has_portfolio_domains_any_perm"
+HAS_PORTFOLIO_DOMAINS_VIEW_ALL = "has_portfolio_domains_view_all"
+HAS_PORTFOLIO_DOMAIN_REQUESTS_ANY_PERM = "has_portfolio_domain_requests_any_perm"
+HAS_PORTFOLIO_DOMAIN_REQUESTS_VIEW_ALL = "has_portfolio_domain_requests_view_all"
+HAS_PORTFOLIO_DOMAIN_REQUESTS_EDIT = "has_portfolio_domain_requests_edit"
+HAS_PORTFOLIO_MEMBERS_ANY_PERM = "has_portfolio_members_any_perm"
+HAS_PORTFOLIO_MEMBERS_EDIT = "has_portfolio_members_edit"
+HAS_PORTFOLIO_MEMBERS_VIEW = "has_portfolio_members_view"
```
Matt-Spence, line 22: Realizing now that this is basically a half-way implementation of Role Based Access Control, and it might benefit us in the long run to go a step further and adopt an Attribute Based Access Control system. We should huddle in slack to discuss what this might look like further, but just to illustrate a few of the problems we might run into with this setup (which to be clear is an excellent implementation of the original proposal):
1. if we ever want to change permissions for a role application-wide, we will have to edit the permissions for every view that invokes that rule (eg if we change what managers have access to, we need to check every instance of is_manager instead of changing it in one configuration file)
2. Rules constants are a mix of references to roles, permissions, and different models. This is super flexible, which is great, but the list of rules is already long and will likely only grow. Over time we risk having duplicate permissions or, worse, wrongly applied permissions because they get too complicated.
3. Granting access based on model or user attributes is complicated and doesn't necessarily live in a central location.

Matt-Spence (approved):
I found some things while testing that we might want to address, but they're not part of this PR. Amazing work.

PR #3613, opened 2025-03-06
#3548: Fix portfolio permissions on resources - [ZA]

### LINKED ISSUE #3548: Users with Portfolio View All Domain Requests permissions can view Domain Requests outside their portfolio
### Issue description

Currently, any user with permission to view all domain requests for a portfolio can access a view-only view of any domain request by going into the browser and hitting the /domain-request/viewonly/{domain id} endpoint manually. This appears to be because the PortfolioDomainRequestsViewOnly view only checks if that view all permission exists, and not whether it exists *for the domain requests portfolio*. 

This same issue also exists for members and domains.

### Acceptance criteria

- [ ] Portfolio users can only view/manage/delete domain requests that are in their portfolio
- [ ] Portfolio users can only view/manage/delete members that are in their portfolio
- [ ] Portfolio users can only view/manage domains that are in their portfolio

### Additional context

Update from Zander: Additional testing has found that this issue also occurs on the member and domains pages. [See this thread for info](https://cisa-corp.slack.com/archives/C05BGB4L5NF/p1741211146279669). 
_No response_

### Links to other issues

_No response_

zandercymatics:
## Ticket

Resolves #3548 

## Changes

- Adds permission checks on portfolio resources like Domain, DomainRequest, and Member (UserPortfolioPermission/PortfolioInvitation)
- Fixes permissions on the MemberDelete view
- Unit tests

## Context for reviewers
This PR was originally tasked with correct an issue with domain request where you could access any other portfolio domain request (when the waffle flag was on) by simply changing the id in the url. As part of this effort, I also identified the same issue with domain and member.

## Setup
You will need to test domain, domain request, and member. For domain/request, you will need to test in org mode and non-org mode.

I've found that the easier approach to this is by creating a new portfolio, and adding some domains, domain requests, and members to them. Then add yourself to that portfolio. On getgov-za, use `AMTRAK` for this exact purpose.

Afterwards, navigate to the edit, manage, and view contexts of each record type (if applicable). In the url bar, change the id to a valid id of another record. You should get a 403 error.

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References

### DIFF
--- a/src/registrar/decorators.py
+++ b/src/registrar/decorators.py
@@ -235,11 +235,11 @@ def _is_domain_manager(user, **kwargs):
 def _domain_exists_under_portfolio(portfolio, domain_pk):
     """Checks to see if the given domain exists under the provided portfolio.
     HELPFUL REMINDER: Watch for typos! Verify that the kwarg key exists before using this function.
-    Returns True if the pk is None. Otherwise, returns a bool if said object exists.
+    Returns True if the pk is falsy. Otherwise, returns a bool if said object exists.
     """
     # The view expects this, and the page will throw an error without this if it needs it.
     # Thus, if it is none, we are not checking on a specific record and therefore there is nothing to check.
-    if domain_pk is None:
+    if not domain_pk:
         logger.warning(
             "_domain_exists_under_portfolio => Could not find domain_pk. "
             "This is a non-issue if called from the right context."
@@ -251,11 +251,11 @@ def _domain_exists_under_portfolio(portfolio, domain_pk):
 def _domain_request_exists_under_portfolio(portfolio, domain_request_pk):
     """Checks to see if the given domain request exists under the provided portfolio.
     HELPFUL REMINDER: Watch for typos! Verify that the kwarg key exists before using this function.
-    Returns True if the pk is None. Otherwise, returns a bool if said object exists.
+    Returns True if the pk is falsy. Otherwise, returns a bool if said object exists.
     """
     # The view expects this, and the page will throw an error without this if it needs it.
     # Thus, if it is none, we are not checking on a specific record and therefore there is nothing to check.
-    if domain_request_pk is None:
+    if not domain_request_pk:
         logger.warning(
             "_domain_request_exists_under_portfolio => Could not find domain_request_pk. "
             "This is a non-issue if called from the right context."
@@ -267,11 +267,11 @@ def _domain_request_exists_under_portfolio(portfolio, domain_request_pk):
 def _member_exists_under_portfolio(portfolio, member_pk):
     """Checks to see if the given UserPortfolioPermission exists under the provided portfolio.
     HELPFUL REMINDER: Watch for typos! Verify that the kwarg key exists before using this function.
-    Returns True if the pk is None. Otherwise, returns a bool if said object exists.
+    Returns True if the pk is falsy. Otherwise, returns a bool if said object exists.
     """
     # The view expects this, and the page will throw an error without this if it needs it.
     # Thus, if it is none, we are not checking on a specific record and therefore there is nothing to check.
-    if member_pk is None:
+    if not member_pk:
         logger.warning(
             "_member_exists_under_portfolio => Could not find member_pk. "
             "This is a non-issue if called from the right context."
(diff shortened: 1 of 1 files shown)


src/registrar/decorators.py
```diff
@@ -98,24 +104,38 @@ def _user_has_permission(user, request, rules, **kwargs):
     if not user.is_authenticated or user.is_restricted():
         return False
 
+    portfolio = request.session.get("portfolio")
     # Define permission checks
     permission_checks = [
         (IS_STAFF, lambda: user.is_staff),
-        (IS_DOMAIN_MANAGER, lambda: _is_domain_manager(user, **kwargs)),
```
Matt-Spence, line 104: I'm looking at this and wondering about testing. We have talked about needing to expand our unit test coverage for these permission checks. It might be easier to do that if, rather than putting the permission logic in-line, we write a matching function for each rule. Like how this one works:
```
(IS_STAFF_MANAGING_DOMAIN, lambda: _is_staff_managing_domain(request, **kwargs))
```
This way, we can more easily write unit tests for every rule. This doesn't have to be done as part of this ticket per se, I just want to throw the idea out there since it's been a subject of discussion recently.

src/registrar/decorators.py
```diff
@@ -129,34 +149,55 @@ def _user_has_permission(user, request, rules, **kwargs):
...
+            and _domain_request_exists_under_portfolio(portfolio, kwargs.get("domain_request_pk")),
         ),
         (
             HAS_PORTFOLIO_MEMBERS_ANY_PERM,
             lambda: user.is_org_user(request)
             and (
-                user.has_view_members_portfolio_permission(request.session.get("portfolio"))
-                or user.has_edit_members_portfolio_permission(request.session.get("portfolio"))
+                user.has_view_members_portfolio_permission(portfolio)
+                or user.has_edit_members_portfolio_permission(portfolio)
```
Matt-Spence, line 171: This is more of a theoretical question, but do we always assume that edit permissions imply view permissions? In theory there are situations where you might want a user to be able to write something without being able to view it. Not sure any such situations apply in our case, but something to think about.

src/registrar/decorators.py
```diff
@@ -129,34 +149,55 @@ def _user_has_permission(user, request, rules, **kwargs):
...
+            )
+            and (
+                # AND rather than OR because these functions return true if the PK is not found.
+                # This adds support for if the view simply doesn't have said PK.
+                _member_exists_under_portfolio(portfolio, kwargs.get("member_pk"))
+                and _member_invitation_exists_under_portfolio(portfolio, kwargs.get("invitedmember_pk"))
```
Matt-Spence, line 177: One possible edge case we might run into with this and: what if an analyst deletes an invitation but leaves a portfolio permission in place? Are we ok with that causing a user to fail this check? Same below.

src/registrar/decorators.py
```diff
@@ -191,6 +232,70 @@ def _is_domain_manager(user, **kwargs):
...
+    Returns True if the pk is None. Otherwise, returns a bool if said object exists.
+    """
+    # The view expects this, and the page will throw an error without this if it needs it.
+    # Thus, if it is none, we are not checking on a specific record and therefore there is nothing to check.
+    if not domain_pk:
+        logger.info(
+            "_domain_exists_under_portfolio => Could not find domain_pk. "
+            "This is a non-issue if called from the right context."
+        )
+        return True
```
Matt-Spence, line 247: I still don't fully grasp why we're returning true in these functions if the pk is falsy.

src/registrar/decorators.py
```diff
@@ -191,6 +232,70 @@ def _is_domain_manager(user, **kwargs):
...
+    return Domain.objects.filter(domain_info__portfolio=portfolio, id=domain_pk).exists()
+
+
+def _domain_request_exists_under_portfolio(portfolio, domain_request_pk):
+    """Checks to see if the given domain request exists under the provided portfolio.
+    HELPFUL REMINDER: Watch for typos! Verify that the kwarg key exists before using this function.
+    Returns True if the pk is None. Otherwise, returns a bool if said object exists.
```
Matt-Spence, line 254: technically this returns true any time the pk is falsy, not just None. Unsure if that matters just yet, but important to keep in mind.

src/registrar/decorators.py
```diff
@@ -191,6 +232,70 @@ def _is_domain_manager(user, **kwargs):
...
 
+def _domain_exists_under_portfolio(portfolio, domain_pk):
+    """Checks to see if the given domain exists under the provided portfolio.
+    HELPFUL REMINDER: Watch for typos! Verify that the kwarg key exists before using this function.
+    Returns True if the pk is None. Otherwise, returns a bool if said object exists.
+    """
+    # The view expects this, and the page will throw an error without this if it needs it.
+    # Thus, if it is none, we are not checking on a specific record and therefore there is nothing to check.
+    if not domain_pk:
+        logger.info(
```
Matt-Spence, line 243: IMO these should be warn logs, since there are situations where this would represent an incorrect use of the function.

Matt-Spence (commented):
Couple of questions, but overall great work

src/registrar/decorators.py
Matt-Spence: Sorry I meant this to apply to the other functions as well. Anywhere you felt the need to comment `This is a non-issue if called from the right context.` probably deserves a warn log IMO.

PR #3792, opened 2025-05-16
Delete .github/workflows/issue-label-notifier.yaml

### LINKED ISSUE #2723: Notify users based on issue labels runs and fails when adding a label on a PR
### Current Behavior

If Alysia adds a label to a PR to indicate Review Priority or carryover the githhub action "Notify users based on issue labels" is triggered and then fails.

Workflow run on a PR: https://github.com/cisagov/manage.get.gov/actions/runs/10741106356
Error message for the failure: `TypeError: Cannot read properties of undefined (reading 'number')`

Workflow run (successful) on an Issue: https://github.com/cisagov/manage.get.gov/actions/runs/10741487874

### Expected Behavior

This action should only run when "design-review" is added to an issue or a pull request. It should not run for any other label and should not fail when added to a pull request. 

### Steps to Reproduce

1. Add a label to a PR
2. see the action being triggered
3. see the fail on the action
4. Creator of label gets an email saying things failed in big scary letters

### Environment

github

### Additional Context

workflow is the issue-label-notifier.yaml file

I suspect given the libraries name is issue-label-notification-action, that it may just support issues only, and not PRs. 

See comments below that suggest switching the libarary used  to [action-label-notification](https://github.com/iamfj/action-label-notification)
### Issue Links

_No response_

katypies:
## Ticket

- Resolves #2723 by removing the action altogether

## Changes

- Removes the issue-notifier action

## Context for reviewers

This only removes the file, so no code changes to review. It can't really be tested on Staging since it's a GitHub workflow.


erinysong (approved):
LGTM! Hopefully we can do this directly on GitHub and don't need to manage this as our own workflow anymore 🫡

PR #3808, opened 2025-05-20
Define platform in docker-compose file and update readme

kimallen:
## Ticket

~Resolves #00~ Adhoc update

## Changes

- Updates the `docker-compose.yaml` to define the platform for services `app`, `node`, `db` and `pa11y`
- Updates the readme: removes the troubleshooting step that the above update makes unnecessary

## Context for reviewers

While trying to get the application running, I ran into an issue because of my M1 Mac. I used the troubleshooting step to fix the problem. However, in the process of exploring the issue, I found another solution that would be a one-time fix that would eliminate the issue for future devs. I wasn't completely sure if it was necessary for pa11y or all the ones I added a platform for, but figured it wouldn't hurt.

## Setup

Ideally someone with an M1 _and_ someone who has an intel chip or windows machine each test this.

 1. If you have an M1 (or M2 etc), and have exported the env't var in your bash-profile or zshrc (or equivalent), comment out that line and source the file. 
 2. run `docker compose up --build` and make sure that it builds and the app runs properly
    


src/docker-compose.yml
```diff
@@ -1,5 +1,6 @@
 services:
   app:
+    platform: linux/amd64
```
therealslimhsiehdy, line 3: **Question:** My two cents is that I get that this line build/runs the image using this platform even if the main system has diff architecture, but: 
1. Would this fail for Docker on Windows and/or would this build slower (not sure how to test)?
2. Should we allow users to choose their platform versus forcing them to use this architecture?

Open to thoughts/ideas/suggestions bc I dont have strong expertise in this section, and maybe @DaisyGuti could lend a hand with their expertise?

kimallen: I don't have much expertise on this either, so definitely relying on others' expertise to stop me if there's any downstream issues I'm creating. I did ask similar questions in an engineering thread while researching this, and the responses were:

"This seems like it's about what platform to target, so I think it should be fine regardless of the user's OS."
and 

"Windows developers run Docker in a Linux VM, typically WSL 2. If you have platform: linux/amd64 in the docker-compose.yaml then it shouldn't break things for them since they're already running containers inside a Linux environment that supports amd64 by default."

Issue #3449, opened 2025-02-03
A portfolio invitation with a retrieved status will block you from re-inviting the same email


rachidatecs:
### Current Behavior

If you delete a previously invited and retrieved member from a portfolio, you can't re-invite that member. The retrieved invitation triggers "This user is already assigned to a portfolio invitation".

### Expected Behavior

A stale invitation with a status of retrieved should have no effect on the creation of new invitations.

### Steps to Reproduce

1. Invite an an existing user to a portfolio. This will auto-retrieve the portfolio invitation.
2. Delete the new portfolio member. The invitation with status retrieved should still exist, this is desired behavior.
3. Try to re-invite the same user. You should get an error This user is already assigned to a portfolio invitation"

### Additional Context

It's possible that all that's needed is to filter out retrieved invitations here:

![Image](https://github.com/user-attachments/assets/9dbdbdd7-659c-4e11-9bb4-44d33629d08f)

PR #4293, opened 2025-10-16
4035 - Update the query to look for non case sensitive match - [AD]

### LINKED ISSUE #4035: Invitations shouldn't be case-sensitive
### Current Behavior

Analysts have reported that some users have been invited to domains, but then can't see the domain upon login. They're able to add a new invitation, and then the user can see the domain as expected.

After investigation, it appears that there might be an issue with invitations that have capital letters in them, in a few different flows (see examples/replication steps). 

In the case originally reported, the first invitation (created by an analyst) had an email address like `Firstname.Lastname@domain.gov` which is still in the retrieved state. The new invitation that the analyst sent used all lowercase letters, and is "retrieved".

### Expected Behavior

Since email addresses and domain names aren't case-sensitive in general, invitations based on email addresses shouldn't be either.

### SCENARIO 1
 
1. Create a new domain invitation in /admin, using capital letters somewhere in the email address (or alternate capitalization from the email address, if a user already exists with that email address)
2. Invitation is sent
3. Log in to retrieve the invitation
4. Domain is missing
5. Add a new invitation without capital letters (or matching the email address for the user account)
6. Domain invitation is immediately "retrieved" since the user already exists in the system

### SCENARIO 2

On Staging, my account (email address with all lowercase letters) is part of a portfolio, and manages a few domains but not all of them in the portfolio.

**2A**

1. Go to domain invitations table
2. Create new domain invitation for a domain I _don't_ already manage
3. Add my email address, but with some capital letters in the address
4. Confirmation message indicates:
   * ~A new invitation to the portfolio was sent (shouldn't happen, because my email address is already part of the portfolio - **portfolio invitation that's generated from an /admin domain invitation is case-sensitive**)~
   * An invitation to manage the domain was sent (**expected**, because I don't yet manage the domain, in any version of my email address)
   * The invite was created with the status of "invited"
5. Invite not retrieved (it should be, since I'm already a member of the portfolio, and a registrar user)

**2B**

1. Go to domain invitations table
2. Create new domain invitation for a domain I _do_ already manage
3. Add my email address, but with different capital letters in the address
4. Domain invitation is sent

**2C**

1. Go to portfolio invitations table
2. Create a new portfolio invitation for my email address (all caps this time)
3. Appropriately get an error: "This user is already assigned to a portfolio"

### Environment

Stable, replicated on Staging

### Additional Context

See [Slack thread](https://cisa-corp.slack.com/archives/C06VDEWTGSH/p1753975533036959) for specific user examples on Stable.

The above replication steps are in /admin, but as of 9/17/25, we have at least one additional report of someone being invited to a portfolio by an org admin, and the invitation being case-sensitive.

### Issue Links

🔄 Looks related to #3657 (which maybe was only resolved in the portfolio invitations, or on the user front-end?)

asaki222:
## Ticket

Resolves #4035 

## Changes

- When a queried our db, there were no email that were mixed case. However, this isn't enforced in our code.
- The email in lowercase is now enforced on the model level
- Within the invitation admin, the email is now lowercased before querying the User table.

To verify that the code works as expected:

**Scenario 1**

- Create a new domain invitation in /admin, using capital letters somewhere in the email address (or alternate capitalization from the email address, if a user already exists with that email address)
- Invitation is sent
- Log in to retrieve the invitation
- Domain should be there with the option to manage it.
- Add a new invitation (same domain) without capital letters (or matching the email address for the user account), and  you should get an error that this user is already a domain manager. 

**Scenario 2**

**2A**

- Go to domain invitations table
- Create new domain invitation for a domain I don't already manage
- Add my email address, but with some capital letters in the address
- Confirmation message should indicate that the invitation was sent

**2B**

- Go to domain invitations table
- Create new domain invitation for a domain with a user that is already managing that domain.
- Add the email address, but with different capital letters in the address
- Domain invitation should not be sent

**Scenario 3**
```
cf ssh getgov-ad
/tmp/lifecycle/shell
./manage.py shell
from registrar.models import Portfolio, PortfolioInvitation
from registrar.models.utility.portfolio_helper import UserPortfolioRoleChoices

portfolio = Portfolio.objects.get(organization_name="Wish You Were Here")

PortfolioInvitation.objects.create(email="DifferentCase@examplE.com", portfolio=portfolio, roles=[UserPortfolioRoleChoices.ORGANIZATION_MEMBER])

portfolio.get_domains()
```
- from the domains listed, chose one 
- Go into the domain invitations table, and add the email that you just added via the terminal to that domain.
- It should not fire a Portfolio Invitation, just a domain invitation.

**Scenario 4**
- Turn on the multiple waffle flag feature
- Add PortfolioInvitations with the same email and portfolio with different casing
- It should show you different errors based if the user already has a role within the portfolio or there is an existing invitation. 

**RETRIEVED STATE NOTE*
It will allow add of an invitation if the invitation is in the retrieved state. This is the expected behavior according to this issue. [3449](https://github.com/cisagov/manage.get.gov/issues/3449)
- To test this working properly, delete the user portfolio permissions of a user that has an invitation status of "retrieved".
- Try adding the portfolio invitation with the same email, it should allow you to do so.

#### Also things to note

- if the domain managers aren't in the allowed email list, there will be an warning at the top that says that "Could not send email notification to existing domain managers." 
- To remediate this,  you can remove the email that is not in the allowed list since a bunch of them are dummy emails. And you can add your account or  test account as the domain manager in its place.
- PortfolioInvitations will not send if the user is already part of another portfolio

  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References

### DIFF
--- a/src/registrar/tests/test_admin.py
+++ b/src/registrar/tests/test_admin.py
@@ -1379,7 +1379,9 @@ class TestPortfolioInvitationAdmin(TestCase):
     @patch("django.contrib.messages.error")  # Mock the `messages.warning` call
     def test_save_does_not_send_email_if_requested_user_exists(self, mock_messages_error, mock_send_email):
         """On save_model, an email is NOT sent if an the requested email belongs to an existing user.
-        It also throws a warning."""
+        It also throws a warning.
+        Integrated testing for casing normalization to lowercase for email Portfolio Invitation
+        """
         self.client.force_login(self.superuser)
 
         # Create an instance of the admin class
@@ -1389,9 +1391,12 @@ class TestPortfolioInvitationAdmin(TestCase):
         existing_user = create_user()
         UserPortfolioPermission.objects.create(user=existing_user, portfolio=self.portfolio)
 
+        # An email with different capitalization that the existing user
+        existing_user_email_with_capitalization = "STaff@example.com"
+
         # Create a PortfolioInvitation instance
         portfolio_invitation = PortfolioInvitation(
-            email=existing_user.email,
+            email=existing_user_email_with_capitalization,
             portfolio=self.portfolio,
             roles=[UserPortfolioRoleChoices.ORGANIZATION_ADMIN],
         )


kimallen (changes requested):
Fixes it!
Could we also update the DomainInvitation `save` to save the email to match the casing of the User email? Open to discussion if you think otherwise.

src/registrar/admin.py
```diff
@@ -1861,6 +1861,10 @@ def save_model(self, request, obj, form, change):
             requested_user = get_requested_user(requested_email)
             requestor = request.user
 
+            # set object email to appropiate user email if it exists
+            if requested_user and requested_user.email:
+                obj.email = requested_user.email
+
             member_of_a_different_org, member_of_this_org = get_org_membership(
```
asaki222: @kimallen  I have added this to add the user email from the User model if it exists.

src/registrar/admin.py
```diff
@@ -1856,11 +1856,15 @@ def save_model(self, request, obj, form, change):
...
             domain_org = getattr(domain.domain_info, "portfolio", None)
-            requested_email = obj.email
+            requested_email = obj.email.lower()
             # Look up a user with that email
             requested_user = get_requested_user(requested_email)
             requestor = request.user
 
+            # set object email to appropiate user email if it exists
+            if requested_user and requested_user.email:
+                obj.email = requested_user.email
```
kimallen: I believe you want this to be `requested_email = requested_user.email` since below in the `DomainInvitation` create it uses `requested_email`

src/registrar/admin.py
```diff
@@ -1856,14 +1856,12 @@ def save_model(self, request, obj, form, change):
         if not change:
             domain = obj.domain
             domain_org = getattr(domain.domain_info, "portfolio", None)
-            requested_email = obj.email.lower()
+            obj.email = obj.email.lower()
```
kimallen: You could simplify these two lines to `request_email = obj.email.lower()`

kimallen (approved):
Thanks for being willing to refactor and come up with a different solution- I think this update is more robust!

Do we know that existing records of DomainInvitations are also all lowercase in stable? If not, I'd suggest including the change to `get_requested_user` to query with `email_iexact`. It would cover past situations, and if we use that method in other places, we won't get caught with a similar bug. I suggest that change, but not blocking on it.

One thing I noticed in re-testing is that when submitting a Domain Invitation, it `.lowers()` the value in the field, but not in the alert message. I'm not too worried about that user experience since it's just analysts, but in an ideal world they would be aligned.
Non-blocking.

image

asaki222:
> Thanks for being willing to refactor and come up with a different solution- I think this update is more robust!
> 
> Do we know that existing records of DomainInvitations are also all lowercase in stable? If not, I'd suggest including the change to `get_requested_user` to query with `email_iexact`. It would cover past situations, and if we use that method in other places, we won't get caught with a similar bug. I suggest that change, but not blocking on it.
> 
> One thing I noticed in re-testing is that when submitting a Domain Invitation, it `.lowers()` the value in the field, but not in the alert message. I'm not too worried about that user experience since it's just analysts, but in an ideal world they would be aligned. Non-blocking.
> 
> image

I did some digging the capitalization comes from usdws styling. They use sentence case for the warning messages.

kimallen:
When I add a second portfolio invitation with the same email but with different casing, It still creates another invitation. and I see these alerts:
image

image

CocoByte:
> When I add a second portfolio invitation with the same email but with different casing, It still creates another invitation. and I see these alerts: image
> 
> image

We will address this in follow-on tickets (see comment on #4317).

src/registrar/tests/test_admin.py
kimallen: Could you add some capitalization to the email used here to make sure that it still works? (and put a comment as to the fact that it's different because it's testing caps diff)

PR #4343, opened 2025-10-30
#4151: Set up mocks for external apis for local development

### LINKED ISSUE #4151: Set up a way to mock external apis to use in the local dev env't
### Sub-issue description

Currently, anytime we want to add a dns record, we hit the actual Cloudflare api. In a local environment, we ideally don't want to be hitting external apis like that and creating actual external records. Let's research and set up options  to intercept the api calls. If we end up using a FE js based framework, [Mock Service Worker](https://mswjs.io/) is an option. Another possibility is Django's responses library. 

This should be used in a local environment only. 

### Acceptance criteria

- [ ] We've designated a method for mocking external apis
- [ ] We have an example set up for mocking
- [ ] Mock all cf apis if it makes sense to be part of this ticket, otherwise create a followup

### Additional context

_No response_

kimallen:
## Ticket

Resolves #4151

## Changes

- Adds library respx
- Adds a MockCloudflareService to mock the external (cloudflare) apis
- Adds a bool environment variable `DNS_MOCK_EXTERNAL_APIS`
- Refactors how the account name is generated (and it's format)
- Adds documentation to developer README on how to set up the mock

## Setup

1) Please try out the instructions I added to the readme for setting up the mocks.

1) Add a dns record to a domain. Check the logs to see that the mock was used, and it should also be a successful submit.

2) Test the error triggers by adding the corresponding record names:
`error-403*` --> raises a 403 error
`error-400*` --> raises a 400 error
`error*` (but not the above) --> raises a 500 error

3) Create a domain `exists.gov`
Add a record. Check the logs to see that neither an account nor zone was created, but the dns record was.
Note: for this case, once saving of db accounts is implemented, this will need modification so that we do save an account and zone in the db the first time `exists.gov` has a record created. (maybe a check that it's in the db as part of a conditional that gates the response)

### DIFF
--- a/src/registrar/views/domain.py
+++ b/src/registrar/views/domain.py
@@ -772,7 +772,7 @@ class PrototypeDomainDNSRecordForm(forms.Form):
 class PrototypeDomainDNSRecordView(DomainFormBaseView):
     template_name = "prototype_domain_dns.html"
     form_class = PrototypeDomainDNSRecordForm
-    valid_domains = ["igorville.gov", "domainops.gov", "dns.gov", "age-bit-low-far.gov", "exists.gov"]
+    valid_domains = ["igorville.gov", "domainops.gov", "dns.gov", "exists.gov"]
 
     def __init__(self):
         self.dns_record = None


docs/developer/README.md
kimallen: Most of the changes here are linter-induced. Go to the bottom of the file to see the new instructions.

src/registrar/config/settings.py
```diff
@@ -809,12 +810,22 @@ def format(self, record):
...
+# region: DNS----------------------------------------------------------###
+
+# SECURITY WARNING: keep all DNS variables in production secret!
 SECRET_DNS_TENANT_KEY = secret_dns_tenant_key
 SECRET_DNS_TENANT_NAME = secret_dns_tenant_name
 SECRET_DNS_SERVICE_EMAIL = secret_registry_service_email
 SECRET_DNS_TENANT_ID = secret_dns_tenant_id
 
+# Configuration
+DNS_MOCK_EXTERNAL_APIS = dns_mock_external_apis
```
kimallen: I followed the existing pattern for env vars, but I wonder why we don't just initially set `DNS_MOCK_EXTERNAL_APIS` to the value of `dns_mock_external_apis`, so:
`DNS_MOCK_EXTERNAL_APIS = env.bool("DNS_MOCK_EXTERNAL_APIS", default=False)`

If you know or have context, lmk

erinysong:
Verified mock API when:

✅ Create record successfully
```app-1   | [30/Oct/2025 22:29:06] DEBUG [registrar.services.mock_cloudflare_service:242] 😃 mocking dns record creation
[Mock objects with mock UUIDs]
```

✅ Record with name prefix `error-403` ➡️ Get 403 error

✅ Record with name prefix `error-400`  ➡️ Get 400 error

✅ Record with name prefix `error` ➡️ Get 500 error

src/registrar/utility/constants.py
```diff
@@ -1,6 +1,9 @@
 from django.db import models
 
 
+DNS_ACCOUNT_NAME_PREFIX = "Account for "  # Must be constant since existing accounts are found by account name
```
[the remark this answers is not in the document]
erinysong: I think this makes sense! It lets us distinguish between what objects our DNS items are without needing to dig into its model type

src/registrar/views/domain.py
```diff
@@ -772,7 +772,7 @@ class PrototypeDomainDNSRecordForm(forms.Form):
 class PrototypeDomainDNSRecordView(DomainFormBaseView):
     template_name = "prototype_domain_dns.html"
     form_class = PrototypeDomainDNSRecordForm
-    valid_domains = ["igorville.gov", "domainops.gov", "dns.gov"]
+    valid_domains = ["igorville.gov", "domainops.gov", "dns.gov", "age-bit-low-far.gov", "exists.gov"]
```
erinysong: Nonblocking:
```suggestion
    valid_domains = ["igorville.gov", "domainops.gov", "dns.gov", "exists.gov"]
```

src/registrar/services/mock_cloudflare_service.py
```diff
@@ -0,0 +1,290 @@
...
+
+logger = logging.getLogger(__name__)
+
+fake = Faker()
+
+
+class MockCloudflareService:
+    _instance = None
+    _mock_context = None
+    fake_zone_id = fake.uuid4()
```
kimallen: I used uuids here because it was the only Faker id I found, but just fyi the ids that CF uses are not uuids

PR #4392, opened 2025-11-07
#4126 Script to clean up dns vendor accounts

### LINKED ISSUE #4126: Clean up Cloudflare accounts
### Issue description

Right now, every time we run the the dns prototype for creating a record (and an account and zone additionally if not yet created), we create an actually record in Cloudflare. There's also a number of existing accounts from when the prototype was initially set up.

Let's clean up unnecessary accounts (plus their zones and records). Note that not all accounts show up in the UI because of needing to be a member of each one to be viewable. All existing accounts will show up through hitting the `/accounts` api (though only by page).

We should also come up with a way to regularly clean up any created accounts going forward. This could involve either a script or a function that runs and deletes an account after we've created it and returned the desired data.

### Acceptance criteria

- [ ] Existing unnecessary accounts are removed (not sure any are necessary?)
- [ ] We have a means to delete accounts after they are no longer necessary

### Additional context

_No response_

### Links to other issues

_No response_

kimallen:
## Ticket

Resolves #4126

## Changes

- Adds a script with various ways to delete accounts. We can delete all, specific accounts by name or id, and "all except" by id

## Context for reviewers

Before mocking the apis, we were creating CF accounts from local dev every time we manually tested. This allows us to easily delete accounts as needed.

Once this is approved, I'll be able to actually delete the accounts. We can't actually delete any of the older accounts now until CF turns on some permissions temporarily, which I'll do as a next step.

## Setup

You will need to disable the mock apis to actually hit CF
    - in your `.env`, set `DNS_MOCK_EXTERNAL_APIS=False`
    - `docker compose stop app`
    - `docker compose build app`
    - `docker compose up app`

To test, check what you would delete by using the `--dry-run` arg with any of the other args
You could try actually deleting one of the random `account-*`. You will get a 404 for now because CF needs to turn on some permissions for any of our older accounts to actually access them for deleting.


abe-alam-ecs (approved):
Reviewed code. As I could not execute the commands (due to SSL issues), I had Erin run them for me. [See this slack thread](https://cisa-corp.slack.com/archives/C06K96473FA/p1762551773697499) for output.

There is one quirk, where Erin tried adding the same account id twice to see what happened and saw the script output this as 2 accounts. It should not affect anything as the ID is the same, but I'll leave this to Kim's discretion, whether it needs to be corrected or not. For the time being, I think it's more important that this script is accessible. 

LGTM!

kimallen:
> There is one quirk, where Erin tried adding the same account id twice to see what happened and saw the script output this as 2 accounts on the dry run. It should not affect anything as the ID is the same, but I'll leave this to Kim's discretion, whether it needs to be corrected or not. For the time being, I think it's more important that this script is accessible.

I believe if this was not a dry run, it would try to delete the account that was already deleted, so would return a 404 (or some sort of error response) since it no longer exists after the first one was deleted.

PR #4448, opened 2025-11-28
#4140: Check the database for an existing zone before checking CF [none]

### LINKED ISSUE #4140: Check the database for an existing zone before checking CF
### Sub-issue description

Right now, we find out if we have an existing account based on the zone name and account id by querying Cloudflare. Instead, now that we have the data stored in the db, check the database to see if the account already exists. If it does not, check CF. if it does not already exist, create the Cloudflare zone. If it does, save the data to the db.

### Acceptance criteria

- [ ] We check the db first to find the existing zones
- [ ] If the zone doesn't exist in the db, we check if it exists in cf, and if it does, we save the data in the db. If not, we create in CF and then save in db
- [ ] MockCloudflareService is updated as needed for the flow to work in local dev

### Additional context

_No response_

abe-alam-ecs:
## Ticket

Resolves #4140

## Changes

- Adds a method to check the db for an existing zone, called before checking cloudflare for the existing zone
- Adds `get_active_x_zone_id` as a model method on `DnsZone`
- Fixes a bug (discovered during this) where we were not saving the nameservers to the db
- Updates the cloudflare mocks to work with updated code
- Refactored some methods to simplify what values we are returning

## Context for reviewers

It's looking like we could use some test data utils (for creating accounts, zones, etc). I didn't do that here, though may either include that here or in a separate PR.

## Setup

The following should work in both local and sandbox env't (both need to have env't vars set up - see [readme](https://github.com/cisagov/manage.get.gov/blob/main/docs/developer/README.md#dns-hosting-feature)), though see updates in this PR. You are welcome to use my sandbox (kma) since I have the env't vars set up. In local just use the mocks)

Go to a managed domain for which you have not yet made a record. 
Create a record.
Check the logs to see that the account and zone were created in "cf" "Zone for domain {domain_name} does not exist".
Then submit a second record and see in the logs that they were not created in cf and the zone "Already has an existing zone"

### DIFF
--- a/src/registrar/services/dns_host_service.py
+++ b/src/registrar/services/dns_host_service.py
@@ -259,7 +259,9 @@ class DnsHostService:
                 dns_account = DnsAccount.objects.get(name=zone_account_name)
                 dns_domain = Domain.objects.get(name=domain_name)
 
-                dns_zone = DnsZone.objects.create(dns_account=dns_account, domain=dns_domain, name=zone_name, nameservers=nameservers)
+                dns_zone = DnsZone.objects.create(
+                    dns_account=dns_account, domain=dns_domain, name=zone_name, nameservers=nameservers
+                )
 
                 ZonesJoin.objects.create(
                     dns_zone=dns_zone,


src/registrar/views/domain.py
```diff
@@ -825,7 +825,7 @@ def post(self, request, *args, **kwargs):  # noqa: C901
                 domain_name = self.object.name
                 zone_id = ""
                 try:
-                    _, zone_id, nameservers = self.dns_host_service.dns_setup(domain_name)
+                    zone_id, nameservers = self.dns_host_service.dns_setup(domain_name)
```
erinysong: Is this meant to be the db `zone_id` or the cf `x_zone_id`? I noticed `dns_setup` still returns `x_zone_id` and wanted to double check

src/registrar/services/dns_host_service.py
```diff
@@ -146,16 +164,31 @@ def _find_existing_account_in_db(self, account_name):
...
+            zone = DnsZone.objects.get(name=domain_name)
+        except DnsZone.DoesNotExist:
+            logger.debug(f"Zone for domain {domain_name} does not exist")
+            return None, None
+
+        x_zone_id = zone.get_active_x_zone_id()
+        nameservers = zone.nameservers or []
+
+        # temporarily returning nameservers until we retrieve nameservers directly
         return x_zone_id, nameservers
```
erinysong: If we're searching for a zone in our database, is the primary object we're looking for the zone's CF data and is this why we want to return the `x_zone_id` here?

src/registrar/services/dns_host_service.py
```diff
@@ -195,6 +228,7 @@ def save_db_zone(self, vendor_zone_data, domain_name):
         x_zone_id = zone_data["id"]
         zone_name = zone_data["name"]
         zone_account_name = zone_data["account"]["name"]
+        nameservers = zone_data["name_servers"]
```
erinysong: Good catch adding these!

src/registrar/services/dns_host_service.py
```diff
@@ -146,16 +164,31 @@ def _find_existing_account_in_db(self, account_name):
...
             x_zone_id = self._find_id_by_name(zones, zone_name)
             nameservers = self._find_nameservers_by_zone_id(zones, x_zone_id)
+            zone_data = self._find_zone_json_by_name(zones, zone_name)
         except APIError as e:
             logger.error(f"Error fetching zones: {str(e)}")
             raise
 
+        return nameservers, zone_data
+
+    def _find_existing_zone_in_db(self, domain_name):
```
erinysong: I may need to just look at what we're doing here more, but is there a significance to getting and returning the CF `x_zone_id` of a zone if we're doing a db lookup? If yes, would we want to rename this function to clarify what we're doing?

PR #4599, opened 2026-01-21
#4243: UI: create validations for the Add Record form fields - [aa]

### LINKED ISSUE #4243: UI: create validations for the Add Record form fields (A Record only)
### Sub-issue description

For each MVP record type, they will share the name and content fields, but each type may have different validation criteria for those fields.

Write validations for DNS Record specific to an A record. Don't assume that what was written for the prototype is correct.

Looks like we should inherit from [ModelForm](https://docs.djangoproject.com/en/5.2/topics/forms/modelforms/#django.forms.ModelForm) for the form's class so that it applies to both the model and form

From [CF Docs](https://developers.cloudflare.com/dns/manage-dns-records/reference/dns-record-types/)
**name**
    Be 63 characters or less
    Start with a letter and end with a letter or digit
    Only contain letters, digits, or hyphens (underscores allowed but discouraged)

**content** for A record
    - checkout [validate_ipv4_address ](https://docs.djangoproject.com/en/5.2/ref/validators/#validate-ipv6-address)
    - required (for A type)

[**ttl**](https://developers.cloudflare.com/dns/manage-dns-records/reference/ttl/)
 For our plan type, we cannot currently proxy, so the options are 60 to 86400

Establish where we are keeping the validation code (model vs form vs validation class)

### Acceptance criteria

- [ ] content field is validated for an A record
- [ ] name field is validated for an A record
- [ ] The validations apply to both the form and the model (db) by using `ModelForm`
- [ ] All custom validations have tests

abe-alam-ecs:
## Ticket

Resolves #4243 

## Changes

- Reusable DNS name/label logic lives in `validations.py`
- record-level constraints enforced in DnsRecord.clean()
- updated Dns Record form to use ModelForm, and moved it to live with other domain forms
- clarified existing testing to explicitly test bad `name` and bad IPv4 `content`

## Context for reviewers

Copy/pasted from the ticket for ease of review: 

For each MVP record type, they will share the name and content fields, but each type may have different validation criteria for those fields.

Write validations for DNS Record specific to an A record. Don't assume that what was written for the prototype is correct.

Looks like we should inherit from [ModelForm](https://docs.djangoproject.com/en/5.2/topics/forms/modelforms/#django.forms.ModelForm) for the form's class so that it applies to both the model and form

From [CF Docs](https://developers.cloudflare.com/dns/manage-dns-records/reference/dns-record-types/)
name
Be 63 characters or less
Start with a letter and end with a letter or digit
Only contain letters, digits, or hyphens (underscores allowed but discouraged)

content for A record
- checkout [validate_ipv4_address](https://docs.djangoproject.com/en/5.2/ref/validators/#validate-ipv6-address)
- required (for A type)

[ttl](https://developers.cloudflare.com/dns/manage-dns-records/reference/ttl/)
For our plan type, we cannot currently proxy, so the options are 60 to 86400

Establish where we are keeping the validation code (model vs form vs validation class)

## Setup

Ensure dns_hosting flag is on
Navigate to a domain you manage
Click DNS > DNS Records
Add A records with invalid data to make sure the form cannot be submitted
Enter valid A records to ensure form is submitted

- [ ] screen reader (such as NVDA with Chrome or Voiceover with Safari):
  - [ ] landmarks
  - [ ] page header structure
  - [ ] links

- [x] screen reader (such as NVDA with Chrome or Voiceover with Safari):
  - [x] Tested general usability
  - [ ] landmarks
  - [x] page header structure
  - [x] links

- [ ] screen reader (such as NVDA with Chrome or Voiceover with Safari):
  - [ ] landmarks
  - [ ] page header structure
  - [ ] links
  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References


src/registrar/forms/domain.py
```diff
@@ -779,3 +780,100 @@ def __init__(self, *args, domain_state=None, **kwargs):
...
+        initial=300,
+        required=False,
+        widget=forms.Select(
+            attrs={
+                "class": "usa-select",
+            }
+        ),
+    )
+
+    comment = forms.CharField(
```
kimallen: From what I've [read](https://docs.djangoproject.com/en/6.0/topics/forms/modelforms/#overriding-the-default-fields), if you configure a field in the Meta class, the field definitions will match the model definition. Configuring them declaratively (as you did here) overrides the model field, and are generally used for more customization. 

It seems like some of the fields could use more customization and some do not need it. Is there a reason for all of these to be declared outside of the Meta class?

src/registrar/forms/domain.py
```diff
@@ -779,3 +780,100 @@ def __init__(self, *args, domain_state=None, **kwargs):
...
+        widget=forms.Select(
+            attrs={
+                "class": "usa-select",
+                "required": "required",
+                "x-model": "recordType",
+            }
+        ),
+    )
+
+    name = forms.CharField(
```
kimallen: To control the validation message on the client side, you can use `error_messages` attribute when configuring the field in the ModelForm. Would that work for the more customized messages that design has asked for?

abe-alam-ecs: Corrected successfully!

src/registrar/forms/domain.py
```diff
@@ -779,3 +780,100 @@ def __init__(self, *args, domain_state=None, **kwargs):
...
+            "name": forms.TextInput(attrs={"class": "usa-input"}),
+            "comment": forms.Textarea(
+                attrs={
+                    "class": "usa-textarea usa-textarea--medium",
+                    "rows": 2,
+                }
+            ),
+        }
+
+    type_field = forms.ChoiceField(
```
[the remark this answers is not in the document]
asaki222: Maybe we can rename it to something like `record_type`?

src/registrar/forms/domain.py
kimallen: After us chatting, and after doing a little more research, here's my suggestion:

1. For `comment` and `name`, it doesn't look like labels or validations change dependent on the record type. Those can get pulled into Meta (if `validate_dns_name` can be pulled onto the model field) or stay declared as is (if `validate_dns_name` has to remain called in the clean method).
2. Leave ttl where it is currently declared (since we need to use ChoiceField)
3. That leaves us with `content`, which is the only field whose attributes may change.  To determine attributes based on record type, we need to utilize __init__. Leave `content` here for now, and in the AAAA ticket (which upon re-reading is the one where we establish conditional logic), set up the `__init__`

kimallen (commented):
Suggested approach for final cleanup:
1. I left a [comment](https://github.com/cisagov/manage.get.gov/pull/4599#discussion_r2761443650) with suggestions for the ModelForm.
2. See if it's possible to move the `validate_dns_name` into the model field.
3. I think you're right about the need to do the validations for content in the Model.clean method, so leave `validate_ipv4_address` as is. I believe you will need to duplicate it in the Form.clean method in order to call it on the client side as well.

src/registrar/forms/domain.py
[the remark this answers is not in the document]
abe-alam-ecs: I renamed to type and am not getting the same errors I was getting previously. Something must have changed between Erin's HTMX updates and when I moved the form to a new file. It'll be `type`.

src/registrar/forms/domain.py
```diff
@@ -785,34 +786,40 @@ def __init__(self, *args, domain_state=None, **kwargs):
...
+                attrs={
+                    "class": "usa-input",
+                    "hide_character_count": True,
+                }
+            ),
+            "comment": forms.Textarea(
+                attrs={
+                    "class": "usa-textarea usa-textarea--medium",
+                    "rows": 2,
+                    "hide_character_count": True,
```
kimallen: Do we want to hide the character count on comment?  I think this is the one that includes it in the designs.

src/registrar/tests/views/test_domain_dns_record_and_form.py
```diff
@@ -118,3 +118,22 @@ def test_post_invalid_form_throws_error(self):
...
+            record_form["name"] = "testing!"
+            record_form["content"] = "192.0.2.10"
+
+            session_id = self.app.cookies[settings.SESSION_COOKIE_NAME]
+            self.app.set_cookie(settings.SESSION_COOKIE_NAME, session_id)
+            response = record_form.submit()
+
+            # Field assertions for A Type records. Additional tests will be needed for different types of records.
+            self.assertIn("Name", response.text)
+            self.assertIn("IPv4 Address", response.text)
```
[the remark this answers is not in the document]
abe-alam-ecs: I'm actually thinking the name test can be deleted now, because we test for name in test_dns_record.py

I'll adjust for IPv4 though, that's a good catch.

abe-alam-ecs:
> I am able to add records with a single letter. Was that intentional?

One letter is valid, @asaki222 !

kimallen (approved):
Great work! Thanks for all of your diligence and responsiveness and patience with feedback.

One non-blocking comment

PR #3628, opened 2025-03-10
(DRAFT) #2060- Upgrade Django FSM to Viewflow [backup]

allly-b:
## Ticket

Resolves #00

## Changes

- Change 1
- Change 2

#### Satisfied acceptance criteria and met development standards

- [ ] Met the acceptance criteria, or will meet them in a subsequent PR
- [ ] Created/modified automated tests
- [ ] Update documentation in READMEs and/or onboarding guide

#### Ensured code standards are met (Original Developer)

- [ ] If any updated dependencies on Pipfile, also update dependencies in requirements.txt.
- [ ] Interactions with external systems are wrapped in try/except
- [ ] Error handling exists for unusual or missing values

#### Validated user-facing changes (if applicable)

- [ ] Tag @dotgov-designers in this PR's Reviewers for design review. If code is not user-facing, delete design reviewer checklist
- [ ] Verify new pages have been added to .pa11yci file so that they will be tested with our automated accessibility testing
- [ ] Checked keyboard navigability
- [ ] Tested general usability, landmarks, page header structure, and links with a screen reader (such as Voiceover or ANDI)

#### Reviewed, tested, and left feedback about the changes

- [ ] Pulled this branch locally and tested it
- [ ] Verified code meets all checks above. Address any checks that are not satisfied
- [ ] Reviewed this code and left comments. Indicate if comments must be addressed before code is merged
- [ ] Checked that all code is adequately covered by tests
- [ ] Verify migrations are valid and do not conflict with existing migrations

#### Validated user-facing changes as a developer
**Note:** Multiple code reviewers can share the checklists above, a second reviewer should not make a duplicate checklist. All checks should be checked before approving, even those labeled N/A. 

- [ ] New pages have been added to .pa11yci file so that they will be tested with our automated accessibility testing
- [ ] Checked keyboard navigability
- [ ] Meets all designs and user flows provided by design/product
- [ ] Tested general usability, landmarks, page header structure, and links with a screen reader (such as Voiceover or ANDI)
- [ ] (Rarely needed) Tested as both an analyst and applicant user

#### Verified that the changes match the design intention

- [ ] Checked that the design translated visually
- [ ] Checked behavior. Comment any found issues or broken flows.
- [ ] Checked different states (empty, one, some, error)
- [ ] Checked for landmarks, page heading structure, and links

#### Validated user-facing changes as a designer

- [ ] Checked keyboard navigability
[...]

PR #4867, opened 2026-04-02
 #3723 - Django 5.2 Upgrade - [potato] 

### LINKED ISSUE #3704: Investigate alternatives to viewflow
### Issue description

We have determined Viewflow would be risky to adopt in place of django fsm, without providing many benefits. Somethings we can do instead are the following:

1. Use an alternative library
2. make our own fsm that we control and maintain
3. fork django fsm and maintain the folk (would need to handle dependencies
4. Keep using the archived version of django fsm (not ideal)

There may be additional options not listed above, and this ticket should look into what are alternatives to viewflow. This is purely a research ticket and not for making an ADR. (And ADR can come after a decision is made based on the research but is out of scope for this ticket)

### Acceptance criteria

- [ ] look at why django fsm was deprecated
- [ ] look at the django fsm code for understanding of what it's doing
- [ ]  research alternative options to viewflow
- [ ] sum up research

### Additional context

Note: I (Alysia) investigated some of the other library alternatives to django fsm and didn't see ones that seemed better suited to our needs. Also, I confirmed that github does not update dependencies or do dependabot checks for archived repos (ie option #4 should be avoided long term).

### Links to other issues

- came out of #2060

### LINKED ISSUE #3723: Upgrade Django and identify dependency issues
### Sub-issue description

When software like Django and Python are upgraded, very often there are dependencies that won't support the new versions.  We'll need a plan to address these dependencies.

### Acceptance criteria

- [ ] Read the[ Django 5.2 release notes](https://docs.djangoproject.com/en/5.2/releases/5.2/) to identify any possible issues
- [ ] Upgrade to Django 5.2 and Python 3.10
- [ ] Upgrade related dependencies
- [ ] Make recommendations for dependencies that can't be upgraded
- [ ] Manually test functionality to find any unexpected behaviors.  
    - At this point, a good starting point is to run through some UAT testing for switching (confirm that they are working though)
- [ ] Talk to Eng Leads about the plan

### Additional context
- We expect this evaluation to take multiple sprints so be sure to claim a shared sandbox

allly-b:
## Ticket

Resolves #3723 
Resolves #3704 

## Changes
-  session now stores portfolio as an integer ID instead of a model object to comply with Django 5.2's strict JSON session serialization
- Upgraded django-fsm to django-fsm-2 for Django 5.2 compatibility
- Added adr for why django-fsm-2 vs viewflow
- Replaced removed django.utils.timezone.utc with datetime.timezone.utc across management scripts and tests 
- Minor admin filtering update with what variable needs to be used.
- Fixed Concat() type error on mixed CharField/EmailField by adding explicit output_field=CharField()
- Fixed test deletion ordering to remove UserPortfolioPermission before Portfolio to satisfy relation constraints
- Resolved circular import in models/user.py by replacing direct Portfolio import with apps.get_model("registrar", "Portfolio")
- Renamed assertQuerysetEqual → assertQuerySetEqual (capital S) across tests to match the new django 5.2 recommendations
- Added a migration 
- Updated Pipfile and requirements.txt for all dependency version bumps & django-fms-2 addition (will likely need to reupdate it when Nicolle merges her PR)

## Context for reviewers

This is an large upgrade, so it's worth checking every flow you can think of. This includes admin filtering/ using the admin buttons/ making a submission etc. Try breaking it and look for anything off. 

Also, definitely read the ADR and give feedback if you have counter points to the suggested solution, or even just want clarity/ more info.

**type ignores**
for django-fsm-2, I ; added `# type: ignore[list-item]` on  FSM conditions= lists as django 5.2 throws errors with them not being cast to a specific type. Casting them seemed like it would add more complexity here, so I used the ignore as this PR already has a lot of changes. Let me know if people prefer trying to cast these.

**Migration**
The migration file is not indicative of a model actually changing. Django 5.2 requires any " through_fields" to be included in a migration. There was one of these fields in the User model, so django required a new migration just so everything matches it's new requirements. The model & functionality shouldn't have changed

**Session**
Django can't serialize a portfolio object into a json, for storing inside of `session,` so it throws an error and storing it as id is the least risky solution. The session change actually shouldn't be any issue. It looks like django has be silently erroring when storing portfolio data in the session. I ssh'd into a sandbox and printed the session contents for "portfolio" and each one gave me this print out `WARNING [django.security.SuspiciousSession:108] Session data corrupted`. I don't think changing the session to id will cause any issues when deployed to stable/staging.

### Additional context
Good 'ole [Medium article about migrating from 4.2-5.2](http://medium.com/@awaisq/upgrade-django-4-2-to-5-2-key-changes-d6b073709352)

[Rebecca's google doc ](https://docs.google.com/document/d/15cJcQwxlkamTIbJ_WjbY2TC1JdSkMr31X6jzEtsUIwg/edit?pli=1&tab=t.0)with steps I've gone through

viewflow  reference PRs:
https://github.com/cisagov/manage.get.gov/pull/3628
https://github.com/cisagov/manage.get.gov/pull/3670
https://github.com/cisagov/manage.get.gov/pull/3629

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons
  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References


docs/architecture/decisions/0030-django-state-control.md
```diff
@@ -0,0 +1,67 @@
+# 30. Django state control after upgrading to Django 5.2
```
CocoByte, line 1: This ADR is excellent!  Easy to follow and thorough on trade-offs.  Thanks for the dive into the technicalities and the write-up.

src/registrar/config/settings.py
```diff
@@ -977,7 +979,10 @@ def format(self, record):
 # session engine to cache session information
 SESSION_ENGINE = "django.contrib.sessions.backends.db"
 
-SESSION_SERIALIZER = "django.contrib.sessions.serializers.PickleSerializer"
+# SESSION_SERIALIZER = "django.contrib.sessions.serializers.PickleSerializer"
+SESSION_SERIALIZER = "django.contrib.sessions.serializers.JSONSerializer"
+# JSONSerializer is also default
+# FYI everyone will get logged out and will have to re login again
```
CocoByte, line 985: Just double-checking - Is the logout a one-time thing that happens when we deploy this update?

src/registrar/models/user.py
```diff
@@ -442,7 +443,13 @@ def is_org_user(self, request):
             DeprecationWarning,
             stacklevel=2,
         )
-        portfolio = request.session.get("portfolio")
+        portfolio_id = request.session.get("portfolio")
+        if portfolio_id:
+            Portfolio = apps.get_model("registrar", "Portfolio")
```
CocoByte, line 448: Thanks for explaining the reason for this in your PR :)

CocoByte:
Paths reviewed so far:
- Admin models sort and filter as expected (primarily tested Domains, Domain Reqs, Portfolios, and a few others).  Drill downs and modifications of fields appear to all still behave as expected
- Members flows (can add/remove members. Error messages that prevent removal of members with active domain requests fires as expected.  E-mail sending messaging behaves as expected.  Can edit domain assignments)
- Can view/edit domain requests 
- Can view/manage domains
- CSV exports from registrar appear to behave as expected and contain the right data (esp data from parent portfolio)
- embedded hyperlinks in /admin models that navigate between domain, domain req, users, and portfolios, along with hyperlinks for managing these in registrar all appear to work

CocoByte:
**Question (I think out of scope but want to double-check):**

When exporting "All domain metadata" from /admin analytics, I notice it only grabbed domains that were in the current portfolio in session....I can't remember if this is what we wanted or not?

CocoByte:
**Styling Change?**

The icons for editing /admin fields are larger than usual....

image

CocoByte:
**Bug?**

These toggles appear to no longer work.  Unsure if it is related to your changes...

image

docs/architecture/decisions/0030-django-state-control.md
```diff
@@ -0,0 +1,67 @@
...
+
+(#1) Use `django-fsm-2`
+
+The library is a direct drop-in replacement for the archived `django-fsm`, is already integrated into the project, and provides the same class of state protection as viewflow. The license, migration cost, and design fit all favor staying with django-fsm-2.
+
+## Considered Options
+
+### 1. django-fsm-2
+
+This would mean replacing the `django-fsm` import with a `django-fsm-2` import and adopting that library as a dependency. `Django-fsm-2` is a community-maintained fork of the original `django-fsm`, transferred to and maintained under [django-commons](https://github.com/django-commons/django-fsm).
```
erinysong: Is the link meant to direct to the link for [django-fsm-2](https://github.com/django-commons/django-fsm-2)? If it was originally meant for django-fsm, it looks like the link doesn't work anymore if the original fsm is no longer under django commons

docs/architecture/decisions/0030-django-state-control.md
```diff
@@ -0,0 +1,67 @@
...
+In this option, the state field is a FSMfield and transition logic is on the model.
+
++ Simply requires replacing the import in the pipfile-zero migration cost, same import path (`from django_fsm import FSMField, transition`)
```
erinysong, line 30: oh woah I was originally reading this as a rich diff so I can read more easily and the rich text format doesn't differentiate between `+`,`-`, or `*` when converting things into bullets. Would it be helpful to label the + and - lists under a Advantages/Risks header? 

Image

docs/architecture/decisions/0030-django-state-control.md
```diff
@@ -0,0 +1,67 @@
...
+### 2. viewflow FSM
+
+Adopt viewflow and add it as a dependency, reworking the code to be compatible with it's fsm implementation. [Viewflow](https://viewflow.io) is more than just an FSM it also provides "BPMN-style workflows", "process orchestration,"  and more, meant to be used for complex system workflows.
+
+In this option, the state field is a regular char field using enums (not an FSMfield) and transition logic is in a seperate flow class (not the model).
+
++ Actively maintained with a wide audience (2.9k GitHub stars, 416 forks, 365 dependent projects); with code commits as recent as Jan 2026
++ Supports Django 4.2–6.0
++ Uses Python enums for state definitions
++ Always enforces protected state at the Python descriptor level, meaning there ability to add `protected=False` and override functionality
```
erinysong: nit:
```suggestion
+ Always enforces protected state at the Python descriptor level, meaning there's no ability to add `protected=False` and override functionality
```

docs/architecture/decisions/0030-django-state-control.md
erinysong, line 1: thank you for compiling this! From what's on here I agree with moving forward with `django-fsm-2` - I don't see many immediate benefits refactoring to viewflow atm

src/registrar/config/settings.py
```diff
@@ -281,7 +281,8 @@
 ]
 
 # Stop using table-based default form renderer which is deprecated
-FORM_RENDERER = "django.forms.renderers.DjangoDivFormRenderer"
+# FORM_RENDERER = "django.forms.renderers.DjangoDivFormRenderer"
```
erinysong, line 284: jw is there a difference between removing these settings vs. commenting them out?

src/registrar/management/commands/transfer_transition_domains_to_domains.py
```diff
@@ -296,6 +296,9 @@ def try_add_domain_invitation(self, domain_email: str, associated_domain: Domain
 
         # check that the given e-mail is valid
         if domain_email is not None and domain_email != "":
+            # If domain hasn't been saved yet, no invitation can exist for it
+            if associated_domain.pk is None:
```
erinysong, line 300: what's the difference between checking if an object's pk doesn't exist vs. if the object doesn't exist? If a domain is mid-creation and not yet saved, does would that domain object still exist?

src/registrar/management/commands/transfer_transition_domains_to_domains.py
```diff
@@ -459,7 +462,7 @@ def update_or_create_domain_information(
         # Get associated domain
         domain_data = Domain.objects.filter(name=transition_domain.domain_name)
         if not domain_data.exists():
-            logger.warn(
+            logger.warning(
```
erinysong, line 465: good catch!

src/registrar/models/user.py
```diff
@@ -442,7 +443,13 @@ def is_org_user(self, request):
             DeprecationWarning,
             stacklevel=2,
         )
-        portfolio = request.session.get("portfolio")
+        portfolio_id = request.session.get("portfolio")
```
erinysong, line 446: hm thoughts on making a helper to get the portfolio object from session id if we're going to be grabbing it from session often?

src/registrar/tests/test_admin.py
```diff
@@ -283,7 +283,7 @@ def test_omb_analyst_filter_feb_domain(self):
         self.portfolio.save()
         response = self.client.get(
             reverse("admin:registrar_domaininvitation_changelist"),
-            {"status": DomainInvitation.DomainInvitationStatus.INVITED},
+            {"status__exact": DomainInvitation.DomainInvitationStatus.INVITED},
```
erinysong, line 286: what's the difference between explicitly defining this using `__exact` vs not? Are we excluding the possibility of other status domains included in the status param?

src/registrar/utility/csv_export.py
```diff
@@ -2255,6 +2257,7 @@ def get_computed_fields(cls, delimiter=", ", **kwargs):
                         "other_contacts__last_name",
                         Value(" "),
                         "other_contacts__email",
+                        output_field=CharField(),
```
erinysong, line 2260: nice!

src/registrar/views/domain.py
```diff
@@ -138,7 +140,8 @@ def _update_session_with_domain(self):
         update domain in the session cache
         """
         domain_pk = "domain:" + str(self.kwargs.get("domain_pk"))
-        self.session[domain_pk] = self.object
+        # self.session[domain_pk] = self.object
```
erinysong:
```suggestion
```

src/registrar/views/domain.py
```diff
@@ -482,7 +486,8 @@ def can_access_domain_via_portfolio(self, pk):
         """Most views should not allow permission to portfolio users.
         If particular views allow permissions, they will need to override
         this function."""
-        portfolio = self.request.session.get("portfolio")
+        portfolio_id = self.request.session.get("portfolio")
```
erinysong, line 488: similar to question on User model, would it help to make a helper to pull the portfolio from the session portfolio id? I'm all right with not since it's just 2 lines and noticed it's being used a lot here and in the portfolios view

src/registrar/models/domain_request.py
```diff
@@ -1177,7 +1177,7 @@ def submit(self):
             DomainRequestStatus.INELIGIBLE,
         ],
         target=DomainRequestStatus.IN_REVIEW,
-        conditions=[domain_is_not_active, investigator_exists_and_is_staff],
+        conditions=[domain_is_not_active, investigator_exists_and_is_staff],  # type: ignore[list-item]
```
erinysong, line 1180: for these list-items, what type would django 5.2 expect us to cast these as?

src/registrar/registrar_middleware.py
erinysong, line 1: this can be scoped in a later ticket, but thoughts on refactoring the `portfolio` session variable to `portfolio_id`?

erinysong:
✅ [Resolved] This is expected behavior since the expiration date is set only after a user manages their approved domain request

-------------------------
Submitted a new domain request for approval, then approved the domain request, and noticed the domain's Expiration date doesn't display on the Domains table

image

Though the expiration date displays on the domain's page:
image

erinysong:
**[✅ Resolved]** this happens on staging as well and I think it's because we have a field for all our domains so the long wait time occurs when we reference another table that has a really big number of entries

Not sure if this is happening on other envs, but I noticed the user domain role page seems to take a significantly longer time to load (few seconds more compared to other admin pages)
Starts off incomplete here with message `Transferring data from <environment URL>`
image

Then rest of User domain role form loads
image

Then rest of Django admin loads
image

erinysong:
**✅  [Resolved]** This is a separate issue with the second top navbar displaying on legacy mode when a user has legacy domains and at least 1 portfolio.


------
Tested making myself a manager for both domains from a portfolio and legacy domains. When I switch my session to view my legacy domains, the bottom navbar links for Domains/Domains Requests/Members don't appear

image
image


But they do appear when I switch my session to my portfolio:
image
image


^From my legacy domains session, I'm also unable to access the domain page to the domain I manage, most likely because I don't have permissions to access domains from here 🤔 
image

docs/architecture/decisions/0030-django-state-control.md
allly-b: Ohhhh ewww, it looks so bad in that format! Yes good ask, I will add that!

src/registrar/config/settings.py
allly-b: Going to leave this for now just in case it's causing the form to render differently (such as that button size thing you found). I won't resolve this comment though, this is a good reminder to delete all commented out items once this seems tested enough

src/registrar/config/settings.py
allly-b: This is a one time occurrence, and will only happen when we first deploy. I also changed session portfolio storage to be integer based, so any portfolio users would already have their sessions reset on the first deploy. In general, this upgrade just should happen outside of normal business hours and we should also put up the banner on get.gov for scheduled maintenance. But after this first deploy, we shouldn't need to worry about anything resetting etc.

docs/architecture/decisions/0030-django-state-control.md
allly-b: I think I manually added django-fsm into the link on accident. I meant to link to the django-commons page where it describes the maintainers of the project. Just updated!

src/registrar/management/commands/transfer_transition_domains_to_domains.py
allly-b: Good question this was a weird error. TLDR: Now with 5.2 the domain object exists but pk is null until save.


My early morning over explanation if you're extra curious: Django 5.2 doesn't add a pk to an unsaved model object, it will now only add pk on save BUT the actual object (domain in this case) is not none. So associated domain was filled with stuff, only pk was blank. Then, in line 304 it tries to filter DomainInvitations where domain=associated_domain. In django 4.2, this was fine and it would automatically return 0 results (as that is an unsaved object it can't have a foreign key relation yet). Now Django 5.2, basically says "you can't filter with an unsaved object". This is because, on the backend when we filter with domain=associated domain, django really does the logic of `domain.pk==associtated_domain.pk.` Which in this case means django would be doing domain.pk=None, and Django 5.2 thinks this is a waste of a query (it's picky).

src/registrar/models/user.py
allly-b: hmm... Not a bad idea on using a helper function. I think my only objection would be laziness for changing all the instances again lol. I'll add this to my todo list though to add it in!

src/registrar/models/domain_request.py
allly-b: `Callable[["DomainRequest"], bool]`, which gets a bit ugly to cast with in a list, and I find it makes it unreadable because it becomes `conditions =[ cast( Callable[["DomainRequest"], bool], domain_is_not_active),cast( Callable[["DomainRequest"], bool],  investigator_exists_and_is_staff)]`

not the worst, but doesn't seem to add value and looks messier in my book.

src/registrar/tests/test_admin.py
allly-b: ok tbh, this one drove me bonkers. This test was failing, and I couldn't figure out why, google suggested add __exact which worked, BUT Django docs say filtering with "status" and "status__exact" should be the same.... I kinda just threw my hands up in the air after trying other things and then moved on. But it SHOULD always be using exact even when we don't explicitly say it, unless we do something like __in (which checks if it's one of multiple options). And yes it excludes any other status, just like doing a sql select where status =xyz, it will exclude everything that's not xyz.

src/registrar/views/domain.py
allly-b: yeah, I think I was so focused on just getting it working and not failing tests, I wasn't even questioning how much repeat I was doing. I agree with the helper, I'll add it in shortly!

src/registrar/registrar_middleware.py
allly-b: Hmm I could see that helping with readability. If we want to do that, it would be best to do it now in this PR since changing the name would cause issues for logged in users. @CocoByte want to way in on your thoughts for renaming? I could go either way, though I am starting to lean towards Erin's suggestion.

src/registrar/models/user.py
erinysong: totally fair! I don't think it's an immediate need so we can also scope it in another ticket while we focus on Django functionality

src/registrar/management/commands/transfer_transition_domains_to_domains.py
erinysong: oh interesting, thank you for explaining! that's such a specific change to make and I wouldn't have guessed the object and pk get generated separately

src/registrar/models/domain_request.py
erinysong: oof yeah I'd also lean not casting here if we need to cast within the list items too

src/registrar/tests/test_admin.py
erinysong: yeahh I'm surprised in this case there's a difference when we use status and status__exact since I also would've expected status to default to status__exact when we don't pass in a lookup  :/

src/registrar/views/domain.py
erinysong: happy to also scope that in another ticket since I think it'd be good to test if the upgrade itself works + doesn't introduce breaking changes and then refactoring to convert things to helpers + refactor our portfolio session variable to portfolio_id maybe!

erinysong (approved):
Tested different flows in the request process and session management and things seem to be working the same as they did pre-upgrade. I think the only other changes I'd consider here would be updating any color palette settings on admin though will defer to design's expertise on that. Thank you working through this refactor and documenting!

I'm good with this getting merged once we rebuild our dependencies + any design convos about style updates

lizpearl:
@allly-b Going to through some dev parking lot and I see that this PR should also close 
- #3704 
so I added a resolves line with that ticket

CocoByte (approved):

Issue #4777, opened 2026-03-09
Bug: DNS record form comment field character count remains after previous is canceled


kimmsie:
### Current Behavior

If a user enters characters in the comment field, the character count decreases as expected. However, if a user cancels their form with existing text in the comment field and opens a new record form, the character count from the previously added canceled form remains.

https://github.com/user-attachments/assets/a7043a31-596e-49ab-926d-9bde4069b2de

### Expected Behavior

When a new record form is created, the character count should reset back to 500 and read "500 characters allowed."

### Steps to Reproduce

1. From the DNS records page, add a record.
2. Input some data into the comment field.
3. Without removing your data, select the cancel button.
4. Select the add record button to create a new record form.

### Environment

Environment: Glacier
Browser: Chrome

### Issue Links

- 🔄 Relates to: #4631

PR #4870, opened 2026-04-03
#4775 - Add DNS record form comment field character count changes during an error - [GLACIER]

### LINKED ISSUE #4775: Bug: Add DNS record form comment field character count changes during an error or when a new form is opened
### Current Behavior

After a user adds a record and a new add record form is started without refreshing the page, the comment input field's character count changes to "You can enter up to 500 characters" and does not count down as characters are added. The text color also changes to `#1b1b1b`. This also happens when the form has errors.

Image

### Expected Behavior

When adding a new record form, the character count helper text should say "500 characters allowed" and count down as the user types. The text color should be `#71767a`.

Image

### Steps to Reproduce

1. On the DNS records page, add a record.
2. Without refreshing the page, click the add record button to make a new record.

Also:
1. On the DNS records page, open the add record form.
2. Select save without entering any data.

### Environment

Environment: Glacier sandbox
Browser: Chrome

### Additional Context

When adding a form for the first time, it has the expected behavior. But after saving a record, the next add record form displays the bug

### Issue Links

- 🔄 Relates to: #4631

asaki222:
## Ticket

Resolves #4775 

## Changes

- On the dns form class, hide character count flag is turned on.
- Created a new div with the character count classes.
- Created a new function called CreateCharacterCount, which shares the same logic as the content character count in the txt form, and the comment section.
- Created another new function to `commentCharacterEventListener`, which adds the event listener for the input on each row, and the form to generate the text.
- Updated the over character limit text to match our other forms.
- Updated a class on the comment button that I had missed on a previous pr, which closes the gap between the caret and the comment icon on the dns records table.
- Fixed another bug where the content field did not swap from input to a textarea on firefox.
- Added code for accessibility to show the number of characters.
- Updated the txt content section to initially show "allowed" characters instead of "left".

## Setup

- Go to the /admin site > domain requests table > find a domain request with 'in review'
- Approve that domain requests
- Find the approved domain in the domains table. Enroll that domain in dns hosting
- Hit the manage
- Hit add a dns record, cause an error on the form.
- The character count on the box should still have the same styling as when you loaded the page. 
- Add a dns record of any type.
- Access the edit form
- The character count on the box should still have the same styling as when you loaded the page. 

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

- [x] Checked behavior. Comment any found issues or broken flows.

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons
  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References


erinysong:
will defer to @kimmsie on this but I noticed when testing and we submit an invalid record form, the character count flashes disappearing then suddenly reappears. Intuitively, I feel this is disruptive to the user experience since the character count is the only section that has this flashing disappear then reappear. Are there alternative ways to render the character count without this flashing?

https://github.com/user-attachments/assets/9061dc1c-6249-4549-a972-c84fc4eea3a0

erinysong:
> > will defer to @kimmsie on this but I noticed when testing and we submit an invalid record form, the character count flashes disappearing then suddenly reappears. Intuitively, I feel this is disruptive to the user experience since the character count is the only section that has this flashing disappear then reappear. Are there alternative ways to render the character count without this flashing?
> > Screen.Recording.2026-04-06.at.11.24.12.AM.mov
> > You are right there is a very slight flicker. We could add a fade in / out to make it less obvi? IDK though I think it would require some refactoring of the form handling and rendering as this part of input with errors. Is it a large enough issue for it's own ticket? Or is this something I should look into on this ticket? Thoughts @kimmsie
> 
> I can't seem to recreate the flicker. Which browser are you using? 🤔

@kimmsie this is on Firefox!

DaisyGuti:
> > > will defer to @kimmsie on this but I noticed when testing and we submit an invalid record form, the character count flashes disappearing then suddenly reappears. Intuitively, I feel this is disruptive to the user experience since the character count is the only section that has this flashing disappear then reappear. Are there alternative ways to render the character count without this flashing?
> > > Screen.Recording.2026-04-06.at.11.24.12.AM.mov
> > > You are right there is a very slight flicker. We could add a fade in / out to make it less obvi? IDK though I think it would require some refactoring of the form handling and rendering as this part of input with errors. Is it a large enough issue for it's own ticket? Or is this something I should look into on this ticket? Thoughts @kimmsie
> > 
> > 
> > I can't seem to recreate the flicker. Which browser are you using? 🤔
> 
> @kimmsie this is on Firefox!

I see this in Chrome but like barely visible, didn't notice it until Erin pointed it out 😆  @kimmsie @erinysong

DaisyGuti:
Adding the [Slack channel](https://cisa-corp.slack.com/archives/C06BZBJUUF8/p1775238884567689) for discussions on this

kimmsie (approved):
LGTM! Thanks Aditi!! 😺🎉

src/registrar/assets/src/js/getgov/domain-dns-record-content.js
```diff
@@ -1,5 +1,31 @@
 // Establishes javascript for dynamic content label based on type
 
+function getCharCountText (charLimit, textArea) {
+    const charactersLeft = charLimit - textArea.value.length;
+    if (charactersLeft >= 0) {
+      const characters = `character${charactersLeft === 1 ? '' : 's'}`;
+      return `${charactersLeft} ${characters} left`;
+    } else {
+      const characters = `character${Math.abs(charactersLeft) === 1 ? '' : 's'}`;
```
erinysong: Noticed we're using the same math to determine if there's a 1 char diff in both if blocks. Is there a reason we can't just calculate characters outside the if block and just calculate it using Math.abs(charactersLeft) === 1? Since we're only using `"character"` when there's a difference of 1 between the char limit and the user input

src/registrar/templatetags/custom_filters.py
```diff
@@ -337,3 +337,19 @@ def contact_text(contact, num_newlines=2):
     text = unescape(text)
     text = text.rstrip("\n")
     return text + "\n" * num_newlines
+
+
+@register.filter
+def remaining_characters_text(char_limit, field_value):
+    print("HELLO")
+    print(char_limit)
```
erinysong:
```suggestion

```

src/registrar/assets/src/js/getgov/domain-dns-record-content.js
```diff
@@ -57,17 +74,40 @@ export function editAndCommentButtonListener (){
         })
 }
 
+export function commentCharacterEventListener(){
+    function helperEventListener (element){
+        if(!element){
+            return;
+        }
+        const commentText = element.querySelector('.comment-character-count')
```
erinysong, line 81: What's the likelihood we reuse this pattern in future parts of our app? + what's your opinion on generalizing this event listener so it applies to a field in general and we don't need to commit the listener to a specific classname.
For example
```
function fieldCharacterCountListener(fieldClassName) {
   ...
   const commentText = element.querySelector(fieldClassName)
}
```
And then when we init in main.js we initialize that character count listener on inputs with classname comment-character-count
```
commentCharacterEventListener(.comment-character-count)
```

src/registrar/assets/src/js/getgov/domain-dns-record-content.js
```diff
@@ -57,17 +74,40 @@ export function editAndCommentButtonListener (){
         })
 }
 
+export function commentCharacterEventListener(){
```
erinysong, line 76: just curious and this may be more of a design question, what's the difference between using this event listener between input_with_errors and its char limit text? What would be the deciding factor to add this event listener to a form field over using input_with_errors?

erinysong:
I noticed in the ticket AC's the expected behavior says we want to initialize the char limit text to first say "X characters allowed" and once the user types, the char limit text becomes "X characters left." Is this AC still up-to-date?
image

Asking since I noticed the char limit text stays as "X characters left" when you open the form
image

src/registrar/assets/src/js/getgov/domain-dns-record-content.js
```diff
@@ -96,20 +135,22 @@ export function initDynamicDNSRecordFormFields() {
        
         if(selectedType == "TXT"){
             // Swap input type to text area
+            setTimeout(()=>{
```
erinysong, line 139: can you describe what we're doing by setting a timeout? Is this to address the flicker when we update the character count?

erinysong:
When I add a new record and type into it, then cancel adding the record, then open the Add new record form, the character count on the content resets but the character count on the comment does not

**Steps to reproduce:**
Fill in form to add new record. Click cancel
image

Click add record button again. Notice on the TXT form, the content character count resets but the comment character count has the same character count as from when you last filled the first record form. This is true on other record field types as well
image
image
image

src/registrar/templatetags/custom_filters.py
asaki222: ooh I missed this! Thank you

asaki222:
> I noticed in the ticket AC's the expected behavior says we want to initialize the char limit text to first say "X characters allowed" and once the user types, the char limit text becomes "X characters left." Is this AC still up-to-date? image
> 
> Asking since I noticed the char limit text stays as "X characters left" when you open the form image

@erinysong 

This is a separate bug. I picked up the ticket this sprint.

#4777

src/registrar/assets/src/js/getgov/domain-dns-record-content.js
asaki222: I am not sure the likelihood, we would use it but I can generalize it for future use.

asaki222:
> I noticed in the ticket AC's the expected behavior says we want to initialize the char limit text to first say "X characters allowed" and once the user types, the char limit text becomes "X characters left." Is this AC still up-to-date? image
> 
> Asking since I noticed the char limit text stays as "X characters left" when you open the form image

@erinysong when you rendered this, was this after you submitted the form? If so, I believe it is related to this ticket.

#4777

src/registrar/assets/src/js/getgov/domain-dns-record-content.js
asaki222: when I was testing, the input swap for the txt form did not happen on firefox. There is some delay from the dom content loading to the js grabbing the input. In the console, it would not find the "input". After adding setTimeout, it does find it.

src/registrar/assets/src/js/getgov/domain-dns-record-content.js
asaki222: I could take out the helperEventListener out of this method.  for the commentCharacterEventListener, it does runs this method per row, and for the form. I think that might too specific of a general use case.

src/registrar/assets/src/js/getgov/domain-dns-record-content.js
asaki222: The use case for this js function is to use it if an input needs a character count, and is involved in some htmx swap.

src/registrar/templates/domain_dns_record_edit_form.html
```diff
@@ -19,7 +20,16 @@ <h2 class="margin-top-0">Edit record</h2>
                 <input type="hidden" name="type" value="{{ dns_record.type }}" />
 
                 {% include dns_record.form_template with form=dns_record.form %}
-                {% input_with_errors dns_record.form.comment %}
+                <div>
+                    {% with add_group_class="usa-form-group--unstyled-error" %}
+                        {% input_with_errors dns_record.form.comment %}
+                    {% endwith %}
+                    <div class="usa-character-count__status usa-hint comment-character-count">
+                    <!-- Hardcorded value should be removed if/when the model is updated -->
```
erinysong: Can we specify here we'd remove the hardcode value once the dns record comment field's character limit is updated?

src/registrar/templatetags/custom_filters.py
```diff
@@ -337,3 +337,17 @@ def contact_text(contact, num_newlines=2):
     text = unescape(text)
     text = text.rstrip("\n")
     return text + "\n" * num_newlines
+
+
+@register.filter
+def remaining_characters_text(char_limit, field_value):
```
erinysong, line 343: Can we use the same logic we're using in the USWDS/custom js character count here as well for consistency?

erinysong:
Accessibility and UI: On VO when I initially open an Add record form, the comment field VO updates so that we are given the character count as we type. However, when we fail to submit a record form and get prompted with errors, the VO no longer includes the character count and the character count bottom text does not update the value since the original form submission

https://github.com/user-attachments/assets/bbd3ffeb-cc38-44de-86d4-a88452649be6

erinysong:
Similar to the above, when we submit an edit record form with errors, the character count resets to the character count value of the form when the record was last successfully saved, and not the character count of the most recently submitted comment text. In the example below, we edit a DNS record with an initial comment with 48 characters left. When we update the comment then submit an invalid edit to that record, the character count resets to 48 despite our changes

https://github.com/user-attachments/assets/ec5d9028-1206-460f-9a15-f2935c6af199

src/registrar/templates/domain_dns_record_edit_form.html
asaki222: I added specifics to the message. I have here

erinysong (changes requested):
One final change request: Can we update the message so that when a user hits the max character count the message is `0 characters left` to match USWDS rather than `0 characters over limit`?

Image

Image

erinysong (approved):
✅ Good with this being merged - thank you for making updates and answering questions!

PR #4902, opened 2026-04-15
#4745: Update validations for all DNS record name fields [dg]

### LINKED ISSUE #4745: Update validations for all DNS record name fields
### Sub-issue description

Add and/or verify the inclusion of the following validations;

All DNS Records:
Name field
- Name cannot contain consecutive dots (..), or start or end with a dot.
- The record name must not exceed 253 characters when fully qualified. _(Note: the zone name is appended automatically before this limit is evaluated.)_
- Name cannot contain spaces.
- The hyphen character is not allowed at the start or end of a label (e.g., '-example' and 'example-' are invalid). _(Note: cf API allows hyphens anywhere.  This is recommended for consistency with RFC 1035)_
- Name contains an invalid character. The following characters are not allowed: @ ( ) : ;  
    - Exception: A single '@' is accepted as a shorthand for the zone apex (equivalent to leaving the name empty).
- A wildcard is only valid as the entire leftmost label (e.g., *.example.com or *.sub.example.com). Wildcards cannot appear mid-label or in a non-leftmost position (e.g., sub.*.example.com is invalid). _(Note: The Cloudflare API does not currently enforce this. This validation is recommended for UI consistency with Cloudflare documentation.)_
- **(Backend only)** DNS record names are case insensitive for validations and display (Cloudflare silently normalises the name to lowercase on save and displays it in lowercase in the record list.)

Display the appropriate message in the UI (See [Error messages spreadsheet](https://docs.google.com/spreadsheets/d/1hF7_n_zo3EpfbgqqqPUMtSMZXynMtRMsBMlcIqZ91ns/edit?gid=1995836989#gid=1995836989))

**Consider splitting this into multiple PRs if some ACs become more complex**

### Acceptance criteria

**For all Record Name fields (DNS Name) — applies to A, AAAA, CNAME, MX, PTR, TXT:**

**Please cross check with the spreadsheet, which is the source of truth for error messages. Update AC here if you find they diverge**

- [x] **(Backend only)** DNS record names are case insensitive (All DNS record names are normalized to lowercase before validation, storage, or being sent to Cloudflare.)
- [x] **(Backend only)** Name fields are validated and stored in their "fully qualified" format (that is, they have the zone name appended to them -- eg. if zone is "fed.gov", then the a name of "my.domain" becomes "my.domain.fed.gov"). Be aware that appending the domain name may be more complex in cleaning on the model level.
- [x] When a name exceeds 63 characters per label and/or 253 characters _when fully qualified_ (i.e. after the zone name is appended), the error **"Name label must be no more than 63 characters. Full name (including label, domain, dots, and zone) must be no more than 253 characters."** is displayed on save.
- [x] When the name field is empty, the error **"Enter the name of this record."** is displayed on save.
- [x] When a name contains spaces, the error **"Enter the DNS name without any spaces."** is displayed on save.
- [x] When a name contains any of the following, the error **"Enter the name using only letters, numbers, hyphens, periods, or a single @ symbol."** is displayed on save:
  - Consecutive dots (`..`), or a leading/trailing dot
  - A hyphen at the start or end of a label (e.g. `-example` or `example-`)
  - Invalid characters (e.g. `@ ( ) : ;`)
         - Exception: A single `@` is accepted as shorthand for the zone apex and should not trigger an error.
- [x] This validation and corresponding message should be removed: `Enter a name that begins with a letter and ends with a letter or number.`

### Additional context

_No response_

### Links to other issues

_No response_

DaisyGuti:
## Ticket

Resolves #4745 (rescoped). Related: #4945 (error-message copy).

## Changes

Rewrites DNS record `name` validation to match the rescoped AC. All record types (A, AAAA, CNAME, MX, PTR, TXT) share the same name validator.

**`registrar/validations.py`**
- Split leading/trailing-dot from consecutive-dot into separate checks, each with its own constant.
- Added `validate_dns_name_fqdn_length(name, zone_name)` so the 253-char limit is evaluated against the **fully qualified** name — the zone is auto-appended for relative names and for the `@` apex shorthand, matching Cloudflare.
- Retained RFC 1035 allowlist regex (`[a-zA-Z0-9.*-]`) — matches Cloudflare's accepted charset.
- Error copy (from the product copy spreadsheet):

| Constant | Copy |
| --- | --- |
| `DNS_NAME_FORMAT_ERROR_MESSAGE` | `Enter the name without using parentheses, colons, or semicolons.` |
| `DNS_NAME_CONSECUTIVE_DOTS_ERROR_MESSAGE` | `Enter the name without using consecutive periods.` |
| `DNS_NAME_LEADING_TRAILING_DOT_ERROR_MESSAGE` | `Enter the name without using consecutive periods.` |
| `DNS_NAME_HYPHEN_ERROR_MESSAGE` | *placeholder — pending final copy* |
| Length (label > 63 or FQDN > 253) | `Labels must be no more than 63 characters. Full name (including labels, domain, and period) must be no more than 253 characters.` |
| Space in name | `Enter the name without any spaces.` |

**`registrar/forms/domain.py`**
- Empty-name copy → `Enter the name of this record.`
- New `_validate_name_fqdn_length` hook calls the zone-aware length check using `self.domain_name`.

**`registrar/models/dns/dns_record.py`**
- `name` is normalized to lowercase on save (case-insensitive storage/display, matches Cloudflare).
- `name.max_length=255` (migration `0178`) so the custom validator messaging wins over Django's field-level max-length error.

The retired message `Enter a name that begins with a letter and ends with a letter or number.` is already gone from the codebase.

## Error message map

| Trigger | Field error |
| --- | --- |
| Empty name | `Enter the name of this record.` |
| Space | `Enter the name without any spaces.` |
| Label > 63 chars or FQDN (after zone append) > 253 chars | see length copy above |
| Consecutive dots (`..`) | `Enter the name without using consecutive periods.` |
| Leading or trailing dot | `Enter the name without using consecutive periods.` |
| Leading or trailing hyphen on any label | *placeholder — pending final copy* |
| Invalid chars (e.g. `( ) : ;`, `$`, `_`, etc.) | `Enter the name without using parentheses, colons, or semicolons.` |

Exception: a lone `@` is accepted as the zone-apex shorthand.

## Setup
1. Ensure the `dns_hosting` waffle flag is ON and the domain is enrolled in DNS hosting.
2. Run migrations: `docker compose exec app ./manage.py migrate`.
3. Available on [DG sandbox](https://getgov-dg.app.cloud.gov/domain/15075/dns/records).

## Test plan

Trigger each case on the add/edit record form:

- **Empty** → "Enter the name of this record."
- **Space** (`ab cd`) → "Enter the name without any spaces."
- **Consecutive dots** (`ab..cd`) → consecutive-periods copy
- **Leading dot** (`.abc`) / **trailing dot** (`abc.`) → consecutive-periods copy
- **Leading hyphen** (`-abc`) / **trailing hyphen** (`abc-`) / mid-label boundary (`my.-domain`, `my-.domain`) → hyphen placeholder
- **Invalid chars** (`ab$c`, `ab(cd`, `ab)cd`, `ab:cd`, `ab;cd`) → parentheses/colons/semicolons copy
- **`@` alone** → valid (apex)
- **Uppercase** (`ABC`) → valid; stored lowercase
- **Label > 63** (`<64 a's>`) → length error
- **FQDN > 253 via zone append** — on a zone like `example.gov`, enter a 245-char `a`s relative name → length error (245 + "." + "example.gov" = 257)
- **Already fully qualified** (`www.example.gov` on `example.gov`) → passes; not double-appended

## Code review checklist
- [ ] Distinct error constants per trigger (consecutive dots, leading/trailing dot, hyphen, invalid chars)
- [ ] `validate_dns_name_fqdn_length` called from form `clean()` with `self.domain_name`
- [ ] Model field validator still catches direct `Model.save`/admin paths
- [ ] `name.max_length=255` intentional so custom messaging wins
- [ ] Test coverage: empty, `@`, uppercase, wildcard, spaces, invalid chars, dot-structure, hyphen, >63 label, >253 relative-after-append, already-qualified

## Follow-ups
- Replace `DNS_NAME_HYPHEN_ERROR_MESSAGE` with finalized copy once available from #4945.

- [x] [Follow the process for requesting a design review](https://dhscisa.enterprise.slack.com/docs/T02QH7E1MHA/F06CZ6MRUKA). If code is not user-facing, delete design reviewer checklist
- [x] Reviewed [accessibility checklist](https://github.com/cisagov/manage.get.gov/blob/main/docs/dev-practices/code_review.md#accessibility-checklist) using screen reader (such as NVDA with Chrome or Voiceover with Safari), ANDI, or WAVE:
  - [x] Tested general usability
  - [x] Page header structure
  - [x] Landmarks
  - [x] Links and buttons
- [x] Checked for errors or warnings with an a11y browser tool (such as ANDI or WAVE)

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons
  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References


src/registrar/models/dns/dns_record.py
```diff
@@ -25,9 +25,9 @@ class DnsRecord(TimeStampedModel):
     )  # type: ignore
 
     type = models.CharField(choices=DNSRecordTypes.choices)
-
+    # Purposefully not enforcing max_length here in favor of custom DNS validation messaging
     name = models.CharField(
-        max_length=253,
+        max_length=255,
```
kimallen, line 30: Can you say the reason the max_length is being changed here?

erinysong:
Noticed when testing for record name's case conversion that when we go use the Cloudflare tenant, we lowercase our names which makes sense since we're getting the lowercased record data from Cloudflare

Image
Image

But when I test on local, since we're getting a response back from our Mock Cloudflare lib instead of Cloudflare, the record name maintains any uppercase letters on the backend and UI:
Image

Since when our app is connected to the Cloudflare tenant, the record name lowercases, do we want to update our clean (or Mock Cloudflare lib but I think if we follow the pattern on sandbox I'm leaning just updating our clean method) so we lowercase the response?

erinysong:
## ✅  Name label must be <=63 characters
Failed to submit record with a label of 64 characters and received appropriate error message
image

## ❓ Domain name <= 253 characters
When I input a record name sans zone name on local, I don't get a form error message and instead get an error in the model's create since the name + the appended zone name exceeds 255 characters
```
app-1   | [20/Apr/2026 20:35:33] ERROR [registrar.services.dns_host_service:262] Failed to save record {'type': 'A', 'name': 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.aaaaa', 'content': '194.4.4.4', 'ttl': 300, 'comment': None} in database: value too long for type character varying(255)
```
On dg I also just get an overall error that we failed to save the DNS record
image

## ❓ Empty name field
Nit: `"Enter the name for this record."` ➡️ `"Enter the name of this record."`
image

## ✅ Record name must not contain spaces
image

## ✅ Record name must not contain consecutive periods
Content question: Thoughts on having the error message specify that the record name must contain nonconsecutive periods? @montecristo99 
image

## ✅ Name must not start or end with a hyphen
image
image
image

## ✅ Record name must not contain special characters with exception of root @ and wildcard *
image
image
image
image

## ❓ Wildcard * and root @ symbols allowed
I noticed that on cloudflare we allow names with a wildcard * (example: `*name`, `middle*asterisk`, `name*`, and `double**asterisk` all lead to valid responses on CF) but we don't allow that on the registrar. Should we update wildcard handling so that it matches CF?
image
image
image

src/registrar/tests/forms/test_dns_record.py
```diff
@@ -90,24 +91,91 @@ def test_blank_dns_name_throws_error(self):
...
+        self.assertTrue(form.is_valid())
+
+    def test_dns_name_with_wildcard_at_start_valid(self):
+        """Wildcard as entire first label should be valid (e.g., *.example.com)."""
+        data = self.valid_form_data_for_record_type("A", self.VALID_CONTENT_BY_TYPE["A"])
+        data["name"] = "*"
+        form = self.make_form(data)
+        self.assertTrue(form.is_valid())
+
+    def test_dns_name_with_wildcard_not_at_start_throws_error(self):
```
erinysong, line 142: I noticed that these test values pass when passing them as names in the CF API. Is this a restriction we're setting only on our end on the registrar or should we update them to reflect CF?

Image

Image

src/registrar/tests/models/dns/test_dns_record.py
```diff
@@ -114,3 +115,170 @@ def test_clean_mx_with_priority_valid(self):
...
+            record.full_clean()
+        self.assertIn("name", ctx.exception.message_dict)
+
+    def test_dns_record_name_exceeds_total_fqdn_limit_raises(self):
+        """DNS record name exceeding 253 characters when fully qualified should fail validation."""
+        from django.core.exceptions import ValidationError
+
+        # Create a name that when combined with zone exceeds 253 chars
+        long_name = "a" * 250
+        record = DnsRecord(
```
erinysong, line 223: Can we separate this test value into groups of "a" * 63? That way we're testing that the string is adhering to the rule that each label has length < 64

src/registrar/tests/test_validations.py
```diff
@@ -0,0 +1,56 @@
...
+
+        for name in valid_names:
+            with self.subTest(name=name):
+                validate_dns_name(name)
+
+    def test_validate_dns_name_rejects_names_with_spaces(self):
+        self.assert_dns_name_validation_error("ab cd", "Enter the DNS name without any spaces.")
+
+    def test_validate_dns_name_rejects_format_errors(self):
+        invalid_names = [
```
erinysong, line 35: hm looks like I was able to get the following values to pass as successful names on CF. Also wanted to check if certain restrictions were on our end or meant to match CF

`-abc` is valid on CF
Image

`abc-` is valid on CF
Image

`ab$c` is valid on CF
Image

`sub.*` is valid on CF
Image

`a*b` is valid on CF
Image

`my.-domain` is valid on CF
Image

`my-.domain` is valid on CF
Image

src/registrar/tests/test_validations.py
```diff
@@ -0,0 +1,56 @@
...
+
+        for name in invalid_names:
+            with self.subTest(name=name):
+                self.assert_dns_name_validation_error(name, DNS_NAME_FORMAT_ERROR_MESSAGE)
+
+    def test_validate_dns_name_rejects_labels_over_63_characters(self):
+        self.assert_dns_name_validation_error("a" * 64, DNS_NAME_LENGTH_ERROR_MESSAGE)
+
+    def test_validate_dns_name_rejects_names_over_253_characters(self):
+        self.assert_dns_name_validation_error("a" * 254, DNS_NAME_LENGTH_ERROR_MESSAGE)
```
erinysong, line 56: same ask here - can we have this string separated into substrings of a*63 + separated by however many periods till we have 254 characters?

erinysong (changes requested):
thank you for working on these validations! Left some notes and comments and will also post CF findings on slack thread

src/registrar/models/dns/dns_record.py
DaisyGuti: Sure, listed in my PR Changes list: "Custom validator enforces business logic (253 char max) and runs first" 

It is so our custom checks and messaging can kick in prior to Django's default based on the model.

src/registrar/tests/forms/test_dns_record.py
DaisyGuti: Per the ticket: A wildcard is only valid as the entire leftmost label (e.g., *.example.com or .sub.example.com). Wildcards cannot appear mid-label or in a non-leftmost position (e.g., sub..example.com is invalid). (Note: The Cloudflare API does not currently enforce this. This validation is recommended for UI consistency with Cloudflare documentation.)

src/registrar/tests/models/dns/test_dns_record.py
DaisyGuti: Added more to cover the case you mention.

src/registrar/tests/test_validations.py
DaisyGuti: From our meetings it was my understanding, this is a combination of what is allowed per the RFC's and CF. 

I will ask that we that if we're going to change scope and/or ACs we please create a ticket to refactor all of this work to match the new requirements/ACs.

I will let Nicole and Kim chime in. @CocoByte @kimallen

src/registrar/tests/forms/test_dns_record.py
kimallen: Apologies @DaisyGuti for crossed messages. Here's a [link](https://cisa-corp.slack.com/archives/C074E2QJ7J9/p1775520061253459?thread_ts=1775503686.196209&cid=C074E2QJ7J9) where we discussed the wildcards. Looks like we removed the AC item from the ticket and the error item from the alerts spreadsheet, but neglected to remove the content about wildcards in the description. * are allowed anywhere in the name. We are following restrictions based on CF's api behavior, and don't want to create additional restrictions that are not dictated by the RFC.

src/registrar/models/dns/dns_record.py
kimallen: So if you keep it at 253, it hits the model error first, and doesn't ever hit our custom check? It's surprising if there's not some other way to get custom messaging. It feels a little hacky to change the model limit. Do you see any other options for having the custom messaging override the model default messaging?

src/registrar/tests/test_validations.py
kimallen: @DaisyGuti - did this get resolved by the re-refinement of the ticket?

DaisyGuti: We had a 30 min convo yesterday during our pod mtg on this ticket. Tara is tagged in the ticket and for now I'm adding placeholder ticket, while she decides if she wants: 

1. hierarchy in error messaging
2. error messaging text specific to certain cases (on the ticket)
3. multiple error messages to display (tied to # 1 likely)

erinysong:
Updated Fri 4/24
# ✅ Empty → "Enter the name of this record."
image

# ✅ Space (ab cd) → "Enter the name without any spaces."
image

# ✅ Consecutive dots (ab..cd) → consecutive-periods copy
image

# ❓ Leading dot (.abc) / trailing dot (abc.) → consecutive-periods copy ➡️ let's address this with content in another scope outside this PR
Error message matches the error spreadsheet but was wondering if the error message for this is accurate/should be updated later since a leading/trailing period isn't a consecutive period
image
image

# ✅ Leading hyphen (-abc) / trailing hyphen (abc-) / mid-label boundary (my.-domain, my-.domain) → hyphen placeholder
image
image
image
image

# ✅ Invalid chars (ab$c, ab(cd, ab)cd, ab:cd, ab;cd) → parentheses/colons/semicolons copy
image
image
image
image
image

# ✅ Commas
[update 4/24] - other special characters excluding explicitly invalid ones (parentheses, colons, semicolons) work
image
image

[4/23]
Noticed that CF allows commas in their response but we don't include it in our allowed validations and our error messages indicate commas aren't an excluded character. Should we add it to the allowed regex chars in our DNS record name?
image
image

# ✅ @ alone → valid (apex)
record name @ converts to root name

# ✅ Uppercase (ABC) → valid; stored lowercase
image
image

# ✅ Label > 63 (<64 a's>) → length error
image

FQDN > 253 via zone append — on a zone like example.gov, enter a 245-char as relative name → length error (245 + "." + "example.gov" = 257)
Entered name that would surpass 253 when zone name appended
image

# ✅ Already fully qualified (www.example.gov on example.gov) → passes; not double-appended
image
image

src/registrar/validations.py
```diff
@@ -56,28 +57,128 @@ def get_max_length_attrs(limit: int) -> dict[str, str]:
...
-DNS_NAME_FIELD_REGEX = re.compile(r"^[a-zA-Z0-9.-]+$")
+DNS_NAME_FORMAT_ERROR_MESSAGE = "Enter the name without using parentheses, colons, or semicolons."
+DNS_NAME_CONSECUTIVE_DOTS_ERROR_MESSAGE = "Enter the name without using consecutive periods."
+DNS_NAME_LEADING_TRAILING_DOT_ERROR_MESSAGE = "Enter the name without using consecutive periods."
+DNS_NAME_HYPHEN_ERROR_MESSAGE = "placeholder error message for leading or trailing hyphen"
+DNS_NAME_LENGTH_ERROR_MESSAGE = (
+    "Labels must be no more than 63 characters. "
+    "Full name (including labels, domain, and period) must be no more than 253 characters."
+)
+DNS_NAME_VALID_CHAR_REGEX = re.compile(r"^[a-zA-Z0-9.*-]+$")
```
erinysong: would it make sense to also add the `,` character here since CF seems to allow it and our error messages don't seem to indicate they're not allowed?

comma in middle of name
Image

trailing comma
Image

leading comma
Image

comma before label
Image

src/registrar/tests/test_validations.py
```diff
@@ -0,0 +1,91 @@
...
+class TestValidateDNSNameFQDNLength(SimpleTestCase):
+    ZONE = "example.gov"
+
+    def assert_error(self, name: str, zone: str | None) -> None:
+        with self.assertRaises(ValidationError) as ctx:
+            validate_dns_name_fqdn_length(name, zone)
+        self.assertEqual(ctx.exception.messages, [DNS_NAME_LENGTH_ERROR_MESSAGE])
+
+    def test_no_error_when_zone_missing(self):
+        validate_dns_name_fqdn_length("a" * 300, None)
```
erinysong, line 86: here are the fqdn and the labels in the fqdn treated as separate entites? For example can a valid fqdn technically have labels > 63 characters? Or are we just treating just the validators for total fqdn length and individual label lengths separately?

src/registrar/models/dns/dns_record.py
DaisyGuti: I'll take a deeper look.

src/registrar/tests/test_validations.py
DaisyGuti: Both. Separate and together they must be <= 253 chars, each label (dot seperated string) must be <= 63.

That's why we have two distinct validators:

_validate_dns_name_label_length — runs on each label of the entered name.
validate_dns_name_fqdn_length — runs on the entered name + appended zone, so the 253 limit is enforced against the true FQDN the user is creating, not just what they typed.

DaisyGuti: So no — a "valid" FQDN cannot technically have a label > 63 chars.

src/registrar/models/dns/dns_record.py
DaisyGuti: Now I recall the reason I bumped the model to 255 was so the FQDN-after-zone check could kick in, since the model's max_length only sees the raw entered string and can't account for the appended zone. But I think I found a way. Trying now!

src/registrar/validations.py
DaisyGuti: Good catch that the messaging could be addressed in https://github.com/cisagov/manage.get.gov/issues/4945. . 

The AC lists those explicitly disallowed but doesn't say they're the only disallowed ones.

On whether to allow, specifically, I'd lean toward keeping it rejected. RFC 1035's preferred name syntax is letters, digits, and hyphens only; , isn't part of it. Cloudflare being permissive doesn't mean it's safe or correct.

DaisyGuti: FYSA @montecristo99

DaisyGuti: On messaging: I'd suggest it either (a) lists the allowed character set positively ("Use letters, numbers, hyphens, and periods.") rather than trying to enumerate everything disallowed, or (b) echoes back the first offending character so the user can see what the validator tripped on.

DaisyGuti:
> Since when our app is connected to the Cloudflare tenant, the record name lowercases, do we want to update our clean (or Mock Cloudflare lib but I think if we follow the pattern on sandbox I'm leaning just updating our clean method) so we lowercase the response?

ood call — since the lowercase-on-save behavior was introduced on this branch, it makes sense to close the loop on the read path here too rather than leave it for a follow-up. @erinysong

src/registrar/tests/test_validations.py
DaisyGuti: @kimallen It’s still a bit of a gray area. My current implementation follows the RFCs, which are stricter than Cloudflare's API. I'm hesitant to open up validations to match Cloudflare's 'incorrect' behavior without a formal change to the AC.

Given the work already completed, would it be better to merge the current RFC-compliant logic and use follow-up tickets for any specific edge cases (like commas or mid-label wildcards) once Product confirms they want to support them? 

Regarding error messaging could be addressed in [#4945](https://www.google.com/search?q=https://github.com/cisagov/manage.get.gov/pull/4945) too. 

Echoing in Slack.

kimmsie (approved):
LGTM! 🎉

src/registrar/tests/test_validations.py
kimallen: @DaisyGuti - I'd prefer that since we spent the time together refining the ticket, that we stick with the ticket. The ticket's requirements are what we want to follow. If you interpret the RFC differently, we can talk about that, but let's stick with the ticket.

src/registrar/validations.py
kimallen: The AC we laid out in the ticket for allowed characters (Invalid characters (e.g. @ ( ) : ;)) is meant to mean that they are the only disallowed characters. That's why the message was written that way, and why it was changed from the original message that the regex matched. So anything but the disallowed characters is ok. We've already done a lot of back and forth, so let's stick with the intent of the ticket (not RFC's "preferred name syntax").

DaisyGuti: Updated to be like the AC

erinysong (approved):
LGTM - thank you for answering my questions and explaining the updated ACs!

PR #4979, opened 2026-05-04
Restore request button and remove banner

h-m-f-t:
This PR restores the button to submit new requests and drops the banner.

* See #4700 and #4767
* See also cisagov/get.gov#540


lizpearl (approved):
Looks good!

Deployed on sandbox:
- Checked that banner was removed
- Button worked and was restored in legacy and enterprise mode

allly-b (approved):
Love the name "unshutdown." LGTM, just ran it locally.

Issue #4999, opened 2026-05-12
[P1] Create and approve user-facing copy for our defined error codes


DaisyGuti:
### Issue Description

When someone edits their domain's DNS records and something goes wrong, we show them a short message. This ticket is for writing — and approving — the exact words in those messages in these situations where a DNS change can fail, in two kinds of ways:

1. Things the person can fix (like a duplicate record) — the message shows right next to the field they were editing.
2. Things they can't fix (like the DNS provider being down) — we show a general message with a reference code they can give to support.

This is a one-time sign-off before launch — wording can still be adjusted later without breaking anything, so the goal is just to get the launch wording right.

| The situation | Draft wording | Where it shows |
|---|---|---|
| We can't find the DNS area for this domain | "We can't find the DNS zone. It might not be set up yet." | By the field |
| A record with that name already exists | "A record with that name already exists. Names must be unique." | By the field |
| The data they entered isn't valid (a rule the provider catches that our form didn't) | Show the provider's reason when it's clear; otherwise "That data isn't valid." | By the field || They're making changes too quickly | "You're going too fast. Wait a moment and try again." | By the field / banner |
| We're not allowed to make the change | shared banner (below) | Page banner |
| We couldn't reach the provider in time | shared banner (below) | Page banner |
| The DNS provider is having problems | shared banner (below) | Page banner |
| Something unexpected happened | shared banner (below) | Page banner |

Shared "page banner" wording:
> "We couldn't reach our DNS provider. Please try again in a moment. If the problem persists, share this reference with support: `1a2b3c4d-5e6f-7890-1234-567890abcdef`."

### Acceptance Criteria
- [ ] Final wording approved for each situation in the table above.
- [ ] Decision on "the data isn't valid": show the provider's own reason when it's clear, or write our own wording
- [ ] Shared "not your fault" banner wording approved.
- [ ] Approved wording handed to engineering (#4950).

### Additional Context
- This ticket is Product/Content work — writing and approving the message. 
- #4950 is the engineering follow-up.

### Issue Links
- Related: #4892

### Links to other issues
- 🚧 Blocks #4950

h-m-f-t:
@kimallen this is slotted for sprint 18. Can you confirm we want to prioritize this and size it?

kimallen:
> [@kimallen](https://github.com/kimallen) this is slotted for sprint 18. Can you confirm we want to prioritize this and size it?

[...]

Issue #4997, opened 2026-05-12
[P2] Tighten register_nameservers error handling


DaisyGuti:
### Issue Description
The goal here is to use our structured-logging pattern: on failure, we log the values in the call's `extra`= so OpenSearch can filter on them instead of them being buried in an f-string. 

This should focus on [`register_nameservers` in DnsHostService ](https://github.com/cisagov/manage.get.gov/blob/e882737562060538b7aa504f26c6bbc96ebee5be/src/registrar/services/dns_host_service.py#L288-L297)catches `(RegistrySystemError, Exception) `and raises. Apply a narrower catch and structured-log pattern, catch only the registry exceptions it can raise, let others propagate, and log failures with structured fields.

### Acceptance Criteria
- [ ] `register_nameservers` catches the specific registry exceptions it can raise, and other exceptions propagate.
- [ ] Failure log lines carry these 3: domain_name (which domain failed), nameservers, and error_class (the exception type name, e.g. `RegistryError` Which answers "what kind of failure" in OpenSearch) as structured fields.
- [ ] This method also has a `# TODO: first check domain state? or status?` resolve it (implement the check or remove it if it's not needed comment on it)
- [ ] Lint and tests pass.

### Additional Context
Catch RegistryError, the registry failure this method owns and log it. We allow / want any others to propagate (don't catch them): NameserverError / ActionNotAllowed, Domain.DoesNotExist etc. all worth surfacing. A catch-all would log all of them as registry failures and hide potential real bugs. So instead we might try something like:
```
try:
    domain.nameservers = nameserver_tups          # the registry call
except RegistryError as e:                         # the one we care about for this services purpose
    logger.error("...", extra={"domain_name": ..., "error_class": type(e).__name__})
    raise
# NameserverError, ActionNotAllowed, Domain.DoesNotExist etc. are intentionally NOT caught here.
# They travel up to the caller for handling

```

### Links to other issues
- ⛔️ Is blocked by #4924

erinysong:
> Apply the narrow-catch and structured-log pattern this epic uses elsewhere, catch only the registry exceptions it can raise, let anything else propagate, and log failures with structured fields.

Can we link the pattern(s) we want to establish here so it's easier to find for whoever ends up working on this ticket?

DaisyGuti:
nameserver_count? @kimallen @lizpearl @CocoByte @katypies I added this but might be extra info / not super helpful. Please edit to what you feel might be best for debugging purposes.

[...]

Issue #4996, opened 2026-05-12
[P1] Register DNS models with django-auditlog


DaisyGuti:
### Issue Description
Today /admin has no record of which DNS records changed or who changed them, so a "who changed this record?" question has nothing to look at. Register DnsRecord, DnsZone, and DnsAccount with `django-auditlog`, following the existing pattern (e.g. Contact). That gives a who/what/when history for every successful DNS model change.

### Acceptance Criteria
- [x] A successful create / update / delete of a DnsRecord, DnsZone, or DnsAccount appears at /admin/auditlog/logentry/ with the acting user, timestamp, and field diff.
- [ ] Lint and tests pass (no new tests needed)
- [x] When a new record is created, one can view the related audit log

### Additional Context
Files: models/__init__.py (add the auditlog.register(...) calls).

Scope: successful create / update / delete on the three models — a failed call leaves the model unchanged, so auditlog records nothing for it.

### Links to other issues
- 🔄 Relates to #4927

Issue #4950, opened 2026-04-24
[P2] Wire approved user-facing copy into _error_mapping


DaisyGuti:
### Issue Description
Wire the [approved DNS error copy](https://docs.google.com/spreadsheets/d/1hF7_n_zo3EpfbgqqqPUMtSMZXynMtRMsBMlcIqZ91ns/edit?gid=896778226#gid=896778226) into the per-code error dict in utility/errors.py, replacing the prototype's draft strings, so each error code has its approved message.

### Acceptance Criteria
- [x] Every DnsHostingErrorCodes value has its approved string
- [x] The generic 5xx message (the one that carries the reference id) uses the approved wording.
- [ ] Consider adding additional mocked returned status codes to the CF mocks for testing
- [x] Lint and tests pass - assure we add test around the codes matching the expected messaging/copy.

### Additional Context
- Copy is owned by this ([spreadsheet](https://docs.google.com/spreadsheets/d/1hF7_n_zo3EpfbgqqqPUMtSMZXynMtRMsBMlcIqZ91ns/edit?gid=896778226#gid=896778226)).
- Draft strings to replace: the prototype table ([errors.py#L367-L382]

Note: there is a bug where the edit form closes after a banner error is displayed, and upon reopening it, there is no data in the form (Starts with just the record type form). To be handled in another ticket.

### Links to other issues
- ⛔️ Is blocked by #4999

montecristo99:
Approved wording is in the ticket tab in the alert messaging spreadsheet [here](https://docs.google.com/spreadsheets/d/1hF7_n_zo3EpfbgqqqPUMtSMZXynMtRMsBMlcIqZ91ns/edit?gid=896778226#gid=896778226).

PR #5001, opened 2026-05-12
#4893: Add DNS error handling developer guide (Phases 1 & 2)

DaisyGuti:
## Ticket

Resolves #4893

## Changes

- Adds `docs/developer/dns-error-handling.md` — developer guide for DNS error handling scoped to Phases 1 & 2 only
- Documents the typed DnsHostingError hierarchy, the 8 error codes, the canonical JSON envelope, the httpx timeout + retry policy, and the "services raise, views catch" rule
- Maps every formal AC from #4893 and every consideration in the planning ticket description to a section of the doc
- Lists the Phase 1 and Phase 2 sub-tickets (#4920–#4925, #4928, #4950, #4996, #4997, #4999) so reviewers can trace each piece of the doc to a ticket

## Context for reviewers

Doc-only PR. No code changes.

The team asked to keep this PR scoped to Phases 1 & 2 only. Sections that describe Phase 3+ capabilities (admin OpenSearch deep-links, admin-editable copy, structured logging fields, support runbook) were cut so this doc only covers what's actually getting built right now. Those phases are tracked under epic #4892 and #5005 will be re-evaluated after Phases 1 & 2.


docs/developer/dns-error-handling.md
```diff
@@ -0,0 +1,377 @@
...
+- **`DnsHostService`** (`src/registrar/services/dns_host_service.py`) — wraps Cloudflare exceptions with `raise APIError(str(e)) from e` in a couple of methods. This re-wraps the same error type and produces a second log line for the same failure.
+- **View layer** (`src/registrar/views/domain.py`) — catches `(APIError, RequestError)` and shows a generic `messages.error("Failed to save DNS record.")`. JSON responses use two different shapes (`{"status": "error", "message": ...}` and `{"error": ...}`).
+- **Exception classes** (`src/registrar/utility/errors.py`) — `NameserverError`, `DsDataError`, and `SecurityEmailError` already follow a typed code-based pattern. DNS hosting has only a bare `APIError(Exception)` with no codes, no subclasses, no status propagation. We mirror the `NameserverError` pattern.
+- **Logging** — `logging_context.py` uses ContextVars for `user_email`, `ip_address`, `request_path`, but the middleware that sets them is gated on `IS_PRODUCTION`. `DatabaseConnectionMiddleware` reads `HTTP_X_REQUEST_ID` directly but doesn't share it. DNS service logs are f-strings with no `request_id` or `error_code`.
+
+---
+
+## The 8 Error Types
```
kimallen: Given that this will be just general documentation, I think this doc eventually can start here (though maybe with an intro?) and everything before this can be removed (once built, devs won't need to know the info above). I think it's reasonable to keep it in here for now to reference before we build it. Could you create a separate ticket to update this doc after phase 1 and 2 are implemented to remove the above and other parts so it can just be a current-practice reference doc?

docs/developer/dns-error-handling.md
```diff
@@ -0,0 +1,377 @@
...
+
+When Cloudflare says "no," here's what it means:
+
+| What went wrong | Error name | User sees |
+|---|---|---|
+| Zone doesn't exist | `DNS_ZONE_NOT_FOUND` | "We can't find the DNS zone. It might not be set up yet." |
```
kimallen: What do you think about just putting "TBD" for all the error messages (except the known "record already exists" message)? I wouldn't want people to mistake these messages as final copy. You could add an AC to Tara's content ticket #4999 to update the message column here with the final approved copy.

docs/developer/dns-error-handling.md
```diff
@@ -0,0 +1,377 @@
...
+Tell user what happened (friendly message + request_id)
+```
+
+Every request gets a unique `request_id` from `RequestLoggingMiddleware` (#4924). The same ID flows into every log line for the request and into the JSON error envelope returned to the browser, so a user-reported reference ID can be traced back to the exact server-side events.
+
+---
+
+## What Developers Do
+
+### The full failure flow (Cloudflare → service → view)
```
kimallen: Thanks for adding in this full example!

docs/developer/dns-error-handling.md
```diff
@@ -0,0 +1,377 @@
...
+    return exc_cls(code=code, upstream_status=status, context=ctx)
+
+
+class CloudflareService:
+    def create_dns_record(self, zone_id, record_data):
+        url = f"/zones/{zone_id}/dns_records"
+        try:
+            resp = self.client.post(url, json=record_data)
+            resp.raise_for_status()
+        except HTTPStatusError as e:
```
kimallen, line 232: What if there is an APIError here? Should that also be included?

Also, should there be a fallback catch for if it's neither an APIError nor an HTTPStatusError?

docs/developer/dns-error-handling.md
```diff
@@ -0,0 +1,377 @@
...
+Python's `from e` chain preserves the original `HTTPStatusError` in the traceback, so engineers searching by `request_id` in logs can still see Cloudflare's response body, status, and `cf-ray`.
+
+### 1. Raise Specific Errors
+
+```python
+# Don't:
+raise APIError("something went wrong")
+
+# Do:
+raise DnsNotFoundError(
```
kimallen: This `raises DnsNotFoundError`, while the example above `raises _typed_dns_error(. . .)` which in the end returns a DnsNotFoundError, though looks different. I wonder if the different way the examples are written will be confusing? May be ok as is?

docs/developer/dns-error-handling.md
kimallen: I'm thinking this "The full failure flow" section might work better below the 4 steps. That way this section starts with high level understanding, then shows the more detailed example of it in practice.

kimallen (commented):
This is looking really great! Thanks for all this documentation. I included a few suggestions.

docs/developer/dns-error-handling.md
DaisyGuti: I created ticket #5012

docs/developer/dns-error-handling.md
DaisyGuti: done

docs/developer/dns-error-handling.md
DaisyGuti: restructured

docs/developer/dns-error-handling.md
DaisyGuti: Great question!

For APIError that's the generic exception. This whole effort is essentially replacing that.

Once [#4921](https://github.com/cisagov/manage.get.gov/issues/4921) (typed errors in CloudflareService)  is done, the cf svc will raise the typed DNSHostingError. [#4922](https://github.com/cisagov/manage.get.gov/issues/4922) (drops the duplicate wrapping in DnsHostService) takes the re-wrap  out of the dns svc. So by the time the view hits in [#4925](https://github.com/cisagov/manage.get.gov/issues/4925)  (standardize to the DNS API envelope), APIError is gone from the DNS path. DnsHostingError is the only catch we need.

I added a note in the code to make this clear.

For a fallback except, I left that off on purpose. If something other than a DnsHostingError reaches the view (a DatabaseError, a TypeError from a bug, etc.), Django's default 500 handler picks it up and rolls back the request's transaction, which is exactly what we want.

Once [#4928](https://github.com/cisagov/manage.get.gov/issues/4928)  (request_id on the 500 page) completes, the user gets a ref ID to share, and the full traceback lands in OpenSearch under that same ID. A catch-all here would step on that flow and hide bugs we'd rather see.

Let me know if that makes sense or if you want me to dig in deeper anywhere!

docs/developer/dns-error-handling.md
DaisyGuti: Good call — they really are doing the same thing, just from different starting points. 

Added a short blurb to Rule #1 to explain when to use each: direct raise DnsNotFoundError(...) when you already know which condition you're handling (you're checking something explicitly and want to signal a specific failure), and _typed_dns_error(e, ...) when you've caught an HTTPStatusError from cf and need to map the status code to the right subclass. 

The helper exists so the status-to-subclass mapping lives in one table rather than getting repeated in every service method.

docs/developer/dns-error-handling.md
```diff
@@ -0,0 +1,332 @@
...
+    429: (DnsRateLimitError,  DnsHostingErrorCodes.RATE_LIMIT_EXCEEDED),
+}
+
+
+def _typed_dns_error(e: HTTPStatusError, **context) -> DnsHostingError:
+    """Map a Cloudflare HTTP error to the right DnsHostingError subclass and log once."""
+    status = e.response.status_code
+    ctx = {"cf_ray": e.response.headers.get("cf-ray"), **context}
+    exc_cls, code = _STATUS_TO_ERROR.get(status, (DnsHostingError, DnsHostingErrorCodes.UNKNOWN))
+    logger.error("Cloudflare returned %s for DNS request", status, extra={"upstream_status": status, "error_code": code.name, **ctx})
```
lizpearl: Can you document somewhere what parts of this logging message is supported by cloud.gov?

docs/developer/dns-error-handling.md
lizpearl: What changes in code do we need to get that stack trace in our production logs? and what does it look like?  I'm not asking for a copy to be put in this document but I want to make sure that we see all the data that we're expecting

docs/developer/dns-error-handling.md
```diff
@@ -0,0 +1,332 @@
+# DNS Error Handling — A Developer's Guide
+
+**Last updated** 2026-05-15
+
```
CocoByte, line 6: I found reading this .md to be very data-heavy and would love to have an intro section that tells me at a high level how to approach this as a dev.  For example, something like...


```suggestion
## Intro

For DNS error handling we decided to create an admin table that maps custom error codes (that we maintain) to their corresponding error messages. This is to simplify the complexity associated with hardcoded error messages scattered throughout the code. By centralizing error codes and their messages in a single admin table (the error mapping), we not only keep a centralized record of our error messages, but also make it easier to maintain such that non-devs with access to this table could make error messaging adjustments (which were once only possible by doing a code edit).

**How to use it in the code:** When something goes wrong (Cloudflare is down, a zone doesn't exist, the network times out), raise a typed DnsHostingError subclass from the service layer. Using this architecture, the error code stays attached as you bubble up through the domain layer into the view. The view then catches it, looks up the user-facing message in our error mapping, and returns a consistent JSON envelope with the error code, message, and a request_id for logging and support.

**What's in this doc:**

- **Error Types** — the 8 error codes we handle today, what triggers each, and how to look them up
- **What Developers Do** — the 4 rules: raise specific errors, catch only in views, include context, test by code not by message
- **Worked Example** — end-to-end flow of one failed DNS save to show the rules in action
- **Network Timeouts & Retries** — why we timeout, how long we wait, and when we retry (write operations never retry; read operations do with backoff)
- **Key Files** — where the error classes, services, views, and logging live

**Maintaining this doc:**  This guide covers what's being built right now. Future phases will add deeper features that will result in this doc requiring an update (including changing the "last updated" date at the top of the doc).
```

CocoByte: ^ I know kim mentioned the possibility of adding an intro later as well, so if this is out of scope, feel free to hold off!  Also, this is just an example of what makes sense to me so no pressure to apply it.

P.S. Copilot was used to help create the "How to use it in the code" and "What's in this doc" sections, with human edits

CocoByte: (cc: @kimallen)

CocoByte (approved):
I understand this doc will be evolving over time and after reading through things I don't have much to add outside of Kim's comments except maybe adding an intro for ease of digesting the information.

Approving with non-blocking recommendation (Add an intro?).

docs/developer/dns-error-handling.md
```diff
@@ -0,0 +1,345 @@
...
+**What's in this doc:**
+
+- **Error Types** — the 8 error codes we handle today, what triggers each, and how to look them up
+- **What Developers Do** — the 4 rules: raise specific errors, catch only in views, include context, test by code not by message
+- **Worked Example** — end-to-end flow of one failed DNS save to show the rules in action
+- **Network Timeouts & Retries** — why we timeout, how long we wait, and when we retry (write operations never retry; read operations do with backoff)
+- **Key Files** — where the error classes, services, views, and logging live
+
+**Maintaining this doc:**  This guide covers what's being built right now. Future phases will add deeper features that will result in this doc requiring an update (including changing the "last updated" date at the top of the doc).
+How DNS errors are handled:
```
CocoByte:
```suggestion

**How DNS errors are handled (summary):**
```
(minor formatting for readability -- looks like I forgot to add a space at the bottom of my commit suggestion)

CocoByte (commented):
minor formatting suggestion!

docs/developer/dns-error-handling.md
```diff
@@ -0,0 +1,346 @@
+# DNS Error Handling — A Developer's Guide
+
+## Intro
+
+For DNS error handling we decided to create an admin table that maps custom error codes (that we maintain) to their corresponding error messages. This is to simplify the complexity associated with hardcoded error messages scattered throughout the code. By centralizing error codes and their messages in a single admin table (the error mapping), we not only keep a centralized record of our error messages, but also make it easier to maintain such that non-devs with access to this table could make error messaging adjustments (which were once only possible by doing a code edit).
```
kimallen: I'd exclude the second part of this since we are not implementing that yet (and there's not been a decision whether we will or not). It definitely may get implemented later, but how much later I don't know and don't want to confuse people until that happens.

```suggestion
For DNS error handling we decided to create an admin table that maps custom error codes (that we maintain) to their corresponding error messages. This is to simplify the complexity associated with hardcoded error messages scattered throughout the code. 
```

docs/developer/dns-error-handling.md
kimallen: Thanks for the explanation and added note in the doc- that makes sense!

kimallen (approved):
Updates look good to me- one requested update to the intro, but approving otherwise!

docs/developer/dns-error-handling.md
```diff
@@ -0,0 +1,346 @@
...
+## Exception Contract
+
+Two rules for `DnsHostingError` and its subclasses:
+
+1. **Only simple values on the exception.** `str`, `int`, dict-of-primitives. The parallel test runner pickles exceptions across processes; an `httpx.Response` object or a lambda in `context` breaks the pickle and the test run fails in confusing ways.
+2. **The code, not the string, is the source of truth.** `_error_mapping` in `errors.py` provides the default user-facing copy. Tests assert on `exc.code` rather than the message string.
+
+---
+
+## Network Timeouts & Retries
```
lizpearl, line 328: Need a comment here that this pattern is not to be used until testing is done for this

docs/developer/dns-error-handling.md
```diff
@@ -0,0 +1,346 @@
...
+httpx.Timeout(connect=3, read=10, write=10, pool=5)
+```
+
+- `connect=3` — Cloudflare's edge is normally sub-second. 3 seconds covers transient issues without hanging.
+- `read=10` / `write=10` — DNS changes are small payloads. 10 seconds is enough.
+- `pool=5` — `httpx` keeps a small pool of open HTTP connections to Cloudflare and reuses them across requests (so we're not opening a fresh TCP connection every time). The `pool=5` timeout sets the longest a request will wait for a free connection from that pool. If it's waiting more than 5 seconds, it means we have more concurrent DNS calls happening than the pool can serve. Either:
+  - the pool is too small for our actual load, OR
+  - something is holding connections longer than expected (a slow Cloudflare day, a connection leak in our code, an unexpected spike).
+
+  We want this to fail fast instead of queuing silently — the timeout fires, `httpx` raises a `PoolTimeout` (a `RequestError`), our service raises `DnsTransportError(code=UPSTREAM_TIMEOUT)`, and the failure shows up in the logs. That's the signal to investigate (increase the pool size, check for a leak, or look at concurrency).
```
lizpearl: I would change queuing -> waiting for a free connection and also make it clear that this is tied to the pool=5s configuration.  When you used the word `queueing` I immediately thought to myself that we don't have queueing in our application

lizpearl (changes requested):
Nice edits!  

The only overall comment that I have is to check that sections that are still to be implemented say that in the documentation so devs that are using this right now don't get confused

docs/developer/dns-error-handling.md
DaisyGuti: There were good questions brought up around how exceptions are handled and logged. After re-evaluating (future management cmd callers, admin view, etc.)  I've made adjustments to our approach and hope to have addressed clarity around logging in this next revision for your review.

Please let me know if you think I need to go further or adjust any sections.

docs/developer/dns-error-handling.md
DaisyGuti: Done

docs/developer/dns-error-handling.md
DaisyGuti: Mentioned the two tickets related to doing the work and monitoring. Please let me know if we would like to approach the documentation of this work a different way.

docs/developer/dns-error-handling.md
DaisyGuti: I had a difficult time understanding your question but I hope I addressed it in my response here and the latest revision of the doc.

Opensearch dynamically maps JSON fields, unless we purposely block the default behavior. To verify we are not blocking the behavior I ran a test, I raised an exception using the proto branch on my dg instance. 

Please view a quick example of that (IS_Prod = false so some fields don't show yet, that is in the Phase 1 tickets to do.) below:

Image

We can see some of the items the logger added: app.user_email, app.lineno, ...

I have added to and, clarified ticket #4924 in Phase 1 so hopefully there is no ambiguity. This ticket is already flagged as requiring external DNS Pod review and review by an Eng Lead.

DaisyGuti: Please let me know if you meant something else? Or could you rephrase the ask please?

lizpearl (approved):
Looked over the changes and looks good!

PR #5014, opened 2026-05-15
#4861: Validate hostname content for CNAME, MX, and PTR records [glacier]

### LINKED ISSUE #4861: Update validations for PTR, CNAME, and MX content fields
### Sub-issue description

In #4745 we updated the logic for validating DNS name fields.  We need to apply that same validation to PTR, CNAME, and MX content fields, with some alterations.  

_Differences between name field validation and PTR, CNAME, MX content field validation:_
- Content fields do NOT have zone name appended to the end of it before validation.  Therefore, the 253 character limit is for the entire user-input string (instead of the string+parent zone)
- Content fields allow trailing dots, which are silently stripped in the backend
- The rightmost label of content fields cannot consist entirely of digits (e.g., 'host.123' is invalid)

Display the appropriate message in the UI (See [Error messages spreadsheet](https://docs.google.com/spreadsheets/d/1hF7_n_zo3EpfbgqqqPUMtSMZXynMtRMsBMlcIqZ91ns/edit?gid=1995836989#gid=1995836989))

### Acceptance criteria

**For PTR, MX, and CNAME Content fields:**

- [ ] _Inherits all DNS name validations listed in #4745 with the following additions/alterations:_

(Logical changes)
- [ ] **(Backend only)** Unlike the name field, content field is validated against the raw user input, without zone name appended. (Cloudflare does NOT append zone name to the end of content fields before vaildation and storage)
- [ ] **(Backend only)** Unlike the name field, trailing dots in content field is allowed (Cloudflare strips trailing dots from content field user input (eg. `test.gov.` becomes `test.gov`) before vaildation and storage).  The content field is stored with the trailing dot stripped.

(Specific wording to distinguish content fields from name fields)
- [ ] When the MX mail server content fails a DNS name validation, replace "name" in the resulting error message with "mail server"
- [ ] When the PTR domain name field fails a DNS name validation, replace "name" in the resulting error message with "domain name"

_Additional validations that only apply PTR, MX, and CNAME Content fields:_
- [ ] When the content field's rightmost label consists entirely of digits (e.g. `host.123`), the error **"Enter a domain name using only letters."** is displayed on save.

- [ ] TODOs in `registrar/tests/forms/test_dns_record.py` are completed and removed

### Additional context

_No response_

erinysong:
## Ticket

Resolves #4861 

## Changes

- Validate hostname content (records with type CNAME, MX, PTR) for the following
   - [x] No spaces
   - [x] No consecutive periods
   - [x] No leading periods (a single trailing period is accepted)
   - [x] Last label is not a number
   - [x] Includes none of the following characters: `@`, `(` `)` `:` `;`. Exception is a single `@` which is treated as the zone root name.
   - [x] Hostname length does not exceed 253 characters
   - [x] Hostname labels (eg: `label.label.label`) do not exceed 63 characters

- On some validations messages, transition from using error message enums to validation requirement enums. Motivation is that we frequently reuse the same error message for different fields (name, content) and content field differs by record type (mail server, domain name, target, etc...). Switching from constant error messages to just the validation's requirement lets us be more flexible with what field we're validating.
- Update Mock Cloudflare API to convert `@` hostname content to zone name ([Slack discussion](https://cisa-corp.slack.com/archives/C074E2QJ7J9/p1778014183950379))
- Remove validating against record names that start and end with hyphens, given that CF allows record names and labels to start and end with hyphens ([Slack discussion](https://cisa-corp.slack.com/archives/C074E2QJ7J9/p1778017137680319))

## Setup

On the DNS records page, test the following validations work on a DNS record form for a hostname content (CNAME, MX, PTR) record. 
CNAME errors should refer to the field as `target`. MX errors should refer to the field as `mail server`. PTR records should refer to the field as `domain name`.

   - [x] Submit a hostname record with content including spaces, for example : `a b`. Expected error message: `Enter the <field> without any spaces.`
   - [x] Submit a hostname record with content including consecutive periods, for example : `a..b`. Expected error message: `Enter the <field> without using consecutive periods.`
   - [x] Submit a hostname record with content beginning with a periods, for example : `.a`. Expected error message: `Enter the <field> without using consecutive periods.` **Note: This is the current error message for trailing/leading period values on the validations spreadsheet. We should revisit the language of this in the future.**
   - [x] Submit a hostname record with a number as the last label, for example : `a.123`. Expected error message: `Enter content that ends with a domain name.`
   - [x] Submit a hostname record with any of the characters `@`, `(` `)` `:` `;`. , for example : `a(`. Expected error message: `Enter the <field> without using parentheses, colons, or semicolons.`
   - [x] Submit a hostname record with content exceeding 253 characters : `123456789012345678901234567890123456789012345678901234567890123.123456789012345678901234567890123456789012345678901234567890123.123456789012345678901234567890123456789012345678901234567890123.12345678901234567890123456789012345678901234567890123456789012d`. Expected error message: `Labels must be no more than 63 characters. <field> (including labels, domain, and period) must be no more than 253 characters.`
   - [x] Submit a hostname record with content with a label exceeding 63 characters : `1234567890123456789012345678901234567890123456789012345678901234.hello`. Expected error message: `Labels must be no more than 63 characters. <field> (including labels, domain, and period) must be no more than 253 characters.`

- [x] [Follow the process for requesting a design review](https://dhscisa.enterprise.slack.com/docs/T02QH7E1MHA/F06CZ6MRUKA). If code is not user-facing, delete design reviewer checklist
  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons
- [x] Checked for errors or warnings with an a11y browser tool (such as ANDI or WAVE)

- [x] Verified code meets all checks above. Address any checks that are not satisfied

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons
  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References


src/registrar/validations.py
```diff
@@ -80,56 +72,158 @@ def get_max_length_attrs(limit: int) -> dict[str, str]:
...
+    By default refers to validated field as "name".
+    """
+    # If content field not given, we are validating the name field
+    if not content_field:
+        return f"Enter the name {requirement}."
+    return f"Enter the {content_field} {requirement}."
+
+
+def get_fqdn_error_message(content_type=None) -> str:
+    """Returns fully qualified domain name error message by field that is validated."""
```
kimallen: Super nit: Maybe obvious and not necessary, but 
```suggestion
    """Returns fully qualified domain (fqdn) name error message by field that is validated."""
```

src/registrar/validations.py
```diff
@@ -80,56 +72,158 @@ def get_max_length_attrs(limit: int) -> dict[str, str]:
 MX_CONTENT_SPACES_ERROR_MESSAGE = "Enter the mail server without any spaces."
 TXT_RECORD_CONTENT_QUOTES_ERROR_MESSAGE = "Enter content using quotation marks at neither the beginning nor end."
 TXT_RECORD_CONTENT_MAX_LENGTH_ERROR_MESSAGE = "Content must be no more than 4080 characters."
+HOSTNAME_CONTENT_TRAILING_NUMBER_ERROR_MESSAGE = "Enter content that ends with a domain name."
+
+
+def get_content_type_by_record_type(record_type):
```
kimallen: I keep getting a little confused when I see `content_type` since it's not parallel to `record_type`. I know you have content_type throughout, but I can't really think of a better suggestion to differentiate it. 

 I wonder if here, since it's defining content/copy, if `content_type_label` might make more sense? Probably fine as is, but mentioning in case you have ideas/thoughts on naming.

src/registrar/validations.py
```diff
@@ -80,56 +72,158 @@ def get_max_length_attrs(limit: int) -> dict[str, str]:
...
+    # If content field not given, we are validating the name field
+    if not content_field:
+        return f"Enter the name {requirement}."
+    return f"Enter the {content_field} {requirement}."
+
+
+def get_fqdn_error_message(content_type=None) -> str:
+    """Returns fully qualified domain name error message by field that is validated."""
+    # If a field type is not given, default to referring to fqdn as name
+    if not content_type:
```
kimallen: Nice- I like how you can use the same function for both the name and content fields to get the message.

src/registrar/validations.py
```diff
@@ -80,56 +72,158 @@ def get_max_length_attrs(limit: int) -> dict[str, str]:
...
+
+
+def _validate_dns_hostname_content(content: str, field_type: str | None) -> None:
+    """
+    Validates a DNS record hostname content for CNAME, PTR, and MX records.
+    Handles fully qualified names (e.g., 'www.example.gov') but not relative names (e.g., 'www').
+
+    Normalizes to lowercase and validates:
+    - No spaces
+    - Valid characters only (letters, numbers, hyphens, periods, @ for apex)
```
kimallen: Is this correct? I thought other symbols were allowed characters (just not `@ ( ) : ;`)

src/registrar/validations.py
```diff
@@ -80,56 +72,158 @@ def get_max_length_attrs(limit: int) -> dict[str, str]:
...
+    - No consecutive dots
+    - No leading dots (but trailing dots accepted)
+    - No hyphens at start/end of labels // TODO: Check if this still applies
```
kimallen: Do we need to do some research for this? Did you want to resolve that for this PR or separately?

src/registrar/validations.py
```diff
@@ -208,12 +300,28 @@ def validate_txt_content(content: str) -> None:
         raise ValidationError(TXT_RECORD_CONTENT_MAX_LENGTH_ERROR_MESSAGE)
 
 
+def validate_cname_content(content: str) -> None:
+    """Validates a CNAME record's target value."""
+    from registrar.utility.enums import DNSRecordTypes
```
kimallen, line 290: I'm curious the reason to import inside of these functions instead of at the top?

src/registrar/validations.py
kimallen, line 1: I'm noticing that this file is spaced to cover the entire application.  I'm wondering if we should namespace somehow with DNS, either using an embedded dns/validations.py (or validations/dns.py?) or by naming each validation specifically related to dns records, or even creating a DnsRecordValidations class. It's otherwise not clear that these functions are specific to dns records.

That could involve a larger refactor outside the scope of this ticket, so we culd create a new ticket to deal with that if it feels like it shouldn't be dealt with here.

kimallen (changes requested):
Clean and thorough! Looking great.
Mostly I have some nits and a question about validations organization.

Also noticed (testing locally) two periods for the error message for a blank record for CNAME and PTR (but not MX)

Image

And this may be unrelated to your PR, but the Priority blank error shows up twice:

Image

src/registrar/tests/views/test_domain_dns_record_and_form.py
```diff
@@ -59,24 +59,31 @@ class TestDomainDNSRecordsView(TestWithDNSRecordPermissions, WebTest):
...
-        #     "id": "test-ptr",
-        #     "name": "www",
-        #     "type": "PTR",
-        #     "content": "www.example.com",
-        #     "ttl": 300,
-        #     "comment": "Mocked record created",
-        # },
+        {
+            "id": "test-cname",
+            "name": "test",  # CNAME record must use differnet name than A/AAAA records
```
kimallen: typo nit:
```suggestion
            "name": "test",  # CNAME record must use different name than A/AAAA records
```

erinysong:
> Also noticed (testing locally) two periods for the error message for a blank record for CNAME and PTR (but not MX)

Good catch on the extra periods! Turns out I left a period at the end of that required content requirement message constant and just removed it. Thank you for finding that!

I also noticed we display the priority error message twice for our MX records. I didn't have the chance to look at it more closely, but I believe it has to do with the fact we call that priority message at both the form and model level. I can make a ticket to address it

src/registrar/validations.py
erinysong: hm yeah I was also unsure what the right language would be to use this, since technically we're validating a fqdn hostname but we're just specifying what exactly that fqdn hostname is used for in relation to the record type. Open to suggestions as well and happy to change this to `content_type_label`

src/registrar/validations.py
erinysong: oh that is not! I copied that from our name validation method description - I'll go ahead and update that and the name description. Thank you for calling that out!

src/registrar/validations.py
erinysong: hm when I tested what I shared in this [slack post](https://cisa-corp.slack.com/archives/C074E2QJ7J9/p1778017137680319) I found hostname content allows hyphens at both the start and end. Just verified for all CNAME, MX, and PTR records that we can have content start and end with a `-` so I think it's safe to remove this description since we aren't validating for it either to match with CF

src/registrar/validations.py
erinysong: yeah! If I recall correctly, it's to prevent circular imports since when we initialize validations, our enums are not fully initialized yet with `DnsRecordTypes`. We started fixing circular import issues by only importing in places where they're necessary so I ended up following the pattern here, but open to any other ways to address circular imports if there are other alternatives!

src/registrar/validations.py
erinysong: that's a good idea, especially as our validations file is getting much larger from DNS hosting

erinysong:
> > Also noticed (testing locally) two periods for the error message for a blank record for CNAME and PTR (but not MX)
> 
> Good catch on the extra periods! Turns out I left a period at the end of that required content requirement message constant and just removed it. Thank you for finding that!
> 
> I also noticed we display the priority error message twice for our MX records. I didn't have the chance to look at it more closely, but I believe it has to do with the fact we call that priority message at both the form and model level. I can make a ticket to address it

There's some behavior on the MX record priority validation that I don't feel familiar enough yet to update without affecting other validations here ( `test_duplicate_mx_shows_only_duplicate_error_on_priority_not_required_error` method description says `the model's _validate_mx_priority fires on None and injects a spurious second error.`, so I filed ticket #5041  to look into what's happening and fix the double error message

src/registrar/validations.py
kimallen: Thanks for the explanation! No recs, just wasn't remembering why.

src/registrar/validations.py
kimallen: Yup, sounds reasonable to remove it and not validate for that. Would you mind also removing that row from the spreadsheet for the error alert message?

kimallen (approved):
Updates LGTM! Nice work!

witha-k (changes requested):
Everything looks good but I do have a question:

For specifically the error that shows up when the hostname record has content exceeding 253 characters, I noticed that usually the error will include the word "**Full**" like so (and this is true for both PTR and CNAME records):

> Labels must be no more than 63 characters. **Full** <field name> (including labels, domain, and period) must be no more than 253 characters.

Example: 
Image

However, for MX records, the word "**Full**" is ommitted before the field name:
> Labels must be no more than 63 characters. Mail server (including labels, domain, and period) must be no more than 253 characters.

Image

Is this intentional?

erinysong:
> Everything looks good but I do have a question:
> 
> For specifically the error that shows up when the hostname record has content exceeding 253 characters, I noticed that usually the error will include the word "**Full**" like so (and this is true for both PTR and CNAME records):
> 
> > Labels must be no more than 63 characters. **Full**  (including labels, domain, and period) must be no more than 253 characters.
> 
> Example: Image
> 
> However, for MX records, the word "**Full**" is ommitted before the field name:
> 
> > Labels must be no more than 63 characters. Mail server (including labels, domain, and period) must be no more than 253 characters.
> 
> Image
> 
> Is this intentional?

yes that's intentional! I referenced the messages in the [alert messages spreadsheet](https://docs.google.com/spreadsheets/d/1hF7_n_zo3EpfbgqqqPUMtSMZXynMtRMsBMlcIqZ91ns/edit?gid=987899751#gid=987899751) and matched the message for each hostname record type's label message:
- CNAME: `Labels must be no more than 63 characters. Full target (including labels, domain, and periods) must be no more than 253 characters.`
- MX: `Labels must be no more than 63 characters. Mail server (including labels, domain, and periods) must be no more than 253 characters.`
- PTR: `Labels must be no more than 63 characters. Full domain name (including labels, domain, and periods) must be no more than 253 characters.`

witha-k (approved):
LGTM!

Everything works as intended. Slight nitpicks to consider in the future:

**1. Error message entering for a hostname record with any of the characters: `@( ): ;`**

Right now the error message is:
> Enter the <field> without using parentheses, colons, or semicolons.

I know this is intended, but it really is confusing if you see this error when you put in an "@" since "@" is not a paranthesis, colon, or semicolon. The user won't know they did wrong. It's probably not mentioned because @ is allowed for the root name but I think some tweaking in the messaging here for specifically the record field would make it more clear. 

Image

**2. Already mentioned as something to revisit, but I really do support changing the error message for putting a period at the beginning of a hostname record since "consecutive periods" isn't the actual issue with that.**

PR #5052, opened 2026-05-29
#4632: Check for hosts before putting expired domains in `deleted` - [aa]

### LINKED ISSUE #4632: Check for hosts before putting expired domains in `deleted`
### Current Behavior

We recently ran into an issue with a domain that was deleted after expiration, but shouldn't have been because it had NS records (ie, was set up)

[Slack thread](https://cisa-corp.slack.com/archives/C06VDEWTGSH/p1769637691840009)

There are a few interrelated things that seem to have happened:
1. This domain (and 4 others) were marked as needing DNS, even though they had NS records and were actively being used (they should have been marked as Ready)
2. We only checked for the domain status in the action to delete the domain after expiration.

### Expected Behavior

1. If a domain has NS records that are active, it should be marked as Ready. 
2. A domain shouldn't be able to be moved to "Unknown" or "DNS Ready" with active NS records. 
3. We have some code that checks for this, but it's currently commented out. We need to figure out why and if this would have prevented the actions that moved these domains from Ready --> DNS Needed

4. As an additional protection, we should check for host records if a domain is in "DNS needed" before deleting that domain. We deleted these records, but their existence alongside the "DNS needed" status should cause an error or flag the domain rather than going forward with the destructive action of deleting the domain. We could trigger an email to the helpdesk when this happens, or to a Slack channel.

### Steps to Reproduce

Not sure if this can be replicated precisely because these domains were in a state that generally shouldn't be possible. Reproduction will require editing db records to get a domain in the correct state:

1. Add NS records for a domain
2. Domain should be marked as Ready
3. Edit the domain state in the DB to be "DNS Needed" instead of Ready, but leave the NS/host records in place

If the domain is expired, it can now be deleted via the [deletion script](https://github.com/cisagov/manage.get.gov/blob/main/src/registrar/management/commands/delete_expired_domains_not_setup.py).

### Environment

_No response_

### Additional Context

_No response_

### Issue Links

_No response_

abe-alam-ecs:
## Ticket

Resolves #4632

## Changes

- Added Github secrets to send new emails to our Slack channels
- Updated domain model (uncommented some code, added a guard)
- Updated deletion script and added test

## Context for reviewers

You will need a domain with nameservers that is expiring today. The state of the domain would normally be ready. 

Since we're testing a bug, we'll need to change the state to either DNS NEEDED or UNKNOWN for this PR's testing. Since this is not a natural transition, I will modify some in aa but if you are a developer, it may be best to test this in local, where you can easily play with your db data. The directions below are for my sandbox. Modify as needed for local testing. 

If you are testing locally, remember to add the email for [this channel](https://dhscisa.enterprise.slack.com/archives/C0B6FKUN4DA) into `Allowed Emails` in `/admin`.

## Setup

Designers/Product, please reach out to me for setup/testing. The changes will appear in the sandbox slack channel for notifications.  

Developers: I have already setup the domains `gogebiccountymi.gov`, `22007apply.gov`, and `fatherhood.gov` for testing on 6-8-2026. You can modify their states to unknown instead of dns_needed as double validation. 

1. Find a Ready domain that is expiring today using the OT&E registry. Ask me for the link if you need it. 
2. Change the state of this domain to DNS Needed / Unknown (I used the shell, but you can use Beekeeper with creds if you would prefer). The code in the shell looks something like this: 

```
from registrar.models import Domain

domain = Domain.objects.get(name="[domain.gov]")
print(domain.state)    # should be ready
Domain.objects.filter(id=domain.id).update(state=Domain.State.UNKNOWN)
Domain.objects.get(id=domain.id).state   # should be dns_needed  
``` 

Note: You may also need to validate the expiration_dates with `print(domain.expiration_date)`. I have noticed some of the sandboxes do not match OT&E. To update the expiration date, you can use...

```
from django.utils import timezone
Domain.objects.filter(id=domain.id).update(expiration_date=timezone.now().date())
```

3. Run the deletion script (`delete_expired_domains_not_setup`) with: `python manage.py delete_expired_domains_not_setup --alert-email [slack-email-from-context-above]` 

- [x] If any updated dependencies on Pipfile, also update dependencies in requirements.txt. - N/A

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons
  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References


src/registrar/models/domain.py
```diff
@@ -2672,6 +2665,14 @@ def _get_property(self, property):
             raise KeyError("Requested key %s was not found in registry cache." % str(property))
 
     def delete_with_no_dns(self, *args, **kwargs):
+        # Guard: stop deletion if domain has active nameservers
```
therealslimhsiehdy, line 2668: **Suggestion (nonblocking):**  Thoughts on maybe leaving another comment mentioning that nameservers is checking the registry and hosts are checking the DB for discrepancies? Not hard pressed either way!

src/registrar/models/domain.py
```diff
@@ -1937,13 +1930,13 @@ def ready(self):
         # domain was READY, then not READY, then is READY again.
         # We do not want to overwrite first_ready.
         if self.first_ready is None:
-            self.first_ready = timezone.now()
+            self.first_ready = date.today()
```
therealslimhsiehdy, line 1933: **Question:** Could you clarify on why this was changed from `timezone.now()` to `date.today()`?

src/registrar/models/domain.py
```diff
@@ -2655,8 +2648,8 @@ def _invalidate_cache(self):
         NOTE: The "on hold date" property is a one off addition - we want to
         make sure that when there is state change we delete the on hold date as well."""
         self._cache = {}
-        logging.info(f"Delete hold date on {self.name}")
-        delattr(self, "on_hold_date") if hasattr(self, "on_hold_date") else None
+        logging.info(f"Cache is empty for domain: {self.name}")
+        # delattr(self, "on_hold_date") if hasattr(self, "on_hold_date") else None
```
therealslimhsiehdy: **Suggestion (nonblocking):** I would add a comment mentioning why this was commented out

src/registrar/management/commands/delete_expired_domains_not_setup.py
```diff
@@ -79,15 +86,26 @@ def get_domains(self):
...
+                logger.error(f"Failed to delete {domain.name}: {action_not_allowed_msg}")
+                if alert_email:
+                    try:
+                        send_templated_email(
+                            template_name="emails/domain_deletion_failed_body.txt",
+                            subject_template_name="emails/domain_deletion_failed_subject.txt",
+                            to_addresses=[alert_email],
+                            context={"domain": domain.name},
+                        )
+                    except Exception as email_err:
```
therealslimhsiehdy: **Suggestion:** `except Exception as email_err:` -> `except EmailSendingError as email_err:`

src/registrar/management/commands/delete_expired_domains_not_setup.py
```diff
@@ -79,15 +86,26 @@ def get_domains(self):
...
             try:
-                domain.deleteInEpp()
-                domain.save()
+                domain.delete_with_no_dns()
                 deleted_domains.append(domain)
-            except Exception:
-                logger.error(f"Failed to delete {domain.name}")
+            except ActionNotAllowed as action_not_allowed_msg:
+                logger.error(f"Failed to delete {domain.name}: {action_not_allowed_msg}")
+                if alert_email:
```
therealslimhsiehdy, line 97: **Question:** Would we want another check for `alert_email` isn't set? IE
```
else:
logger.warning("alert_email not configured, no Slack notification for failed deletion of {domain.name}")
```

or would we never run into this situation?

src/registrar/templates/emails/domain_deletion_failed_body.txt
```diff
@@ -0,0 +1,5 @@
+{% autoescape off %}{# In a text file, we don't want to have HTML entities escaped #}
+
+Attempted to delete domain {{ domain }} and there were already hosts… did not delete.
```
therealslimhsiehdy, line 3: **Question:** Do we have a Google doc for the specific wording of this email that's requested? Or not sure if design/product have anything specific in mind

src/registrar/tests/common.py
```diff
@@ -1976,7 +1978,10 @@ def mockInfoDomainCommands(self, _request, cleaned):
             "nameserverwithip.gov": (self.infoDomainHasIP, None),
             "namerserversubdomain.gov": (self.infoDomainCheckHostIPCombo, None),
             "freeman.gov": (self.InfoDomainWithContacts, None),
-            "threenameserversdomain.gov": (self.infoDomainThreeHosts, None),
+            "threenameserversdomain.gov": (
+                self.infoDomainNoHost if self.mockedSendFunction.call_count >= 8 else self.infoDomainThreeHosts,
```
therealslimhsiehdy, line 1984: **Question:** Why is the magic number 8? Or if there's a reason it's 8, I'd suggest adding a comment on why so in case

src/registrar/tests/test_models_domain.py
```diff
@@ -2051,7 +2083,10 @@ def test_nameservers_are_idempotent(self):
                 call(commands.InfoHost(name="ns1.cats-are-superior3.com"), cleaned=True),
             ]
             self.mockedSendFunction.assert_has_calls(expectedCalls, any_order=True)
-            self.assertEqual(self.mockedSendFunction.call_count, 4)
+
+            # Four additional EPP calls (InfoDomain and InfoHost x3) due to
```
therealslimhsiehdy: **Question:** Sorry me again asking why the magic number is 8 / I don't get the comment bc I thought it added up to 7 :lolsob:

therealslimhsiehdy (commented):
Just some general comments/questions!

src/registrar/models/domain.py
abe-alam-ecs: Yes -- `first_ready` is a DateField (2026-06-09), and `timezone.now()` includes time information (the format is 2026-06-09 13:03:00). This was something Alysia and I caught; we wanted it to match the appropriate format. Probably worth auditing in the db as well just to make sure there isn't any bad data

src/registrar/templates/emails/domain_deletion_failed_body.txt
abe-alam-ecs: I had a meeting with Kaitlin and she verbally gave me these lines! No formal documentation

src/registrar/management/commands/delete_expired_domains_not_setup.py
abe-alam-ecs: Confused on this. Is this best practice?

I intentionally used Exception here because I didn't want some unknown Exception to be skipped and not logged at all.

therealslimhsiehdy: `send_templated_email` already raises `EmailSendingError` on failure, so catching that specifically means unexpected errors still bubble up and fail "loudly" rather than getting quietly logged as an email failure

The pattern is also used quite a bit in the codebase, ie `send_domain_notifications_emails`, `email_current_metadata_report`, etc

Just flagging the tradeoff!

src/registrar/models/domain.py
therealslimhsiehdy: So thoughts on this...

I think this line was there bc `self._cache = {}` doesn't fully invalidate the cache specifically the field of `on_hold_date` (which is on the instance itself?) so we might be getting a stale value so that's why this is fully clearing it out?

therealslimhsiehdy: I think that's why there's the note of 
```
 NOTE: The "on hold date" property is a one off addition - we want to
        make sure that when there is state change we delete the on hold date as well
```
above

src/registrar/tests/common.py
abe-alam-ecs: There's not a great reason for this; it's to pass a test, `test_user_removes_too_many_nameservers`. 

I basically forced the mock to read as 0 hosts in the course of that test by setting this conditional. Here's the call sequence:
1) InfoDomain
2) InfoHost (ns1)
3) InfoHost (ns2)
4) InfoHost (ns3)
5) DeleteHost (ns2)
6) UpdateDomain (Delete other hosts ns1, ns3)
7) DeleteHost (ns3)
8) InfoDomain, is_dns_needed() now evaluates and the mock returns infoDomainNoHost. is_dns_needed --> true, dns_needed() transition fires. domain.is_active() returns False. 

I'll see if I can just make this test better to avoid this hacky solution

therealslimhsiehdy: OMG thank you for listing this out, sorry about that -- feel free to leave as is (maybe add a comment with your best judgement?), I don't want it to balloon the ticket for you 😭

abe-alam-ecs: Yeah it looks like there isn't a clean fix unless I do something like a side_effect that introduces even uglier code. I'm writing up the ticket to clean up the MockEppLib and I'll include cleaning up these conditionals as a part of that work.

src/registrar/tests/test_models_domain.py
abe-alam-ecs: Forgot to hit submit on this one yesterday!

`_invalidate_cache()` clears `self._cache` in the nameserver setter before evaluating transition conditions. When `is_dns_needed` reads `self.nameservers`, the `_get_property` finds the cache empty and calls `_fetch_cache(fetch_hosts=True)`. That triggers `_get_or_create_domain_in_registry` (1 InfoDomain call) and `_fetch_hosts` (1 InfoHost call per nameserver, 3 total.) -- 4 extra calls beyond what is in `expected_calls` above. 

Let me know if this is still unclear...

src/registrar/models/domain.py
abe-alam-ecs: Corrected -- added this line back!

src/registrar/management/commands/delete_expired_domains_not_setup.py
abe-alam-ecs: Added below!

therealslimhsiehdy (approved):
LGTM! Thank you for your patience on all my questions and feedback! 🙏🏼 

Image

src/registrar/models/domain.py
```diff
@@ -2672,6 +2665,15 @@ def _get_property(self, property):
             raise KeyError("Requested key %s was not found in registry cache." % str(property))
 
     def delete_with_no_dns(self, *args, **kwargs):
+        # Guard: stop deletion if domain has active nameservers
+        # self.nameservers checks the registry, self.host.all() checks our db
+        if len(self.nameservers) >= 2 or self.host.all().count() >= 2:
+            logger.error(
+                f"Domain {self.name} has {len(self.nameservers)} nameservers "
```
allly-b, line 2672: *Suggestion:* Since this is an error that could trigger if there are 0 nameservers and 2+ hosts, can you add the host count to this print out too?

src/registrar/management/commands/delete_expired_domains_not_setup.py
```diff
@@ -79,15 +86,30 @@ def get_domains(self):
...
+                            context={"domain": domain.name},
+                        )
+                    except EmailSendingError as email_err:
+                        logger.error(f"Failed to send deletion alert email for {domain.name}: {email_err}")
+                else:
+                    logger.warning(
+                        f"alert_email not configured, no Slack notification for failed deletion of {domain.name}"
+                    )
+            except Exception as e:
+                logger.error(f"Failed to delete {domain.name}: {e}")
```
allly-b:
```suggestion
                logger.error(f"Failed to delete {domain.name}: {e}", exc_info=True)
```
*Strong suggestion/Change before merging:* we really want to make sure we log the traceback in this scenario

src/registrar/management/commands/delete_expired_domains_not_setup.py
```diff
@@ -79,15 +86,30 @@ def get_domains(self):
...
+                if alert_email:
+                    try:
+                        send_templated_email(
+                            template_name="emails/domain_deletion_failed_body.txt",
+                            subject_template_name="emails/domain_deletion_failed_subject.txt",
+                            to_addresses=[alert_email],
+                            context={"domain": domain.name},
+                        )
+                    except EmailSendingError as email_err:
+                        logger.error(f"Failed to send deletion alert email for {domain.name}: {email_err}")
```
allly-b:
```suggestion
                        logger.error(f"Failed to send deletion alert email for {domain.name}: {email_err}", exc_info=True)
```
*suggestion*: same as below we want to make sure we don't lose the traceback if it's failing.

src/registrar/models/domain.py
```diff
@@ -2655,7 +2648,7 @@ def _invalidate_cache(self):
         NOTE: The "on hold date" property is a one off addition - we want to
         make sure that when there is state change we delete the on hold date as well."""
         self._cache = {}
```
allly-b, line 2650: **Question/non blocking** why does the cache reset here?

src/registrar/models/domain.py
```diff
@@ -2655,7 +2648,7 @@ def _invalidate_cache(self):
...
-        logging.info(f"Delete hold date on {self.name}")
+        logging.info(f"Cache is empty for domain: {self.name}")
         delattr(self, "on_hold_date") if hasattr(self, "on_hold_date") else None
```
allly-b, line 2652: **Question/non-blocking:** did you ever figure out why we are deleting this attribute?

src/registrar/tests/test_models_domain.py
```diff
@@ -2051,7 +2084,10 @@ def test_nameservers_are_idempotent(self):
                 call(commands.InfoHost(name="ns1.cats-are-superior3.com"), cleaned=True),
             ]
             self.mockedSendFunction.assert_has_calls(expectedCalls, any_order=True)
-            self.assertEqual(self.mockedSendFunction.call_count, 4)
+
+            # Four extra EPP calls from _invalidate_cache() in the nameserver setter.
```
allly-b, line 2088: *Kudos:* Thanks so much for this comment! Love a clear comment.

allly-b (approved):
See some minor suggestions. Overall, all of this looks great!

src/registrar/templates/emails/domain_deletion_failed_body.txt
katypies: @therealslimhsiehdy / @abe-alam-ecs  - this email is internal-only at this point, so we didn't have a Google doc for the content. We can make one, though, if that would be helpful for any future updates!

src/registrar/models/domain.py
abe-alam-ecs: My understanding is that _cache is cleared here to fetch fresh data from the registry via `_get_property` --> `_fetch_cache`. The nameserver setter just made changes to the registry, so the cached data is stale. The call is made around line 855 in domain.py.

TL;DR: It's so that the next read gets current data

src/registrar/management/commands/delete_expired_domains_not_setup.py
abe-alam-ecs: Done!

src/registrar/management/commands/delete_expired_domains_not_setup.py
abe-alam-ecs: Done!

src/registrar/models/domain.py
abe-alam-ecs: Yes! Rebecca pointed it out above, earlier in this PR... I'll copy from there. 

Screenshot 2026-06-12 at 3 22 47 PM

src/registrar/models/domain.py
abe-alam-ecs: Great idea, I've changed this.

src/registrar/models/domain.py
abe-alam-ecs: done!

src/registrar/models/domain.py
allly-b: Thanks!!

src/registrar/models/domain.py
allly-b: Ahh makes sense

abe-alam-ecs:
There is a follow-on ticket surrounding this work [linked here](https://github.com/cisagov/manage.get.gov/issues/5100).

Additionally, see [this slack thread](https://cisa-corp.slack.com/archives/C06K96473FA/p1781278789592649?thread_ts=1780936908.343819&cid=C06K96473FA) for Kaitlin's approval. 

Merging now.

PR #5076, opened 2026-06-04
#5060: Add Federal Tribe model + migration

### LINKED ISSUE #5060: Create FederalTribe data model with database migration
## Sub-issue description

A new data model named `FederalTribe` needs to be created to store federally recognized tribal data from the source CSV (https://github.com/cisagov/flat-tribal-leaders/blob/main/tribal-leaders.csv). This model will serve as the foundation for importing and storing tribal information in the database.

## Acceptance criteria

- [ ] A new data model named FederalTribe is created for federally recognized tribal data
- [ ] See [Slack thread](https://cisa-corp.slack.com/archives/C05BGB4L5NF/p1780431437936359) for what columns should be mapped to a field in the model
- [ ] Address-related data is stored using patterns consistent with domain_information.py
- [ ] A migration is created to add the FederalTribe table to the database

## Additional context

- Reference the CSV source file structure: https://github.com/cisagov/flat-tribal-leaders/blob/main/tribal-leaders.csv
- Use `domain_information.py` as a reference for address field storage patterns
- This work is a prerequisite for the admin interface and import script sub-issues

therealslimhsiehdy:
## Ticket

Resolves #5060

## Changes

* Created a FederalTribe model
* Also has migrations along with it

## Context for reviewers

We want to map [some of the columns](https://cisa-corp.slack.com/archives/C05BGB4L5NF/p1780431437936359) from [this CSV ](https://github.com/cisagov/flat-tribal-leaders/blob/main/tribal-leaders.csv) to it's own model

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons
  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References


src/registrar/models/federal_tribe.py
```diff
@@ -0,0 +1,123 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class FederalTribe(TimeStampedModel):
+
+    StateTerritoryChoices = DomainRequest.StateTerritoryChoices
+
+    tribe_full_name = models.CharField(
+        null=True,
```
allly-b: **non-blocking question** (this can always be updated later so not blocking just curious) should tribe full name really be nullable? I would think this needs to have something in it. Like wise, should this be a unique field?

src/registrar/models/federal_tribe.py
```diff
@@ -0,0 +1,123 @@
...
+    website = models.CharField(
+        null=True,
+        blank=True,
+    )
+    date_elected = models.CharField(
+        null=True,
+        blank=True,
+        verbose_name="Date elected",
+    )
+    next_election = models.CharField(
```
allly-b: **question:** should this and the one above be a date field instead?

src/registrar/models/federal_tribe.py
```diff
@@ -0,0 +1,123 @@
...
+
+from .utility.time_stamped_model import TimeStampedModel
+from .domain_request import DomainRequest
+
+logger = logging.getLogger(__name__)
+
+
+class FederalTribe(TimeStampedModel):
+
+    StateTerritoryChoices = DomainRequest.StateTerritoryChoices
```
allly-b, line 14: **optional suggestion/non-blocking**: since domain request, portfolio, suborganization, domain info and this one need stateTerritoryChoices, it may be worth moving the state/territory choices to shared location that all of these files can import from. That said, this could be something that's punted to a follow on ticket.

src/registrar/models/federal_tribe.py
```diff
@@ -0,0 +1,123 @@
...
+        max_length=10,
+        null=True,
+        blank=True,
+        verbose_name="Zip code",
+    )
+    urbanization = models.CharField(
+        null=True,
+        blank=True,
+    )
+    phone = models.CharField(
```
allly-b: **blocking suggestion:** phone should be a PhoneNumberField as validation is then handled by django and we don't need to reinvent the wheel there (though to be fair, charfield isn't wrong!)
```suggestion
    phone = models.PhoneNumberField(
```

src/registrar/models/federal_tribe.py
```diff
@@ -0,0 +1,123 @@
...
+        null=True,
+        blank=True,
+    )
+    email = models.CharField(
+        null=True,
+        blank=True,
+    )
+    website = models.CharField(
```
allly-b, line 98: **comment/minor suggestion:** As a best practice in django all charfields are technically supposed to have a max_length. That said, I don't think we have been consistent with this. I would suggest adding max lengths, but in the interest of time, at the very least add a max_length to this one of 255 (max_length = 255) to match what we do for the [website model](https://github.com/cisagov/manage.get.gov/blob/bef8f7ab7a676ee34a053679913ae9fe0bee1634/src/registrar/models/website.py#L6).

src/registrar/models/federal_tribe.py
therealslimhsiehdy: Great catch! I double checked through the CSV and there are no duplicate names (just in case) - will make it unique.

src/registrar/models/federal_tribe.py
therealslimhsiehdy: (Punting this to a follow on clean up ticket for later)

src/registrar/models/federal_tribe.py
therealslimhsiehdy: Added it! I also looked at `contacts.py` and added a max_length of 320 to email just in case 😆

src/registrar/models/federal_tribe.py
```diff
@@ -0,0 +1,125 @@
...
+        null=True,
+        blank=True,
+        verbose_name="last name",
+    )
+    suffix = models.CharField(
+        null=True,
+        blank=True,
+        help_text="Name suffix (ie jr, sr, III)",
+    )
+    aka = models.CharField(
```
allly-b, line 45:
```suggestion
    #  preferred name or nickname of contact
    aka = models.CharField(
```

**optional suggestion:** this field wasn't really clear as to what it would be for until I opened the csv. I would suggest aka become "preferred name" or "nickname" to be more in line with other documents. However, feel free to punt this as we can always rename it. I would suggest keeping a comment here though just for clarity sake.

src/registrar/models/federal_tribe.py
```diff
@@ -0,0 +1,125 @@
...
+        null=True,
+        blank=True,
+        help_text="Shortened tribe name",
+    )
+    tribe_alternate_name = models.CharField(
+        null=True,
+        blank=True,
+        verbose_name="tribe alternate name",
+    )
+    first_name = models.CharField(
```
allly-b, line 31:
```suggestion
    #  Tribal contact info
    first_name = models.CharField(
```

Suggest adding a comment as there is an abrupt change in the type of fields to now being representative of a tribe's primary contact. 

Quick call out that this repeats a very similar structure to the "contacts" model and given how many fields this has, the table would be made smaller by just having a single contact field.  Not a requirement though!

src/registrar/models/__init__.py
```diff
@@ -83,6 +85,7 @@
 auditlog.register(DomainInvitation)
 auditlog.register(DomainInformation)
 auditlog.register(FederalAgency)
+auditlog.register(FederalTribe)
```
allly-b, line 88: **suggestion:** Love that you added this to audit log, you should go ahead and register this in [admin.py ](https://github.com/cisagov/manage.get.gov/blob/0cbd4e7d0207f8102556bfa17fbbe042345797e8/src/registrar/admin.py#L5491) so that it appears in django admin. 


**suggestion:** I would also recommend creating the user group migration right after you add this table to admin so that superusers at least can see this table.  For an example of the create groups migrations see [here](https://github.com/cisagov/manage.get.gov/blob/4d44120160801bf20487ae2944573d2671eaa332/src/registrar/migrations/0145_create_groups_v19.py). Without this you will see the table locally but not on a sandbox unless you manually update permissions for a user.


Also a reminder, when this is ready for analysts to see the table you will also need to update the [create_cisa_analyst_group ](https://github.com/cisagov/manage.get.gov/blob/0cbd4e7d0207f8102556bfa17fbbe042345797e8/src/registrar/models/user_group.py#L17) function to add this table to it. Then re-run the user group migration stuff so analysts can see it.

allly-b (commented):
I had a few suggestions, specifically regarding just going ahead and registering this with django admin and updating permissions so superusers can see it at least. This way you don't have to close this PR only to make another one with a migration when you try to test the code. Let me know what you think, they aren't blocking so I can go ahead and approve if you disagree with doing those right now.

src/registrar/models/federal_tribe.py
therealslimhsiehdy: Will add the comment!

Also open to discussion! I do agree the structure is similar however I'm on the side of keeping it separate since right now this is being used for imports from a CSV which would make it cleaner and wouldn't have to touch FK's with Contact

src/registrar/models/__init__.py
therealslimhsiehdy: Totally open to discussion to questions below:

1. Thoughts on having that migration be done in the /admin ticket that follows this one - I don't mind either way, just thought it'd be cleaner with the /admin done all at once
2. If we were to do the user group migration, wouldn't we do these permissions below all at once? Or what would be the reason for updating the `create_cisa_analyst_group` later?

```
    UserGroup.create_cisa_analyst_group(apps, schema_editor)
    UserGroup.create_omb_analyst_group(apps, schema_editor)
    UserGroup.create_full_access_group(apps, schema_editor)
```

src/registrar/models/federal_tribe.py
allly-b: Yeah I am good with the keeping it all in one table, makes sense

src/registrar/models/__init__.py
allly-b: Only reason for not including the analyst group initially is just so that they don't see an extra table pop up before you're done with all the tickets. Not an imperative though. And yes if you want to include the adding to /admin in the follow on ticket, no objection.

allly-b (approved):
Approving now given the comment that the addition to /admin etc will be coming in the follow on work.

src/registrar/models/federal_tribe.py
```diff
@@ -0,0 +1,127 @@
...
+    )
+    urbanization = models.CharField(
+        null=True,
+        blank=True,
+    )
+    phone = PhoneNumberField(
+        null=True,
+        blank=True,
+    )
+    email = models.CharField(
```
DaisyGuti: Use EmailField here instead of CharField so we get email-format validation, and to match the rest of the codebase

src/registrar/models/federal_tribe.py
```diff
@@ -0,0 +1,127 @@
...
+        max_length=320,
+        null=True,
+        blank=True,
+    )
+    website = models.CharField(
+        max_length=255,
```
DaisyGuti, line 99: public_contact uses EmailField(max_length=320) please make it consistent here too

src/registrar/migrations/0184_federaltribe.py
```diff
@@ -0,0 +1,134 @@
...
+                    "tribe_full_name",
+                    models.CharField(
+                        help_text="Full official name of the federally recognized tribe",
+                        unique=True,
+                        verbose_name="Tribe full name",
+                    ),
+                ),
+                ("tribe", models.CharField(blank=True, help_text="Shortened tribe name", null=True)),
+                ("tribe_alternate_name", models.CharField(blank=True, null=True, verbose_name="tribe alternate name")),
+                ("first_name", models.CharField(blank=True, null=True, verbose_name="first name")),
```
DaisyGuti: FYSA django auto translates '_' to spaces. so verbose name here is 'verbose' 😆

src/registrar/migrations/0184_federaltribe.py
```diff
@@ -0,0 +1,134 @@
...
+                ("last_name", models.CharField(blank=True, null=True, verbose_name="last name")),
+                ("suffix", models.CharField(blank=True, help_text="Name suffix (ie jr, sr, III)", null=True)),
+                ("aka", models.CharField(blank=True, null=True, verbose_name="AKA")),
+                ("job_title", models.CharField(blank=True, null=True)),
+                ("organization", models.CharField(blank=True, null=True)),
+                (
+                    "physical_address_line1",
+                    models.CharField(blank=True, null=True, verbose_name="Physical Address Line 1"),
```
DaisyGuti: For clean up, on the verbose name usage normally we want them all to follow the same casing

src/registrar/migrations/0184_federaltribe.py
```diff
@@ -0,0 +1,134 @@
...
+                ("updated_at", models.DateTimeField(auto_now=True)),
+                (
+                    "tribe_full_name",
+                    models.CharField(
+                        help_text="Full official name of the federally recognized tribe",
+                        unique=True,
+                        verbose_name="Tribe full name",
+                    ),
+                ),
+                ("tribe", models.CharField(blank=True, help_text="Shortened tribe name", null=True)),
```
DaisyGuti, line 28: Just curious why since the other is tribe_full_name this isn't "tribe_short_name"

src/registrar/migrations/0184_federaltribe.py
```diff
@@ -0,0 +1,134 @@
...
+                ("organization", models.CharField(blank=True, null=True)),
+                (
+                    "physical_address_line1",
+                    models.CharField(blank=True, null=True, verbose_name="Physical Address Line 1"),
+                ),
+                (
+                    "physical_address_line2",
+                    models.CharField(blank=True, null=True, verbose_name="Physical Address Line 2"),
+                ),
+                ("city", models.CharField(blank=True, null=True)),
```
DaisyGuti, line 38: missing the physical prefix on the rest of the physical address.

the csv has mailing I belive I saw it. 

So, either we add the mailing, with it's prefix correctly set on both address types OR we remove it from the physical and note somewhere we decided to leave out the mailing?

So one address, no prefix, or both with prefix - thoughts on this? ([noted on Slack as well)](https://cisa-corp.slack.com/archives/C05BGB4L5NF/p1780431437936359)

src/registrar/migrations/0184_federaltribe.py
```diff
@@ -0,0 +1,134 @@
...
+                ("suffix", models.CharField(blank=True, help_text="Name suffix (ie jr, sr, III)", null=True)),
+                ("aka", models.CharField(blank=True, null=True, verbose_name="AKA")),
+                ("job_title", models.CharField(blank=True, null=True)),
+                ("organization", models.CharField(blank=True, null=True)),
+                (
+                    "physical_address_line1",
+                    models.CharField(blank=True, null=True, verbose_name="Physical Address Line 1"),
+                ),
+                (
+                    "physical_address_line2",
```
DaisyGuti: doesn't exist in the csv, are we adding this for later pop in the forms?

DaisyGuti (changes requested):
Nice job! Just a couple of cleanups and some non-blocking questions.

src/registrar/migrations/0184_federaltribe.py
therealslimhsiehdy: It's how it's denoted in [the CSV we are referencing's column header](https://github.com/cisagov/flat-tribal-leaders/blob/main/tribal-leaders.csv). I added a help text for this field for reference.

src/registrar/models/federal_tribe.py
```diff
@@ -0,0 +1,127 @@
...
+    physical_address_line2 = models.CharField(
+        null=True,
+        blank=True,
+        verbose_name="Physical Address Line 2",
+    )
+    city = models.CharField(
+        null=True,
+        blank=True,
+    )
+    state_territory = models.CharField(
```
DaisyGuti, line 72: The csv has the full names, do we have an importer? just curious

src/registrar/migrations/0184_federaltribe.py
therealslimhsiehdy: Will remove the verbose name from the model field.

src/registrar/migrations/0184_federaltribe.py
therealslimhsiehdy: This does not exist in the CSV, but one of the AC's is to match the address to what we have in `domain_information.py` which is why this is also split out into two.

src/registrar/migrations/0184_federaltribe.py
```diff
@@ -0,0 +1,134 @@
...
+                        verbose_name="State, Territory, or Military Post",
+                    ),
+                ),
+                ("zipcode", models.CharField(blank=True, max_length=10, null=True, verbose_name="Zip code")),
+                ("urbanization", models.CharField(blank=True, null=True)),
+                (
+                    "phone",
+                    phonenumber_field.modelfields.PhoneNumberField(blank=True, max_length=128, null=True, region=None),
+                ),
+                ("email", models.CharField(blank=True, max_length=320, null=True)),
```
DaisyGuti: Same not here on the email

src/registrar/migrations/0184_federaltribe.py
therealslimhsiehdy: We are only mapping the physical address column (not the mailing address) -- I am a bit confused what you mean regarding the prefix and mailing - could you elaborate on what you mean by this comment?

src/registrar/migrations/0184_federaltribe.py
therealslimhsiehdy: Could you elaborate on this -- what is "same here" and not on the email?

src/registrar/models/federal_tribe.py
therealslimhsiehdy: This will be handled in the script and can be mapped both ways so not a concern here!

src/registrar/models/federal_tribe.py
therealslimhsiehdy: Not sure if you were looking at a different spot but email field is already marked 320?

src/registrar/migrations/0184_federaltribe.py
therealslimhsiehdy: I see the previous comments now - this has now been updated to EmailField!

src/registrar/migrations/0184_federaltribe.py
DaisyGuti: Sure. It seems we are only using the physical address from the csv. If that is true then having the physical in front of it doesn't really matter. It would only be used if we had another type of address.

I was suggesting we either remove the prefix if we will not have other types of addresses. Or if we are keeping the prefix (because we will have other types) then we use the prefix throughout the full address fields. 

physical_city etc.

DaisyGuti (changes requested):
Please either add physical_ prefix or remove to all the address info (city, state, etc.)

The rest looks good to go.

DaisyGuti (approved):
LGTM, great job!

PR #5082, opened 2026-06-08
#4760: Delete DNS record in DnsHostService

### LINKED ISSUE #4760: DnsHostService can delete record on Cloudflare and database
### Sub-issue description

DnsHostService should support, given a Cloudflare DNS record id, deleting that DNS record on Cloudflare (through CloudflareService) and on the app database.

### Acceptance criteria

- [ ] Add method in DnsHostService to delete DNS record given a Cloudflare record id. The method must:
   - [ ] Delete record in Cloudflare through CloudflareService
   - [ ] Delete record from the db
   - [ ] Cascade delete objects related to the deleted DNS record (eg VendorDnsRecord, DnsRecordVendorDnsRecord) from the db 
- [ ] Write relevant tests

### Additional context

_No response_

erinysong:
## Ticket

Resolves #4760

## Changes

- Add methods in DnsHostService to delete a DNS record in vendor service and database given the record's record id and its zone id

## Context for reviewers

Originally we considered using `transaction.on_commit` when deleting a DNS record - we would first try deleting the record in our database then run another transaction to delete the record in the vendor service when we successfully delete the record and associated objects in the database (compared to our create and update methods which first create/update the record in the vendor service, then create our objects).

We realized that transactions can be rolled back from the top down (outer block failure prevents the inner block), but not from bottom up (inner block failure does not roll back the other block), so we considered following the same pattern as our create/update (update in vendor service, then delete objects). If we run into an issue with deletion, it will most likely happen at the vendor level and the db record should be an accurate reflection of the record that exists on the Internet through the external vendor service. If the record fails to be deleted on the vendor service, we should inform the user that the registrar failed to delete the record because we failed to delete it in the vendor service, as we do our other record operations. 

Our final implementation performs deleting the db record in the database, then deleting the record in the vendor service **in the same transaction block.** This way, if we fail to delete the record in the vendor, we can roll back the entire transaction and restore the record in the database.

## Setup

Run tests to verify that when we call `delete_dns_record` in DnsHostService, the associated DNSRecord, VendorDnsRecord, and DnsRecordVendorDnsRecord objects get deleted from the database. 

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons
  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References


src/registrar/models/dns/dns_record.py
```diff
@@ -448,3 +448,20 @@ def update_from_vendor_data(cls, x_zone_id: str, x_record_id: str, vendor_record
...
+
+    @classmethod
+    def delete_record_from_x_record_id(cls, x_zone_id: str, x_record_id: str):
+        """Delete an existing DnsRecord and its associated DnsVendorRecord and DnsRecordVendorDnsRecord."""
+        try:
+            with transaction.atomic():
+                vendor_dns_record = VendorDnsRecord.objects.get(x_record_id=x_record_id)
+                vendor_dns_zone = VendorDnsZone.objects.get(x_zone_id=x_zone_id)
+                dns_zone = DnsZone.objects.get(vendor_dns_zone=vendor_dns_zone)
+                dns_record = cls.objects.get(vendor_dns_record=vendor_dns_record, dns_zone=dns_zone)
```
kimallen: We actually have a `get_by_x_record_id` method on DnsRecord, so you could bypass using the zone:
```
dns_record = cls.get_by_x_record_id(x_record_id)
```

src/registrar/models/dns/dns_record.py
```diff
@@ -448,3 +448,20 @@ def update_from_vendor_data(cls, x_zone_id: str, x_record_id: str, vendor_record
...
+
+                # DnsRecordVendorDnsRecord object are deleted on cascade
```
kimallen: nit:
```suggestion
                # DnsRecordVendorDnsRecord object is deleted on cascade
```

src/registrar/services/dns_host_service.py
```diff
@@ -263,6 +263,44 @@ def update_and_save_dns_record(self, x_zone_id, x_record_id, form_record_data) -
...
+            raise ValueError("Could not find the DNS record in registrar db to delete.")
+
+        x_record_id = dns_record.get_active_x_record_id()
+        if not x_record_id:
+            raise ValueError("This DNS record is missing an external record id and cannot be deleted.")
+
+        self.delete_and_save_dns_record(x_zone_id, x_record_id, record_name)
+        return dns_record
+
+    def delete_and_save_dns_record(self, x_zone_id: str, x_record_id: str, record_name: str) -> str:
```
kimallen: I find the name a little confusing since you're deleting it from the db and not saving it. What about something like `delete_and_remove_dns_record`?

src/registrar/services/dns_host_service.py
```diff
@@ -263,6 +263,44 @@ def update_and_save_dns_record(self, x_zone_id, x_record_id, form_record_data) -
             raise
         return vendor_record_data
 
+    def delete_dns_record(self, x_zone_id: str, record_id: int) -> DnsRecord:
```
kimallen: I'm confused about this method calling the `delete_and_save_dns_record` method. What does this method do that the other does not? It looks like the difference is that you are retrieving the `x_record_id` and the `record_name` for the logs. Could you do those two things in `delete_and_save_dns_record` since they are just lookups and not doing any kind of operation? I might be missing something.

src/registrar/tests/services/test_dns_host_service.py
```diff
@@ -956,3 +956,48 @@ def test_create_and_save_zone_assigns_custom_nameservers(self):
...
+        x_record_id = self.vendor_record_data["result"].get("id")
+        _, _, zone = create_initial_dns_setup(
+            x_zone_id=x_zone_id,
+            nameservers=self.vendor_zone_data["result"].get("name_servers"),
+        )
+        DnsRecord.create_from_vendor_data(x_zone_id, self.vendor_record_data)
+        record_db_id = DnsRecord.objects.get(dns_zone=zone, name=self.vendor_record_data["result"].get("name")).id
+        vendor_record_db_id = VendorDnsRecord.objects.get(x_record_id=x_record_id).id
+
+        DnsRecord.delete_record_from_x_record_id(x_zone_id, x_record_id)
```
kimallen: If this is the method you are testing, this test should go under `tests/models/dns/test_dns_record.py`. Here you would want to test the DnsHostService methods, like `delete_and_save_dns_record`. Including tests for both (model test and service test) would be ideal.

src/registrar/tests/services/test_dns_host_service.py
```diff
@@ -956,3 +956,48 @@ def test_create_and_save_zone_assigns_custom_nameservers(self):
...
+            x_zone_id=x_zone_id,
+            nameservers=self.vendor_zone_data["result"].get("name_servers"),
+        )
+        DnsRecord.create_from_vendor_data(x_zone_id, self.vendor_record_data)
+        record_db_id = DnsRecord.objects.get(dns_zone=zone, name=self.vendor_record_data["result"].get("name")).id
+        vendor_record_db_id = VendorDnsRecord.objects.get(x_record_id=x_record_id).id
+
+        with patch("registrar.models.DnsRecord.delete", side_effect=IntegrityError("simulated failure")):
+            with self.assertRaises(IntegrityError):
+                DnsRecord.delete_record_from_x_record_id(x_zone_id, x_record_id)
```
kimallen: Same comment as the success test

kimallen (changes requested):
Looks like this will work!
A few change requests around the code organization and some refactor suggestions.

I'm also interested in discussing your thought process more regarding the order of deletion (vendor then db)

src/registrar/models/dns/dns_record.py
```diff
@@ -448,3 +448,20 @@ def update_from_vendor_data(cls, x_zone_id: str, x_record_id: str, vendor_record
...
+                dns_record.delete()
```
kimallen: We should wrap these deletions in a transaction so that if the second fails, the other gets rolled back.

erinysong: hm I may be misunderstanding what `transaction.atomic` does, but is this different from putting all these db operations in an atomic block so that when any of these db operations fail, all db operations in the atomic block roll back?

src/registrar/services/dns_host_service.py
erinysong: I'm good with merging these two functions into one! I was following the pattern where we have an create/update record method to check for the record in the db and call another helper method which performed the vendor db operations. I don't have strong feelings about keeping them as separate methods or combining them. Let me know what you think!

src/registrar/services/dns_host_service.py
erinysong: Good idea! Updating the name for clarity, thank you!

src/registrar/tests/services/test_dns_host_service.py
erinysong: that was completely an oops on my part, that tested function is supposed to be the DnsHostService's delete method. My bad for deleting it from the DnsRecord class method and will update that - thank you for bringing that up!

src/registrar/models/dns/dns_record.py
kimallen: Oh sorry! I totally missed that it _was_ in a transaction already. Disregard!

src/registrar/services/dns_host_service.py
kimallen: My preference is to combine them since they aren't actually completing different operations.

src/registrar/services/dns_host_service.py
```diff
@@ -263,6 +263,44 @@ def update_and_save_dns_record(self, x_zone_id, x_record_id, form_record_data) -
...
+    def delete_and_remove_dns_record(self, x_zone_id: str, x_record_id: str, record_name: str) -> str:
+        """Delete DNS record in vendor service and persist the changes in the local database."""
+        try:
+            self.dns_vendor_service.delete_dns_record(x_zone_id, x_record_id)
+            logger.info(f"Successfully deleted record {record_name} in vendor service.")
+        except (APIError, HTTPStatusError) as e:
+            logger.error(f"Failed to delete record {record_name}: {str(e)}")
+            raise APIError(str(e)) from e
+        except Exception as e:
+            logger.error(f"Failed to delete record {record_name} in database: {str(e)}.")
```
DaisyGuti: Please update the message to vendor / cf not a db issue here right?

src/registrar/services/dns_host_service.py
```diff
@@ -263,6 +263,44 @@ def update_and_save_dns_record(self, x_zone_id, x_record_id, form_record_data) -
...
+            record_name = dns_record.name
+        except DnsRecord.DoesNotExist:
+            raise ValueError("Could not find the DNS record in registrar db to delete.")
+
+        x_record_id = dns_record.get_active_x_record_id()
+        if not x_record_id:
+            raise ValueError("This DNS record is missing an external record id and cannot be deleted.")
+
+        self.delete_and_remove_dns_record(x_zone_id, x_record_id, record_name)
+        return dns_record
```
DaisyGuti: Could you update the docstring or return?

src/registrar/models/dns/dns_record.py
```diff
@@ -448,3 +448,18 @@ def update_from_vendor_data(cls, x_zone_id: str, x_record_id: str, vendor_record
         except Exception:
             logger.exception("Failed to update and save record to database.")
             raise
+
+    @classmethod
+    def delete_record_from_x_record_id(cls, x_zone_id: str, x_record_id: str):
```
DaisyGuti: Please remove x_zone_id, as it is not used.

DaisyGuti: Could you rename to follow the pattern of  delete_by....

src/registrar/models/dns/dns_record.py
```diff
@@ -448,3 +448,18 @@ def update_from_vendor_data(cls, x_zone_id: str, x_record_id: str, vendor_record
         except Exception:
             logger.exception("Failed to update and save record to database.")
             raise
+
+    @classmethod
```
DaisyGuti, line 452: Please add service tests.

src/registrar/models/dns/dns_record.py
```diff
@@ -448,3 +448,18 @@ def update_from_vendor_data(cls, x_zone_id: str, x_record_id: str, vendor_record
...
+    def delete_record_from_x_record_id(cls, x_zone_id: str, x_record_id: str):
+        """Delete an existing DnsRecord and its associated DnsVendorRecord and DnsRecordVendorDnsRecord."""
```
DaisyGuti: typo? VendorDnsRecord

DaisyGuti (changes requested):
Kim covered a good part. Just a couple of minor nits. Great job on this!

src/registrar/services/dns_host_service.py
erinysong: I agree with that! I'll merge the two into a single method

src/registrar/models/dns/dns_record.py
erinysong: @kimallen hm revisiting this and should we expect that multiple records in different zones would ever share the same x_record_id? Or are x_record_id's also unique regardless of if they are in different zones?

erinysong: resolved in 6/15  DNS pod check-in: let's ask Cloudflare to what degree record id's are kept unique (on an overall level vs. a zone/account level) and adjust based on their response

src/registrar/models/dns/dns_record.py
```diff
@@ -448,3 +448,18 @@ def update_from_vendor_data(cls, x_zone_id: str, x_record_id: str, vendor_record
...
+    @classmethod
+    def delete_by_x_record_id(cls, x_record_id: str):
+        """Delete an existing DnsRecord and its associated VendorRecord and DnsRecordVendorDnsRecord."""
+        try:
+            with transaction.atomic():
+                vendor_dns_record = VendorDnsRecord.objects.get(x_record_id=x_record_id)
+                dns_record = cls.get_by_x_record_id(x_record_id)
+
+                # DnsRecordVendorDnsRecord object is deleted on cascade
+                dns_record.delete() # type: ignore
```
erinysong: Ignores mypy error `error: Item "None" of "DnsRecord | None" has no attribute "delete"` given that when we get None from getting the DnsRecord, we expect a rollback on the transaction

kimallen (approved):
LGTM! Thanks for the updates!

For future 👀 will you updated the description to include what we learned with rolling back the db if CF call fails (it currently reflects pre-discussion)?

src/registrar/tests/services/test_dns_host_service.py
```diff
@@ -956,3 +956,71 @@ def test_create_and_save_zone_assigns_custom_nameservers(self):
...
+        )
+        DnsRecord.create_from_vendor_data(x_zone_id, self.vendor_record_data)
+        record_db_id = DnsRecord.objects.get(dns_zone=zone, name=self.vendor_record_data["result"].get("name")).id
+        vendor_record_db_id = VendorDnsRecord.objects.get(x_record_id=x_record_id).id
+
+        with patch("registrar.models.DnsRecord.delete", side_effect=IntegrityError("simulated failure")):
+            with self.assertRaises(Exception):
+                self.service.delete_dns_record(x_zone_id, record_db_id)
+
+        # DnsRecord, VendorDnsRecord, and DnsRecordVendorDnsRecord deleted
```
DaisyGuti: maybe you mean 'not deleted'?

src/registrar/tests/services/test_dns_host_service.py
```diff
@@ -956,3 +956,71 @@ def test_create_and_save_zone_assigns_custom_nameservers(self):
         self.assertNotEqual(initial_zone_data, updated_zone_data)
         self.assertEqual(updated_zone_data["vanity_name_servers"], dns_zone.nameservers)
         self.assertEqual(response["result"], updated_zone_data)
+
+    def test_delete_db_record_success(self):
+        """delete_record_from_x_record_id successfully deletes DnsRecord, VendorDnsRecord,
```
DaisyGuti: pls doc to update to delete_dns_record

src/registrar/tests/services/test_dns_host_service.py
```diff
@@ -956,3 +956,71 @@ def test_create_and_save_zone_assigns_custom_nameservers(self):
...
+        self.service.delete_dns_record(x_zone_id, record_db_id)
+
+        # DnsRecord, VendorDnsRecord, and DnsRecordVendorDnsRecord deleted
+        self.assertFalse(VendorDnsRecord.objects.filter(x_record_id=x_record_id).exists())
+        self.assertFalse(DnsRecord.objects.filter(id=record_db_id).exists())
+        self.assertFalse(
+            RecordsJoin.objects.filter(vendor_dns_record_id=vendor_record_db_id, dns_record_id=record_db_id).exists()
+        )
+
+    def test_delete_db_record_with_db_error_fails(self):
```
DaisyGuti, line 982: You could assert that the `self.dns_vendor_service.delete_dns_record(x_zone_id, x_record_id)` was never called on a db error

DaisyGuti (changes requested):
Nice job! Minor clean up leftovers and one assertion suggestion for the db failure test.

DaisyGuti (approved):
LGTM!

Issue #4925, opened 2026-04-22
[P2] Standardize DNS hosting error responses


DaisyGuti:
### Issue Description
DNS failures surface inconsistently. Some show a generic message, some show nothing (zone-not-found returns JSON). The goal of this work is to standardize this.

Design in progress ref: [dns-error-handling.md](https://github.com/cisagov/manage.get.gov/blob/main/docs/developer/dns-error-handling.md). Old Prototype: #4932.

### Acceptance Criteria
- [x] Every DNS view that can fail catches `DnsHostingError` in a single `except` block and reads code and message off it.
- [x] When form fails due to a DnsHostingError, we render messages as a banner alert (not inline)
- [x] **DID DIFFERENTLY- See comment below** The zone-not-found path raises a typed `DnsNotFoundError` instead of returning the JsonResponse it does now.
- [x]  No leftover JSON error responses on the DNS form path.
- [x] The design doc (docs/developer/dns-error-handling.md) reflects the alert message approach; JSON-envelope references removed.
- [x] EnrollmentNotAllowedError matches the other DNS errors logs shape.
- [x] Remove except (APIError, RequestError) --- as these should be captured in DNSHostingError
- [ ] Lint and tests pass - with a test for the request_id-on-5xx behavior.
- [ ] Ensure /admin enrollment errors include the request_id, and appropriate messages. Consult with the content team as needed.

### Additional Context
Approach for consideration: The original design used a JSON envelope as the standard error shape. This is not how our views read data so we're changing the design in favor of same-page messages only, the view catches `DnsHostingError` and renders the message through Django messages, which fits our current path / HTMX form error display path.
```
except DnsHostingError as exc:
    messages.error(request, exc.message) 
```
- The 4xx vs 5xx decision needs a new home now that the envelope and its status map might be dropped as unnecessary. So if we are removing  [utility/api_responses.py#L39](https://github.com/cisagov/manage.get.gov/blob/e8e15f43c48e52910bba11902ec6dfd3af34d161/src/registrar/utility/api_responses.py#L39) we will need a way to tell apart the types of errors, for example an `is_operational` property on `DnsHostingError` in errors.py.
- `request_id` already flows through the request via `RequestLoggingMiddleware` and the logging context (#4924). Read it from there for the inline reference.
- HTMX banners render via Django messages + the messagesRefresh trigger. We need the TemplateResponse + trigger for HTMX.
[...]

Issue #4922, opened 2026-04-22
[P2] Simplify DnsHostService to pass exceptions through


DaisyGuti:
### Issue Description
Several DnsHostService methods catch Cloudflare exceptions and re-raise them as `APIError`. Now that CloudflareService already raises typed errors, this wrapping just loses the type and logs the failure a second time. Remove it and let the typed exceptions pass straight through.

This also retires the temporary `APIError = DnsHostingError` alias from #4921.

### Acceptance Criteria
- [ ] A Cloudflare failure surfaces from DnsHostService as the original typed `DnsHostingError` (e.g. a 404 as DnsNotFoundError).
- [ ] A single DNS failure produces one log line.
- [ ] Lint and tests pass.

### Additional Context
Full failure flow: [developer guide](https://github.com/cisagov/manage.get.gov/blob/main/docs/developer/dns-error-handling.md#what-developers-do).

### Links to other issues
- ⛔️ Is blocked by #4921

DaisyGuti:
Relates to the https://github.com/cisagov/manage.get.gov/pull/5104 comment by Erin.

PR #5128, opened 2026-06-22
#4921 Raise typed errors in  CloudflareService methods

### LINKED ISSUE #4921: [P2] Migrate CloudflareService to raise typed DnsHostingError subclasses
### Issue Description
CloudflareService catches generic httpx exceptions and re-raises them as one generic `APIError`, so the view sees a single opaque type for every failure. Raise typed `DnsHostingError` subclasses instead, so the view can map each failure (404, 429, timeout) to a specific message and HTTP status. The [developer guide](https://github.com/cisagov/manage.get.gov/blob/main/docs/developer/dns-error-handling.md#what-developers-do) has the status-to-subclass mapping and the failure flow.

Keep `APIError` as a temporary alias to `DnsHostingError` for #4922 to retire.

### Acceptance Criteria
- [x] A Cloudflare 4xx/5xx surfaces from CloudflareService as the typed `DnsHostingError` subclass that matches the wire-code reference.
- [x] The raised error carries upstream_status and cf-ray when Cloudflare returned a response.
- [ ] The view and admin handle `DnsHostingError` and read its wire code and upstream status.
- [x] MockCloudflareService's error triggers raise the matching typed subclasses.
- [ ] Lint and tests pass.

### Additional Context
Prototype: the 404 case (DnsNotFoundError) in [cloudflare_service.py#L168-L177](https://github.com/cisagov/manage.get.gov/blob/e8e15f43c48e52910bba11902ec6dfd3af34d161/src/registrar/services/cloudflare_service.py#L168-L177) (PR #4932); this ticket covers the remaining failures.

### Links to other issues
- ⛔️ Is blocked by #4920

kimallen:
## Ticket

Resolves #4921

## Changes

- For each of the CloudflareService methods, they each except either a `RequestError `or a `HttpStatusError` and raise a DnsTransportError or a subtype of DnsHostingError based on the error's status code.
- Removes unused `CloudflareService.get_page_accounts` and it's uses (it was replaced awhile back with `get_account_by_name`)
- Updates tests and adds a couple new ones that were missing
- Adjust DnsRecordsView and admin view for domain to expect `DnsHostingError`

## Context for reviewers

The ticket says to `Keep APIError as a temporary alias to DnsHostingError for https://github.com/cisagov/manage.get.gov/issues/4922 to retire.`. However, there were not any failing affects when removing the APIError from the CloudflareService methods, so I did remove it from CF. It still is raised in the view.

See some questions around messaging in inline comments. In particular, It seems that we should be returning relatively generic error messages. The way it is currently set up, any 404 returned from a API returns a ZONE_NOT_FOUND, even if a zone is not relevant. It may be that I'm missing something and there's a way we should customize error messages for each error code- lmk if you see an existing path to do so. I did not change it, but anticipating that we will in a future ticket.

The second AC says to surface the `cf_ray` and `upstream_status` in the view. I wasn't sure how I should do that without also doing ticket https://github.com/cisagov/manage.get.gov/issues/4925, so I just logged those items in the view to show that they are available.

in TestCloudflareService, I created a `_get_failure_cases` method so that we can exclude cases with statuses that are irrelevant for a particluar api call.

When testing editing of a record using a mock error (locally), the result is the error message as a banner, but the form closes and when reopening it (before a refresh), it shows the Add record form with no data. Fixing that is outside the scope of this PR, so leaving as is for now.

## Setup

The test corresponding to /admin domain view is `registrar.tests.test_admin_domain.TestDomainAdminWebTest.test_enroll_dns_hosting_button_shows_error_message_on_failure`

Manually test using records that start with `error-400` or `error-403`. Check logs.


src/registrar/tests/services/test_cloudflare_service.py
```diff
@@ -52,25 +134,73 @@ def setUp(self):
         }
         self.service = CloudflareService(client=mock_client)
 
+    def _get_failure_cases(self, cases_to_exclude_by_status_code: list[int] = None):
```
kimallen, line 142: Included this to exclude cases with statuses that would not be expected for particular api calls (ex. GET calls wouldn't expect validation 400 or 409s)

src/registrar/services/cloudflare_service.py
```diff
@@ -4,10 +4,42 @@
...
+    DnsUpstreamError,
+)
 
 logger = logging.getLogger(__name__)
 
+_STATUS_TO_ERROR = {
+    400: (DnsValidationError, DnsHostingErrorCodes.VALIDATION_FAILED),
+    401: (DnsAuthError, DnsHostingErrorCodes.AUTH_FAILED),
+    403: (DnsAuthError, DnsHostingErrorCodes.AUTH_FAILED),
+    404: (DnsNotFoundError, DnsHostingErrorCodes.ZONE_NOT_FOUND),
```
kimallen: 404 is matched very specifically with a zone error. I'm expecting we'd want something more generic, so then it can apply to a "not found" record as well. Would that make sense @DaisyGuti or am I missing something?

src/registrar/views/domain.py
```diff
@@ -1091,6 +1093,10 @@ def post(self, request, *args, **kwargs):
             else:
                 is_first_record, record_id = self._handle_create(request, x_zone_id, form_record_data)
 
+        except DnsHostingError as e:
+            # temp log to show these values are available. Remove in #4892
+            logger.error(f"💕 wire_code: {e.wire_code}, upstream_status: {e.upstream_status}")
```
kimallen: "surfacing" values here with a log to meet AC. I expect it will be properly surfaced in #4925

src/registrar/services/cloudflare_service.py
DaisyGuti: You're not missing anything. It would probably be best to add a 404 generic, DnsNotFoundError. 
in errors.py:
```
NOT_FOUND = x   # generic: a xyz record wasn't found
DnsHostingErrorCodes.NOT_FOUND: (
    "We couldn't find that DNS record. It may have already been removed."
)
```
wire it, then we could use it as 404: (DnsNotFoundError, DnsHostingErrorCodes.NOT_FOUND)

In error cases for zones we can use the 
`DnsHostingErrorCodes.ZONE_NOT_FOUND, upstream_status=404, context={"zone_id": zone_id}`

src/registrar/services/cloudflare_service.py
```diff
@@ -109,13 +142,12 @@ def create_cf_zone(self, zone_name: str, x_account_id: str):
             resp.raise_for_status()
             logger.info(f"Created zone {zone_name}")
         except RequestError as e:
-            logger.error(f"Failed to create zone {zone_name} for account {x_account_id}: {e}")
-            raise
+            raise DnsTransportError(
+                code=DnsHostingErrorCodes.UPSTREAM_TIMEOUT,
+                context={"zone_name": zone_name, "account_id": x_account_id, "exc_class": type(e).__name__},
```
DaisyGuti: Should this be x_account_id on both sides?

src/registrar/services/cloudflare_service.py
```diff
@@ -4,10 +4,42 @@
...
+    logger.exception(
+        "Dns provider returned %s for DNS request",
+        status,
+        extra={"upstream_status": status, "error_code": code.name, **ctx},
+    )
+    return exc_cls(code=code, upstream_status=status, context=ctx)
+
 
 @dataclass(frozen=True)
 class CloudflareDnsSettingsUpdateResponse:
```
DaisyGuti, line 81: This is looking great. One big picture suggestion while we're setting the pattern at the service level. This is the service that makes the httpx calls. Right now each method has its own try/except and we are repeating.

What if we collapse all of this into one small [context manager ](https://stackoverflow.com/questions/60367476/context-manager-that-handles-exceptions) (python [source](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) for more info) on so there's exactly one place for all. 

```
@contextmanager
def _dns_call(self, **context):
    """Context manager to catch HTTPX-related errors and convert them into 
          appropriate typed DnsHostingErrors."""
    try:
        yield
    except HTTPStatusError as e:
        raise _typed_dns_error(e, **context) from e # note: `from e` to keep context
    except RequestError as e:
        raise DnsTransportError(
            code=DnsHostingErrorCodes.UPSTREAM_TIMEOUT,
            context={**context, "exc_class": type(e).__name__},
        ) from e
```

Then every method call would look like for example:

```
....
    url = f"/zones/{zone_id}/dns_records/{record_id}"
    with self._dns_call(resource="record", zone_id=zone_id, record_id=record_id):
        resp = self.client.get(url)
        resp.raise_for_status()
    return resp.json()
```

src/registrar/services/cloudflare_service.py
```diff
@@ -4,10 +4,42 @@
...
+    401: (DnsAuthError, DnsHostingErrorCodes.AUTH_FAILED),
+    403: (DnsAuthError, DnsHostingErrorCodes.AUTH_FAILED),
+    404: (DnsNotFoundError, DnsHostingErrorCodes.ZONE_NOT_FOUND),
+    409: (DnsValidationError, DnsHostingErrorCodes.RECORD_CONFLICT),
+    429: (DnsRateLimitError, DnsHostingErrorCodes.RATE_LIMIT_EXCEEDED),
+    500: (DnsUpstreamError, DnsHostingErrorCodes.UPSTREAM_ERROR),
+}
+
+
+def _typed_dns_error(e: HTTPStatusError, **context) -> DnsHostingError:
```
DaisyGuti: In ref. to R24. I think we need to add cover there for the unknowns / unmapped:

```
def _typed_dns_error(e: HTTPStatusError, **context) -> DnsHostingError:
    status = e.response.status_code
    ctx = {"cf_ray": e.response.headers.get("cf-ray"), **context} 
    exc_cls, code = _STATUS_TO_ERROR.get(status, (None, None))

    if exc_cls is None and 500 <= status <= 599:
        exc_cls, code = DnsUpstreamError, DnsHostingErrorCodes.UPSTREAM_ERROR
    if exc_cls is None:
        exc_cls, code = DnsHostingError, DnsHostingErrorCodes.UNKNOWN # 400s we don't list out 

    # debugging detail for the logs
    # unknown / unexpected: keep the traceback
    unexpected = code in (DnsHostingErrorCodes.UNKNOWN,)
    logger.error(
        "DNS provider /  or CF returned %s for the DNS request",
        status,
        extra={"upstream_status": status, "error_code": code.name, "response_body": e.response.text, **ctx},
        exc_info=unexpected,    # True only for the weird ones only hopefully
    )

    return exc_cls(code=code, upstream_status=status, context=ctx)
```

src/registrar/views/domain.py
DaisyGuti: There is an emoji here that likely needs to be cleaned up? :)

src/registrar/admin.py
```diff
@@ -5605,11 +5606,18 @@ def do_enroll_dns_hosting(self, request, obj):
         except EnrollmentNotAllowedError as e:
             logger.warning("DNS enrollment blocked: %s", e)
             self.message_user(request, str(e), messages.WARNING)
+        except DnsHostingError as e:
+            logger.exception(e)
+            self.message_user(
+                request,
+                f"Failed to enroll domain in DNS hosting. {str(e)}",
```
DaisyGuti: Now that we have our messages centralized we should probably move this but likely best in scope for the other ticket handling the view.

src/registrar/services/cloudflare_service.py
```diff
@@ -4,10 +4,42 @@
...
+    DnsNotFoundError,
+    DnsRateLimitError,
+    DnsTransportError,
+    DnsUpstreamError,
+)
 
 logger = logging.getLogger(__name__)
 
+_STATUS_TO_ERROR = {
+    400: (DnsValidationError, DnsHostingErrorCodes.VALIDATION_FAILED),
```
DaisyGuti, line 23: Doc says 4xx should be warning if you want to make them all error would you update the doc too pls?

DaisyGuti (changes requested):
Looks really good! One overall design cleanup suggestion that hit me while I was reviewing the cf service and one nit  / or decision pt to make on the log level.

src/registrar/services/cloudflare_service.py
kimallen: Good question that I forgot to ask. I noticed that sometimes in our logs we were labeling as `x_account_id` and sometimes as `account_id`. We should be consistent with one or the other. I'd be inclined to use `x_account_id` across the board so that it doesn't get confused with our db `account_id`s. Open to feedback.

src/registrar/services/cloudflare_service.py
kimallen: Great suggestion! I'll work on that.

src/registrar/tests/services/test_cloudflare_service.py
```diff
@@ -52,25 +134,73 @@ def setUp(self):
...
+                    http_error = HTTPStatusError(request="something", response=mock_response, message="other thing")
+
         if error["exception"] == RequestError:
-            http_error = RequestError(request="something", message=error["message"])
-        http_error.response = mock_response
-        mock_response.raise_for_status.side_effect = http_error
-        return mock_response
+            http_error = RequestError(request="something", message="last thing")
+        mock_api_response.raise_for_status.side_effect = http_error
+        return mock_api_response
```
asaki222, line 247: I like the helper function you added. It makes the tests a lot easier to read, and consistent.

src/registrar/services/cloudflare_service.py
asaki222: I think using `x_account_id` would be easier for us to understand when digging through logs.

src/registrar/services/cloudflare_service.py
```diff
@@ -138,17 +170,18 @@ def update_zone_dns_settings(
             resp = self.client.patch(appended_url, json=data)
             resp.raise_for_status()
             logger.info(
-                "Updated zone DNS settings for zone_id=%s (zone_mode=%s, nameservers.type=%s, namservers.ns_set=%s)",
+                "Updated zone DNS settings for zone_id=%s (nameservers.type=%s, namservers.ns_set=%s)",
                 zone_id,
```
asaki222, line 198: Question, do the %s evaluate to a value? Or is it intentional for it to be printed this way in our logs?

asaki222:
This _looks like_ this pr could resolve this issue #4922

src/registrar/tests/services/test_cloudflare_service.py
```diff
@@ -1,32 +1,114 @@
...
+        },
+        {
+            "test_name": "RequestError",
+            "error": {
+                "exception": RequestError,
+                "message": "There was an error getting a response",
+                "raised_error": DnsTransportError,
+                "code": DnsHostingErrorCodes.UPSTREAM_TIMEOUT,
+            },
         },
```
asaki222, line 97: Random 💡 , do you think we could add these errors to the mock cloudflare service?

src/registrar/services/cloudflare_service.py
kimallen: Yes- this is a form of string interpolation. Each `%s` corresponds with the value of the variables listed after the string, in order. 

I didn't write the original code here, so am not sure why this format was chosen. One difference between this and using an f-string (ex. `"My zone: {zone_id}"`) is that using the modulo operator allows for lazy interpolation whereas the f-string builds at runtime and could be a security consideration.

src/registrar/services/cloudflare_service.py
kimallen: I'm fine going with the guidance in the doc, but unclear on where I would specify a warning. Are you referring to the log level? This _STATUS_TO_ERROR mapping was added from the doc. Could you say more about how you would specify these as warnings?

src/registrar/services/cloudflare_service.py
kimallen: That makes sense generally. What does `code in (DnsHostingErrorCodes.UNKNOWN,)` mean? Is that the equivalent to `code == DnsHostingErrorCodes.UNKNOWN`?

src/registrar/services/cloudflare_service.py
kimallen: Updated to include `x` throughout

src/registrar/tests/services/test_cloudflare_service.py
kimallen: We could, though I'm not sure if there's a lot of value in it. If we find we want to simulate certain ones, we could add them.

kimallen: I guess the value could be for manually testing to see how the UI responds.

src/registrar/tests/services/test_cloudflare_service.py
```diff
@@ -1,32 +1,125 @@
...
+            "error": {
+                "exception": HTTPStatusError,
+                "raised_error": DnsRateLimitError,
+                "code": DnsHostingErrorCodes.RATE_LIMIT_EXCEEDED,
+            },
+            "cf_ray": "R2D2",
+        },
+        {
+            "test_name": "UnmappedError",
+            "status_code": 418,
```
kimallen, line 88: @katypies - this ones for you 😉

src/registrar/services/cloudflare_service.py
DaisyGuti: it's just the log level and wording mostly. But it will be effected if we don't log at the right level we won't see warnings. Not sure what level prod is at just something to be aware of.

src/registrar/services/cloudflare_service.py
DaisyGuti: It would cover any cases we don't specifically check for, are not expecting.

src/registrar/utility/errors.py
```diff
@@ -322,17 +322,20 @@ class APIError(Exception):
 class DnsHostingErrorCodes(IntEnum):
     """Error codes for DNS-hosting failures."""
 
-    ZONE_NOT_FOUND = 1
+    NOT_FOUND = 1
```
DaisyGuti, line 323: Thanks

src/registrar/services/cloudflare_service.py
```diff
@@ -1,13 +1,55 @@
...
+    DnsUpstreamError,
+)
 
 logger = logging.getLogger(__name__)
 
+_STATUS_TO_ERROR = {
+    400: (DnsValidationError, DnsHostingErrorCodes.VALIDATION_FAILED),
+    401: (DnsAuthError, DnsHostingErrorCodes.AUTH_FAILED),
+    403: (DnsAuthError, DnsHostingErrorCodes.AUTH_FAILED),
+    404: (DnsNotFoundError, DnsHostingErrorCodes.ZONE_NOT_FOUND),
```
DaisyGuti: I would flip this 404 to a generic NOT_FOUND code instead of ZONE_NOT_FOUND. The reasoning is bc for get_dns_record / update_dns_record / delete_dns_record, labeling it ZONE_NOT_FOUND would mean that "every 404 says zone-not-found". A generic code with context from its path (operation, zone_id, record_id) would help us reuse messages / product copy. That's my thought on this and how I would approach it for reuse, but I realize how this team works around copy and messaging so this may not be the best fit. **Non-blocking comment** -  just what I would normally do in this situation.

src/registrar/utility/errors.py
```diff
@@ -322,17 +322,20 @@ class APIError(Exception):
...
     RECORD_CONFLICT = 2
     VALIDATION_FAILED = 3
     RATE_LIMIT_EXCEEDED = 4
     AUTH_FAILED = 5
     UPSTREAM_TIMEOUT = 6
     UPSTREAM_ERROR = 7
     UNKNOWN = 8
+    SERVER_ERROR = 9
+    ZONE_NOT_FOUND = 10
```
DaisyGuti: As we are keeping the zone one here, then we should also add a RECORD_NOT_FOUND for get/update/delete_dns_record one if we aren't re-using generic not found copy.

src/registrar/utility/errors.py
```diff
@@ -322,17 +322,20 @@ class APIError(Exception):
...
 
 
 _DNS_WIRE_CODES = {
```
DaisyGuti, line 332: So we assure all are wired, let's add a check here e.g. `assert set(DnsHostingErrorCodes) <= set(_DNS_WIRE_CODES) `

src/registrar/utility/errors.py
```diff
@@ -322,17 +322,20 @@ class APIError(Exception):
...
-    ZONE_NOT_FOUND = 1
+    NOT_FOUND = 1
     RECORD_CONFLICT = 2
     VALIDATION_FAILED = 3
     RATE_LIMIT_EXCEEDED = 4
     AUTH_FAILED = 5
     UPSTREAM_TIMEOUT = 6
     UPSTREAM_ERROR = 7
     UNKNOWN = 8
+    SERVER_ERROR = 9
```
DaisyGuti: Let's remove SERVER_ERROR from this list. Anything that isn't a DnsHostingError (a DatabaseError, a programming bug, anything unexpected) propagates up to Django's default 500 handling).

We would likely be using UPSTREAM_ERROR for a DnsHostingError in the 500 (that is DNS related).

src/registrar/admin.py
```diff
@@ -5605,11 +5606,18 @@ def do_enroll_dns_hosting(self, request, obj):
...
+                messages.ERROR,
+            )
         except Exception as e:
             logger.exception(e)
             self.message_user(
                 request,
-                "Failed to enroll domain in DNS hosting.",
+                f"Failed to enroll domain in DNS hosting. {str(e)}",
```
DaisyGuti: This is printing out exception info to the user (UI). I would remove this / undo this. Maybe rephrase (unsure what has been decided on copy) but I believe there is a generic system error type of message Katherine pointed to in that excel sheet for general system failures.

src/registrar/admin.py
```diff
@@ -5605,11 +5606,18 @@ def do_enroll_dns_hosting(self, request, obj):
         except EnrollmentNotAllowedError as e:
             logger.warning("DNS enrollment blocked: %s", e)
             self.message_user(request, str(e), messages.WARNING)
+        except DnsHostingError as e:
+            logger.exception(e)
```
DaisyGuti: Suggestion:  `logger.error("DNS enrollment failed: %s", e)`

src/registrar/services/cloudflare_service.py
```diff
@@ -44,19 +86,27 @@ def __init__(self, client):
         client.headers = self.headers
         self.client = client
 
+    @contextmanager
```
DaisyGuti, line 115: Looks and works great!!! 🥇

DaisyGuti (changes requested):
Nice work. A couple of small changes requested before merge:

- [x] Remove SERVER_ERROR as our side errors (DatabaseError, bugs) deliberately propagate to Django's 500 per the dev guide, not into this DNS Error hierarchy.
- [x] Add a check to assert so codes must be wired up (helps us prevent dead code).
- [x] Clean up the admin generic Exception leaking raw str(e) to the user.

Looks really great! Excited for this foundational portion to merge!

src/registrar/admin.py
kimallen: This one only goes to /admin users, so I was thinking more info would be better so analysts could relay info. Though I guess if we are adding in a request_id, they don't need to know more info. I can remove the exception info, as I believe we will be replacing this copy with the ones in the spreadsheet.

src/registrar/services/cloudflare_service.py
kimallen: I had actually meant to change it to the generic one, which I see I only partially updated (by adding it to the ErrorCodes). The error copy has been approved and it's very generic, so I don't think there's any need to specify a ZONE_NOT_FOUND error anymore, so I went ahead and removed it throughout.

src/registrar/utility/errors.py
kimallen: Not keeping it- I removed these

src/registrar/services/cloudflare_service.py
kimallen: Done

src/registrar/utility/errors.py
kimallen: I added a test for this- pls check if that's covers your request

src/registrar/admin.py
kimallen: Why do you choose `.error` over `.exception` here?

src/registrar/assets/src/js/getgov/form-errors.js
```diff
@@ -5,7 +5,7 @@ export function initFormErrorHandling() {
     if (firstErrorField) {
       // Scroll to the first field in error
       firstErrorField.scrollIntoView({ behavior: 'smooth', block: 'center' });
-  
+
```
DaisyGuti, line 8: Ppl normally ask me to remove these types of edits so just following the teams norms and asking as well :)

docs/developer/dns-error-handling.md
DaisyGuti, line 1: Nice ty!

src/registrar/admin.py
```diff
@@ -5607,17 +5607,17 @@ def do_enroll_dns_hosting(self, request, obj):
             logger.warning("DNS enrollment blocked: %s", e)
             self.message_user(request, str(e), messages.WARNING)
         except DnsHostingError as e:
-            logger.exception(e)
+            logger.error("DNS enrollment failed: %s", e)
```
DaisyGuti: We only `self.message_user(request, "Failed to enroll domain in DNS hosting.", messages.ERROR)` as the service layer already logs errors.

docs/developer/dns-error-handling.md
```diff
@@ -153,7 +153,7 @@ Services log and raise. Views catch and render. A DNS failure produces exactly o
 ```json
 {
   "status": "error",
-  "code": "DNS_ZONE_NOT_FOUND",
+  "code": "DNS_NOT_FOUND",
   "message": "We couldn't find the DNS zone for this domain.",
```
DaisyGuti: Needs generalization

src/registrar/services/cloudflare_service.py
DaisyGuti: We only keep Cloudflare's status code, not its explanation. The explanation lives in the response body, which only this service sees. #4999 needs it to show users 'vendors's reason,' we should capture it. I think I had this on the original branch / version of the prototype, the stuff we need is in the response. 

Image

But, here we example, we could add a helper for reuse:
```
def _cf_error_detail(response: Response) -> dict:
    """Pull Cloudflare's own error code + message out of the response body, if present."""
    try:
        errors = response.json().get("errors") or []
    except ValueError:          # body wasn't JSON (e.g. an empty 500) -> nothing to pull
        return {}
    if not errors:
        return {}
    first = errors[0]
    return {"cf_error_code": first.get("code"), "cf_error_message": first.get("message")}
...
def _typed_dns_error(e: HTTPStatusError, **context) -> DnsHostingError:
    status = e.response.status_code
    ctx = {
        "cf_ray": e.response.headers.get("cf-ray"),
        **_cf_error_detail(e.response),   # <- adds cf_error_code + cf_error_message
        **context,
    }
    ...
    return exc_cls(code=code, upstream_status=status, context=ctx)
```

That would then give us something like:

```
exc.context == {
    "cf_ray": "ZXY321",
    "cf_error_code": 9005,
    "cf_error_message": "Bad request for dns record.",
    "x_zone_id": "54321",
    "record_data": {...},
}
# and for testing
# a 400 fixture whose body has errors[0] = {code: 9005, message: "Bad request for dns record."}
self.assertEqual(exc.context["cf_error_code"], 9005)
self.assertEqual(exc.context["cf_error_message"], "Bad request for dns record.")
```
And then we can relate to the #4999 msg. I think this is all that is left to full close the gap and then the dns host service should be a quick bridge only.

src/registrar/admin.py
```diff
@@ -5607,17 +5607,17 @@ def do_enroll_dns_hosting(self, request, obj):
...
             self.message_user(
                 request,
-                f"Failed to enroll domain in DNS hosting. {str(e)}",
+                "Failed to enroll domain in DNS hosting.",
                 messages.ERROR,
             )
         except Exception as e:
-            logger.exception(e)
+            logger.error("DNS enrollment failed: %s", e)
```
DaisyGuti: this one would be an exception: `logger.exception(e)`

src/registrar/services/cloudflare_service.py
DaisyGuti: Update needed to _typed_dns_error  to handle this too

```
def _typed_dns_error(e: HTTPError, **context) -> DnsHostingError:
    """Map any httpx failure (status error OR transport error) to a typed DnsHostingError, logging once."""
    if isinstance(e, HTTPStatusError):          # a response came back -> classify by status
        status = e.response.status_code
        ctx = {"cf_ray": e.response.headers.get("cf-ray"), **context}
        log_only = {"response_body": e.response.text}   # logged, NOT carried on the exception
        exc_cls, code = _STATUS_TO_ERROR.get(status, (None, None))
        if exc_cls is None and 500 <= status <= 599:
            exc_cls, code = DnsUpstreamError, DnsHostingErrorCodes.UPSTREAM_ERROR
        if exc_cls is None:                     # any 4xx we didn't map
            exc_cls, code = DnsHostingError, DnsHostingErrorCodes.UNKNOWN
    else:                                       # RequestError -> no response, transport failure
        status = None
        ctx = {"exc_class": type(e).__name__, **context}
        log_only = {}
        exc_cls, code = DnsTransportError, DnsHostingErrorCodes.UPSTREAM_TIMEOUT

    exc = exc_cls(code=code, upstream_status=status, context=ctx)
    logger.error(
        "DNS provider failure for DNS request (status=%s)",
        status,
        extra={"upstream_status": status, "error_code": exc.wire_code, **log_only, **ctx},
        exc_info=(code == DnsHostingErrorCodes.UNKNOWN),
    )
    return exc
```

And the context manager:
```
@contextmanager
def _dns_call(self, **context):
    """Catch httpx errors from a Cloudflare call and convert them into typed DnsHostingErrors."""
    try:
        yield
    except (HTTPStatusError, RequestError) as e:
        raise _typed_dns_error(e, **context) from e
```

DaisyGuti (changes requested):
Almost there, just a couple of small clean up items left.

src/registrar/services/cloudflare_service.py
kimallen: Is this a refactor just to further DRY it up? or is there another reason to fold the `DnsTransportError` into `_typed_dns_error`?

docs/developer/dns-error-handling.md
kimallen: Generalized it, though it will soon be replaced with the approved copy.
I also ended up romoving all of the instances of zone-specific error messaging.

src/registrar/utility/errors.py
```diff
@@ -322,19 +320,17 @@ class APIError(Exception):
...
+    RATE_LIMIT_EXCEEDED = 3
+    AUTH_FAILED = 4
+    UPSTREAM_TIMEOUT = 5
+    UPSTREAM_ERROR = 6
+    UNKNOWN = 7
 
 
 _DNS_WIRE_CODES = {
-    DnsHostingErrorCodes.ZONE_NOT_FOUND: "DNS_ZONE_NOT_FOUND",
-    DnsHostingErrorCodes.RECORD_CONFLICT: "DNS_RECORD_CONFLICT",
```
kimallen, line 337: I ended up removing this one. I did some testing of the CF API, and a record conflict is actually a 400 error (not 409), so not separate from the other 400 errors.

src/registrar/services/cloudflare_service.py
DaisyGuti: Yes, STATUS_TO_ERROR only lists the codes we handle - when we do `_STATUS_TO_ERROR.get(status, (None, None))` any status that isn't in that will come back as (None, None).

This function catches these leftovers right:
1. if exc_cls is None and 500 <= status <= 599 -> it's some 5xx we didn't list -> treat as upstream error
2. if exc_cls is None -> still unmapped and not a 5xx -> fall back to `UNKNOWN`

Now Cloudflare can send us any status we didn't map and also isn't 5xx (which is fine), and the `UNKNOWN` catch guarantees we still have a valid typed error. The whole point is we don't have to enumerate every possible HTTP status. 

But an httpx failure can be a HTTPStatusError or RequestError (has not response). So the two cases need handling: one classifies by status, the other has no status to classify by so that's why the transport case can't go through _STATUS_TO_ERROR — there's nothing to map technically. We assign it to DnsTransportError / UPSTREAM_TIMEOUT (the planned / doc's "catch-all for transport") and leave status = None. (and as you prob noticed just an additional note though the order matters, 4xx before 5xx before any crazy 5xx lol)

src/registrar/assets/src/js/getgov/form-errors.js
kimallen: I don't know what was added and removed here, so don't know how to fix it, so not going to waste time fussing with it.

src/registrar/services/cloudflare_service.py
kimallen: We were catching non-HTTPStatusErrors before as a RequestError and raising as a Transport error, so it seems like it's doing the same thing whether we include RequestError inside of _typed_dns_error or outside of it. I did go ahead and update it, but this is what I had. The only difference I see is that the RequestError block doesn't include the log. I was wondering if there was something additional happening in your example folding the RequestError into `_typed_dns_error`
```
        except (HTTPStatusError) as e:
            raise _typed_dns_error(e, **context) from e  # note: `from e` to keep context
        except RequestError as e:
            raise DnsTransportError(
                code=DnsHostingErrorCodes.UPSTREAM_TIMEOUT,
                context={**context, "exc_class": type(e).__name__},
            ) from e

File: src/registrar/services/cloudflare_service.py [hunk 26]
```diff
@@ -83,40 +153,28 @@ def update_account_dns_settings(
             }
         }
 
-        try:
+        with self._dns_call(x_account_id=account_id, zone_mode=zone_mode, nameservers_type=nameservers_type):
             resp = self.client.patch(appended_url, json=data)
             resp.raise_for_status()
+            if not resp.json().get("success"):
```
asaki222: I noticed above this line, we have `raise_for_status`. Why are we usually an additional check? Maybe I am misunderstanding here

src/registrar/validations.py
```diff
@@ -64,7 +64,7 @@
 DNS_RECORD_CONTENT_REQUIRED_ERROR_MESSAGE = "Enter the content for this record."
 DNS_RECORD_PRIORITY_REQUIRED_ERROR_MESSAGE = "Enter a priority for this record."
 DNS_RECORD_PRIORITY_RANGE_ERROR_MESSAGE = "Enter a priority number between 0-65535."
-DNS_RECORD_NAME_CONFLICT_ERROR_MESSAGE = "A record with that name already exists. Names must be unique."
+DNS_RECORD_NAME_CONFLICT_ERROR_MESSAGE = "A record with that name already exists. CNAME names must be unique."
 DNS_RECORD_CNAME_CONFLICT_ERROR_MESSAGE = "A CNAME record with that name already exists."
```
asaki222, line 68: We already have an error message below this one for CNAME duplicates.

src/registrar/admin.py
```diff
@@ -5605,6 +5606,12 @@ def do_enroll_dns_hosting(self, request, obj):
         except EnrollmentNotAllowedError as e:
             logger.warning("DNS enrollment blocked: %s", e)
             self.message_user(request, str(e), messages.WARNING)
+        except DnsHostingError:
+            self.message_user(
+                request,
+                "Failed to enroll domain in DNS hosting.",
+                messages.ERROR,
```
asaki222, line 5614: Could we have the error message in stored in variable since it appears twice in here?

src/registrar/services/cloudflare_service.py
asaki222: I saw the slack thread! Can you add a comment here about the reason behind this??

src/registrar/validations.py
kimallen: That's different. That one is for showing up on A and AAAA records if a CNAME record already exists. The one I updated is the one that shows up if they are on a CNAME record and use a duplicate CNAME record name. It's specific to CNAME because other record names do not need to be unique within the record type.

src/registrar/services/cloudflare_service.py
kimallen: That was me starting to include a recommendation Daisy posted in a Slack thread but I decided to punt to either another PR or until I had more info (it was incomplete code I forgot to remove). But anyway, in future code it will be to catch an 200 responses that still show "success" as False (so 200s that are actually errors- I guess that can happen sometimes, and it would be a defensive check).

I removed that incomplete code.

src/registrar/admin.py
kimallen: Sure. Done.

docs/developer/dns-error-handling.md
DaisyGuti, line 1: "| Cloudflare 409 | Duplicate record (same name+type). Rare — the local model validation at `dns_record.py` should catch most duplicates first; this fires only on races / vendor-side duplicates. | `DNS_RECORD_CONFLICT` | Inline field error using the existing model validation string ("A record with that name already exists. Names must be unique.") | warning |"

Need to update the documentation with the changes made.

DaisyGuti: I would give it a whole review - assure its up to date.

src/registrar/services/cloudflare_service.py
```diff
@@ -24,28 +24,53 @@
...
+        if exc_cls is None and 500 <= status <= 599:
+            exc_cls, code = DnsUpstreamError, DnsHostingErrorCodes.UPSTREAM_ERROR
+        if exc_cls is None:  # Represents 400s we didn't map to a specific status code
+            exc_cls, code = DnsHostingError, DnsHostingErrorCodes.UNKNOWN
+
+    else:  # RequestError -> no response, transport failure
+        status = None
+        ctx = {"exc_class": type(e).__name__, **context}
+        log_only = {}
+        exc_cls, code = DnsTransportError, DnsHostingErrorCodes.UPSTREAM_TIMEOUT
```
DaisyGuti, line 68: `    exc = exc_cls(code=code, upstream_status=status, context=ctx)
`

src/registrar/services/cloudflare_service.py
```diff
@@ -24,28 +24,53 @@
...
 
     unexpected_code = code == DnsHostingErrorCodes.UNKNOWN
     logger.error(
         "Dns provider returned %s for DNS request",
         status,
-        extra={"upstream_status": status, "error_code": code, "response_body": e.response.text, **ctx},
+        extra={"upstream_status": status, "error_code": code, **log_only, **ctx},
```
DaisyGuti: `        extra={"upstream_status": status, "error_code": exc.wire_code, **log_only, **ctx},
`

src/registrar/services/cloudflare_service.py
```diff
@@ -24,28 +24,53 @@
...
         exc_info=unexpected_code,
     )
     return exc_cls(code=code, upstream_status=status, context=ctx)
```
DaisyGuti: `    return exc
`

src/registrar/services/cloudflare_service.py
DaisyGuti: We want to return the string rep from the enum (easier, best for searching and debugging in the logs)

src/registrar/validations.py
kimallen: There's actually no approved copy (no line in the spreadsheet) for that situation, even though we are handling it. So I'll let Content know we need approved copy.

DaisyGuti (dismissed):
It looks really great! It looks like Aditi covered a lot! I was making notes but saw changes popping in as I was reviewing.

Small update needed in the service to pull the String out and we're good to go 🥇

docs/developer/dns-error-handling.md
```diff
@@ -205,41 +202,72 @@ _STATUS_TO_ERROR = {
     401: (DnsAuthError,       DnsHostingErrorCodes.AUTH_FAILED),
     403: (DnsAuthError,       DnsHostingErrorCodes.AUTH_FAILED),
     404: (DnsNotFoundError,   DnsHostingErrorCodes.NOT_FOUND),
-    409: (DnsValidationError, DnsHostingErrorCodes.RECORD_CONFLICT),
     429: (DnsRateLimitError,  DnsHostingErrorCodes.RATE_LIMIT_EXCEEDED),
 }
 
+def _dns_call(self, **context):
```
erinysong, line 208: thank you for updating the docs to reflect our new error flow!

erinysong (approved):
Refactor to use wire code instead of the digit code and doc updates look good. Thank you for making these changes and for the people who gave thorough reviews before!

asaki222 (approved):
LGTM! This is great work. :-)

kimallen:
Review [thread](https://github.com/cisagov/manage.get.gov/pull/5128) for reference

PR #5139, opened 2026-06-26
#5127 - Prevent open redirects via the OIDC next redirect parameter (sm)

### LINKED ISSUE #5127: Prevent open redirects via the OIDC next redirect parameter
### Issue description

djangooidc/views.py:79 stores request.session["next"] = request.GET.get("next", "/") without validating that the URL is safe. The value is used in redirect(request.session.get("next", "/")) at line 126 (normal post-auth redirect) and at line 138 (the StateMismatch retry path).

An attacker constructs a phishing link https://manage.get.gov/openid/login/?next=https://attacker.example/. The victim sees a legitimate registrar URL, completes Login.gov, and is silently redirected to the attacker's domain afterward. Same issue on the StateMismatch path: a user whose session was wiped between auth-start and callback is redirected via the same unvalidated next.

It looks like we haven't set a default for the utility, which we should do, and which should prevent this from happening.

### Acceptance criteria

- [ ] Validate the described solution handles these redirects appropriately, and that there isn't a better solution
- [ ] djangooidc/views.py validates next with django.utils.http.url_has_allowed_host_and_scheme(url=next, allowed_hosts=...) at storage time (line 79), falling back to / on invalid input.
- [ ] Both redirect call sites (line 126 normal flow, line 138 StateMismatch retry) inherit the validation by reading the already-validated session value.
- [ ] Regression tests: external URLs and protocol-relative URLs (//attacker.example/) in ?next= do not redirect off-site.

### Additional context

_No response_

### Links to other issues

_No response_

smishraECS:
## Ticket

Resolves #5127 

## Changes

- Updated redirect logic to validate that redirects are not allowed to external URLs

## Context for reviewers

This security fix is required as it allowed for a vulnerability in which an attacker was able to insert a redirect to a malicious link by forcing a redirect using the ?next property post login.

Confirm that external URL redirection is not allowed post login:
https://getgov-sm.app.cloud.gov/openid/login?next=https://google.com/

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons
  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References


src/djangooidc/views.py
```diff
@@ -123,7 +128,11 @@ def login_callback(request):
 
             # Clear the flag if the exception is not caught
             request.session.pop("redirect_attempted", None)
-            return redirect(request.session.get("next", "/"))
+            if url_has_allowed_host_and_scheme(request.GET.get("next", "/"), None):
```
therealslimhsiehdy: **Question:** I could be wrong here, but I think we just need this check in openid() and so the check here isnt needed if I'm walking through this correctly (?)

Because when a user comes in, they hit something like `/login?next=/domains/` (or google.com per your example) which is in `openid()`, and by the time they come back it would be `login_callback` but it would be read from the session so there wouldn't be any redirection, right?

therealslimhsiehdy (commented):
Left a question + would you mind adding unit tests for this (sorry for being that person 😢)?

src/djangooidc/views.py
smishraECS: Good catch! This was meant to be for the StateMismatch exception. Fixed.

src/djangooidc/views.py
```diff
@@ -144,7 +140,11 @@ def login_callback(request):
             # In the event of a state mismatch between OP and session, redirect the user to the
             # beginning of login process without raising an error to the user. Attempt once.
             logger.warning(f"No State Defined: {nsd_err}")
-            return redirect(request.session.get("next", "/"))
+            if url_has_allowed_host_and_scheme(request.GET.get("next", "/"), None):
```
therealslimhsiehdy, line 143: **Question:** Thoughts on this bc maybe I'm getting the auth flow mixed up

```
if url_has_allowed_host_and_scheme(request.session.get("next", "/"), None):
    return redirect(request.session.get("next", "/"))
else:
    logger.warning(f"Invalid redirect: {request.session.get('next')}")
    return redirect("/")
```

My thought is that the `?next= ` parameter only exist ever on the first request which is when the user hits `/login?next=etc` before being sent off to logingov, and then once they're past that point (success, state mismatch, etc) the URL they're on no longer carries that parameter, so then wouldnt `request.GET.get("next")` be empty or missing every time? Which I think would then mean session (where the values will be held) would be the reference past the first request?

(Pls let me know if I'm going down the wrong whacky path as I'm trying to wrap my head around this)

smishraECS: Because the flow redirects the user to the beginning of the login process, we want to ensure that post login (in case of some sort of attempted CSRF exploit) that we again aren't redirecting off-site. In the case of a StateMismatch where we attempt to reauthenticate the user (stale session cookie or some other session based error) that we aren't allowing for the redirect as I don't believe that flow would go through OIDC.

therealslimhsiehdy: Omg I just re-read the top comment of "redirect the user to the beginning of login process without raising an error to the user." I'm facepalming myself, thanks for the clarification!

therealslimhsiehdy (approved):
LGTM, thank you for all your responses!!

src/djangooidc/views.py
abe-alam-ecs: @smishraECS -- I'm slow. Isn't `request.session.get("/")` looking up a session key named `/` and then returning `None`? 

I don't think this case is actually reachable unless there are stale cookies, expired sessions, etc, like you said.  I think the else block below might raise a TypeError, though...

smishraECS: @abe-alam-ecs are you asking about if url_has_allowed_host_and_scheme(request.session.get("next", "/"), None)?

The None there is the allowed external host whitelist, we don't currently have any allowed hosts so the list is None

abe-alam-ecs: Clarified over Slack, I was referring to line 147. Completing re-review now, as Samir agreed that the line should be `redirect("/")`

abe-alam-ecs (approved):
LGTM. I reviewed this PR on Samir's sandbox and tried redirecting to Google. I was dumped back at the home page. The test also looks appropriate. 

Approved.

PR #5205, opened 2026-07-16
UAT invitations and permissions fixes [cw]

### LINKED ISSUE #4698: UAT for invitations and permissions
### Issue description

After the bulk of the dev work has been completed, we want to conduct UAT testing to catch any bugs or areas of confusion. We want to invite analysts to the UAT session(s) as well since they are directly impacted by the changes we make to invitations and permissions in admin. 

### Acceptance criteria

- [ ] Schedule UAT session(s) 
     - [ ] Notify team (including analysts) of session(s)
     - [ ] Send cal invites and meeting links
- [ ] Create UAT testing doc/spreadsheet 
- [ ] Discuss any issues that came up from testing
     - [ ] Create GitHub tickets for any issues that need to be addressed

### Additional context

_No response_

### Links to other issues

_No response_

chaswick:
## Ticket

Resolves #4698 

## Changes

- Handle multiple existing users with the same email address
- Alert wording changes for clarity about issues / problems including "Member role"
- Alert wording changes on inability to send and email due to allow list restriction
- Remove double period in "An unexpected error occurred: <email> could not be added to this domain.."
- Refinement to the custom admin select box help text location
- Success message refinements

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons
  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References


witha-k (changes requested):
The error alert when an org admin invites an email not in the allowed list still has typos:
> An unexpected error occurred: kristina.yin+help@cisa.dhs.gov could not be added to this domain.. Please try again. If the problem persists, [contact us](https://get.gov/contact/) for assistance.

Analyst accounts still can't see the "User domain roles" page in /admin.

src/registrar/models/user_group.py
allly-b, line 17: **blocking change request:** Missing corresponding migration - Anytime either permission group is altered the dev needs to also add a migration so that the change takes effect on stable and staging. Local can mask this, because it will work totally fine locally but sandboxes will struggle.  see [this example ](https://github.com/cisagov/manage.get.gov/blob/b40aa66e8f060c5439f7191126d1cce8abe2687f/src/registrar/migrations/0192_create_groups_v22.py)it has the exact steps in the "alternatively" section.

src/registrar/services/invitation_service.py
```diff
@@ -29,11 +30,15 @@
...
     """Retrieve a user by email or return None if the user doesn't exist."""
-    try:
-        return User.objects.get(email__iexact=email)
-    except User.DoesNotExist:
+    if not email:
         return None
 
+    users = list(User.objects.filter(email__iexact=email).order_by("pk")[:2])
+    if len(users) > 1:
+        raise MultipleUsersWithEmailError(email)
```
allly-b, line 38: This will still throw an error on the [domain view ](https://github.com/cisagov/manage.get.gov/blob/8254f0db3a6a7f6849fdb37320f154e251c6ebb7/src/registrar/views/domain.py#L1643-L1644)and result in a 500 not the new message. 

I also see the domain invitation admin isn't catching this error either, [see here](https://github.com/cisagov/manage.get.gov/blob/8254f0db3a6a7f6849fdb37320f154e251c6ebb7/src/registrar/admin.py#L2923). Again the user just sees a 500 error with no info. 

**blocking question** In cases where Multiple Users have the same email, it sounds like we want it to at least give a human readable error message, is that right? (if so, this question becomes a change request to just handle errors in these two spots too)

src/registrar/admin.py
```diff
@@ -2372,6 +2389,16 @@ def user_or_email(self, obj):
     def _use_invitation_admin(self, request):
         return flag_is_active(request, "user_domain_role_invitations")
 
+    def message_user(self, request, message, level=messages.INFO, extra_tags="", fail_silently=False):
+        if (
+            self._use_invitation_admin(request)
+            and level == messages.SUCCESS
+            and "was added successfully" in str(message)
```
allly-b, line 2402: **Nitpick/non mandatory:** I'd suggest adding a comment here about why "was added successfully" is being checked. I'm assuming that this is to avoid 2 messages appearing here.

src/registrar/models/portfolio_invitation.py
```diff
@@ -182,21 +183,24 @@ def members_description_display(self):
         return get_members_description_display(self.roles, self.additional_permissions)
 
     @transition(field="status", source=PortfolioInvitationStatus.INVITED, target=PortfolioInvitationStatus.RETRIEVED)
-    def retrieve(self):
+    def retrieve(self, user=None):
```
allly-b, line 186: **change request** I'm still seeing a portfolio object calling the normal .retrieve() in admin.py 

- [finding 1](https://github.com/cisagov/manage.get.gov/blob/8254f0db3a6a7f6849fdb37320f154e251c6ebb7/src/registrar/admin.py#L2948)
-  [finding 2](https://github.com/cisagov/manage.get.gov/blob/8254f0db3a6a7f6849fdb37320f154e251c6ebb7/src/registrar/admin.py#L3067)

src/registrar/models/domain_invitation.py
```diff
@@ -49,21 +50,24 @@ def __str__(self):
         return f"Invitation for {self.email} on {self.domain} is {self.status}"
 
     @transition(field="status", source=DomainInvitationStatus.INVITED, target=DomainInvitationStatus.RETRIEVED)
-    def retrieve(self):
+    def retrieve(self, user=None):
```
allly-b, line 53: **change request:** I'm still seeing some places still using retrieve()

[here a domain invitation object is calling retrieve() not passing a user](https://github.com/cisagov/manage.get.gov/blob/8254f0db3a6a7f6849fdb37320f154e251c6ebb7/src/registrar/admin.py#L2962-L2963)

there looks like another instance [here](https://github.com/cisagov/manage.get.gov/blob/8254f0db3a6a7f6849fdb37320f154e251c6ebb7/src/registrar/admin.py#L2962)

src/registrar/models/domain_invitation.py
```diff
@@ -49,21 +50,24 @@ def __str__(self):
...
-            user = User.objects.get(email__iexact=self.email)
-        except User.DoesNotExist:
-            # should not happen because a matching user should exist before
-            # we retrieve this invitation
-            raise RuntimeError("Cannot find the user to retrieve this domain invitation.")
+        if user is None:
+            User = get_user_model()
+            users = list(User.objects.filter(email__iexact=self.email).order_by("pk")[:2])
+            if len(users) > 1:
+                raise MultipleUsersWithEmailError(self.email)
```
allly-b, line 65: **suggestion:** This section of code is used three times, and I imagine if we make something else that needs user emails we will want to copy the same logic. What if we just moved this to a utility and just called the utility function in the places that need it?

Also I suggest always putting a comment if you adding a number into the code, especially on a line like this that has several actions being preformed. 

**question:** Why are you doing :2 ? I'm not seeing the need, but maybe I'm missing something?

src/registrar/models/domain_invitation.py
```diff
@@ -49,21 +50,24 @@ def __str__(self):
...
-        try:
-            user = User.objects.get(email__iexact=self.email)
-        except User.DoesNotExist:
-            # should not happen because a matching user should exist before
-            # we retrieve this invitation
-            raise RuntimeError("Cannot find the user to retrieve this domain invitation.")
+        if user is None:
+            User = get_user_model()
+            users = list(User.objects.filter(email__iexact=self.email).order_by("pk")[:2])
+            if len(users) > 1:
```
allly-b, line 64: **non-mandatory/non blocking suggestion** I don't think you need to call list() on line 63. Imagine that you don't do the direct list conversion and `users` has the queryset results.  On line 64 you could then do `if users.count() >1:` and below on line 70 you could do `user= users.first()`

It works either way, so not a must change at all.

src/registrar/admin.py
allly-b: Coming back to this comment as I noticed this now exists in a couple places in the code AND if we upgrade django this default django success message could change and it would unlikely that someone remembers to update these lines. Two  new suggestions things in addition to the above:

-  please add in a comment something along the lines of "confirm this still works after any django upgrade".  
-  consider having this string be a constant that is used by the files that need it (not a huge deal though if you skip this suggestion since it's only appearing in 3 places at the moment).

src/registrar/admin.py
```diff
@@ -637,7 +642,9 @@ def _validate_invitation_email_can_be_sent(self, cleaned_data, user, email):
         if models.AllowedEmail.is_allowed_email(email):
             return
 
-        self.add_error("user", f"Could not send email. The email '{email}' does not exist within the allowlist.")
+        self.add_error(
+            "user", "Can't send invitation email because this user doesn't exist in the allowed emails list."
```
allly-b: **question/non-blocking:** Should this be "because this email doesn't exist in ..." since the allowed emails list is -- well -- a list of allowed emails not users?

allly-b (changes requested):
Some minor change requests below. If I say suggestion or nitpick, they are minor things that aren't mandatory by any means unless you agree that they should be changed. The ones marked change request are the ones truly blocking the approval on my end.

src/registrar/models/user_group.py
chaswick: Creating.

src/registrar/services/invitation_service.py
chaswick: That is correct.  Will update admin and domain to fix.

src/registrar/models/domain_invitation.py
chaswick: Will fix.

src/registrar/models/portfolio_invitation.py
chaswick: Will fix.

src/registrar/admin.py
chaswick: Adding explanation/hint. Deferring creating a constant since it is only used twice.

src/registrar/admin.py
chaswick: Your a user, she's a user, we're all users today!  Updating to fix bad user reference, it is an email

src/registrar/models/domain_invitation.py
chaswick: I will elect not to do this right now, if its ok.  Specifically chopping off potentially a very large collection (we could have any number of users matching) in case it helps performance, we only care if there is more than 1.

chaswick:
> The error alert when an org admin invites an email not in the allowed list still has typos:
> 
> > An unexpected error occurred: [kristina.yin+help@cisa.dhs.gov](mailto:kristina.yin+help@cisa.dhs.gov) could not be added to this domain.. Please try again. If the problem persists, [contact us](https://get.gov/contact/) for assistance.
> 
> Analyst accounts still can't see the "User domain roles" page in /admin.

Fixed double period

src/registrar/models/domain_invitation.py
chaswick: I think that is two queries instead of 1 plus a list operation, but good point.  Will defer making this change.

src/registrar/admin.py
allly-b: It still says "use" fyi, in case you forgot to push. Your previous comment made it sound like that was being changed so just a heads up, but again, if you want to leave it as is that also works for me. Definitely not blocking

src/registrar/services/invitation_service.py
allly-b, line 740: **Question:** should this take the user parameter too? `legacy_invitation.retrieve()`

src/registrar/models/domain_invitation.py
allly-b: Just for nerdiness sake I will say, the code User.objects.filter(email__iexact=self.email).order_by("pk")[:2] doesn't execute the query, it's adding list() around it that triggers the query to execute. So likewise, doing users = User.objects.filter(email__iexact=self.email).order_by("pk")[:2] doesn't execute the query yet. Calling .count() would trigger the execution making it still a total of 1 queries. I say this almost entirely because I find this pretty unique to django and worth sharing; but your code as-is is perfect without changing it!

allly-b (approved):
Looks great. Only one question regarding a `legacy_invitation.retrieve()` and some nitpicks. Approving as is, just be sure to read and resolve/comment on the most recent comments before merging.

witha-k (changes requested):
Analysts can now "view" the "User domain roles" page but they can't create a new user domain role entry.

witha-k (changes requested):
Also realized that this error message is shown like this in the public-facing registrar. Is it possible to fix the typo for all instances of this error message?

Image

chaswick:
> Analysts can now "view" the "User domain roles" page but they can't create a new user domain role entry.

I think this was because I had another PR branch deployed to the sandbox.  Worth retresting.

chaswick:
> Also realized that this error message is shown like this in the public-facing registrar. Is it possible to fix the typo for all instances of this error message?
> 
> Image

I checked and it appears all errors are passing their own period, so I fixed it at the top level.  Should only be one period now.

src/registrar/admin.py
chaswick: Updated to 'email' now.

src/registrar/services/invitation_service.py
chaswick: Yes, I have updated.

witha-k (approved):
Everything seems to work now! Great job, thank you!

PR #5210, opened 2026-07-16
#4887 Remediate case sensitive queries (iexact patch)

### LINKED ISSUE #4887: Remediate case sensitive queries on our codebase
### Issue description

Fix

### Acceptance criteria

- [ ] TBD from #4317

### Additional context

_No response_

### Links to other issues

_No response_

CocoByte:
## Ticket

Resolves #4887 

## Changes

Added __iexact throughout the codebase for e-mail related lookups.  

This addresses the immediate need of preventing errors in invitations and member management flows (we have been getting bugs of e-mail based lookups not resolving correctly due to camel-casing).

NOTE:  There might be instances where it looks like the addition of __iexact is redundant or not needed. Part of the reason for this is to account for the possibility that some existing data might have camel casing that we now prevent at the data store level.  This ticket aims to provide a full patch safeguarding against case-sensitivity until a more robust remediation of our case-sensitivity issues is identified/implemented in #5213.

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons
  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References


chaswick (approved):

src/registrar/services/invitation_service.py
```diff
@@ -417,13 +417,15 @@ def _check_existing_domain_invitation(email: str, domain: Domain, requested_user
...
+    invited = UserDomainRole.objects.filter(
+        email__iexact=email, domain=domain, status=UserDomainRole.Status.INVITED
+    ).exists()
     if invited:
         raise AlreadyDomainInvitedError(email)
 
     # Check for duplicates in legacy model
     try:
-        invite = DomainInvitation.objects.get(email=email, domain=domain)
+        invite = DomainInvitation.objects.get(email__iexact=email, domain=domain)
```
allly-b: **Blocking:** this will throw a MultipleObjectsReturned Error if there are two inviitations with the same email just in different cases. It sounds like that's what this PR is trying to prevent. I think you need to orderby created at and call .first().

src/registrar/utility/email_invitations.py
```diff
@@ -92,7 +92,7 @@ def _check_user_org_admin(requestor_email, domains) -> bool:
 def _validate_existing_invitation(email, user, domain):
     """Check for existing invitations and handle their status."""
     try:
-        invite = DomainInvitation.objects.get(email=email, domain=domain)
+        invite = DomainInvitation.objects.get(email__iexact=email, domain=domain)
```
allly-b: **Blocking:** This will throw the MultipleObjectsReturned Error if there are two inviitations with the same email just in different cases. call first() at least.

src/registrar/models/portfolio_invitation.py
```diff
@@ -192,7 +192,7 @@ def retrieve(self):
         # get a user with this email address
         User = get_user_model()
         try:
-            user = User.objects.get(email=self.email)
+            user = User.objects.get(email__iexact=self.email)
```
allly-b, line 195: **Question :** This isn't checking or catching if 2 users are returned. Do we want to catch the case where this hits an error? We could just catch multiple objects returned error below?

src/registrar/utility/email_invitations.py
allly-b: (non blocking) Also there are 2 .exclude(user__email=email) lower in this file should they be changed as well so that it's using iexact in this PR since you're already touching invitations? Or would this be handled later?

src/djangooidc/backends.py
```diff
@@ -77,6 +77,10 @@ def get_user_by_username(self, UserModel, username):
     def get_user_by_email(self, UserModel, email):
         """Retrieve user by email."""
         try:
+            # User e-mails are lowercased before storage, so we don't need __iexact here.
+            # Note: We rely on login.gov to manage the uniqueness of email addresses in a
+            # case-insensitive manner, so we don't need to check for multiple users with
+            # the same email.
             return UserModel.objects.get(email=email)
```
allly-b, line 84: **Blocking question:** Was stable/staging checked to confirm this? I would assume there are old users prior to the lower casing that still exist in the db. But I didn't verify myself in case this has already been done

allly-b (changes requested):
I found a few that I think need to be changed still but haven't completed the review. 1 blocking question regarding the users.

src/djangooidc/backends.py
allly-b: Edit: I just ran a query off of potato which has an old stable copy, and sure enough 0 cases where users have different cases in the email. We still have duplicate emails for users which will make this fail for returning more than one object, but I think that's what we want in this case. 

Removing block and resolving

src/registrar/models/portfolio_invitation.py
CocoByte: Good callout.  Given that this is a patch, I'm inclined to just add .first() to this as well.  It might be worth considering enforcing uniqueness for users at the DB level when pursuing a more robust fix in the follow-on ticket.

src/registrar/services/invitation_service.py
CocoByte: Thanks for catching these!  After parsing so much of the code I think my eyes started to glaze over all the .get instances.

src/registrar/utility/email_invitations.py
CocoByte: yep, I missed those!  Will add.

src/registrar/models/portfolio_invitation.py
allly-b: You would first need to make sure all the users on stable who have duplicate emails already get resolved into one user account before applying that constraint. Otherwise, I believe the migration will hit an error when you try to apply that rule

allly-b (approved):
Thanks for making the changes!!

src/registrar/models/portfolio_invitation.py
CocoByte: Yep ~ data integrity is partially why those higher-level resolutions were broken out into a follow-on ticket.  Definitely requires more work than just applying rulesets.

Issue #5125, opened 2026-06-22
Run backfill script for created_at_reference and x_registry_created_at in staging and stable


lizpearl:
### Sub-issue description

In #4964,  a script was created to backfill both the registry and registrar `created_at` dates. This ticket covers running them in staging and stable

### Acceptance criteria

- [x] Run backfill script in staging.  Check values
- [ ] Run backfill script in stable. Check values

### Additional context
- Script was created in #4964 

FYSA to run:
```
cf run-task getgov-staging -m 1G --name backfill --command "yes | python manage.py populate_domain_created_at_columns"

cf logs getgov-staging --recent 
#OR
cf logs getgov-staging --recent | grep backfill
```

DaisyGuti:
@katypies @CocoByte @lizpearl This was run on Staging today. Please let me know if you'd like me to run on Stable?

DaisyGuti:
FYSA to run: 
```
cf run-task getgov-staging -m 1G --name backfill --command "yes | python manage.py populate_domain_created_at_columns"

cf logs getgov-staging --recent 
#OR
cf logs getgov-staging --recent | grep backfill
```

lizpearl:
Ran script in staging and stable.  Did a spot check and [output of the script is here.](https://docs.google.com/document/d/1ZErH0Y-GVsamZGJgp8qhVZ9uU98okvlPeGvv5i_7kM0/edit?tab=t.0)

PR #5214, opened 2026-07-20
#4435: update domain create_at reads to the new columns [none]

### LINKED ISSUE #4435: Reference the new columns: create_at_reference and registry_created_at
### Sub-issue description

Update code references from the `created_at` field to the use the correct new column.  The `created_at_reference` field should only ever be read, while the `registry_created_at` field should be written to once.  The code that uses the `created_at_reference` field will only be temporary and will eventually be switched back to created at.

### Acceptance criteria

- [ ] Verify that the script from #4376 was run on stable
- [ ] Code should not reference `created_at` field and should only use `created_at_reference` and `registry_created_at` 

### Additional context

_No response_

DaisyGuti:
## Ticket

Resolves #4435

## Changes

- updated domain `created_at` reads to the new columns: `x_registry_created_at` (registry date) and `created_at_reference` (our registrar record date)
- Added `Domain.display_created_at` property (registry date, falling back to the record date for UNKNOWN domains)
- Added checks 2 spots where the new nullable columns are used in date calcs
- Updated test fixtures to populate the new columns

## Context for reviewers

The old `created_at` was one column holding two different dates, depending on the domain's state:

1. For **Registered domains** = **registry creation** date. A _registry sync_ overwrites the column with the date the domain was _created at the registry_.
2. For state **UNKNOWN domains** = the **registrar record** date. These never reach the registry, so there is no registry date to overwrite it, and it keeps the date we created the record.

Prev PR already merged created 2 new columns split these dates and a script to backfills / loads data.

**One exception**: analytics "average approval time" changed**. `created_at_reference` is the approval moment, so every approved domain has one. `x_registry_created_at` is only set once the domain syncs to the registry, which lags approval, so recently-approved domains that have not synced yet have no value at all and are left out of the average calculation.

We are to merge **after** #5125 runs the backfill on staging and stable — the new columns are empty on existing rows until then.

e.g. to run on Staging:
```
cf run-task getgov-staging -m 1G --name backfill \
  --command "yes | python manage.py populate_domain_created_at_columns"

cf logs getgov-staging --recent
```

I've left inline comments on the specific changes for approve/reject.

## Setup
to run locally and have data in your db assure to run migrations then:
`docker compose exec app ./manage.py populate_domain_created_at_columns`
OR clone_to_local.sh and select staging (default db selected)

### References


src/registrar/models/domain.py
```diff
@@ -271,6 +271,12 @@ def save(self, force_insert=False, force_update=False, using=None, update_fields
             self.created_at_reference = self.created_at
         self._original_updated_at = self.updated_at
 
+    @property
```
DaisyGuti, line 274: New ref: registry date, falls back to the registrar record date.

This keeps what the old created_at held - the registry date for registered domains, and the record date for UNKNOWN domains, which never reached the registry and so have no registry date. Centralizing it here keeps the domain page and the admin column in sync.

**Q: OK here? Per the ticket, the created_at_reference is temporary and updates once UNKNOWN domains get real registry dates.**

src/registrar/templates/includes/domain_dates.html
```diff
@@ -1,12 +1,12 @@
...
   <p>
     {% if domain.expiration_date %}
     <strong class="text-primary-dark">Date of expiration:</strong> 
     {{ domain.expiration_date|date }}
     {% if domain.is_expired %} <span class="text-error"><strong>(expired)</strong></span>{% endif %}
     <br/>
     {% endif %}
-    {% if domain.created_at %}
-    <strong class="text-primary-dark">Date created:</strong> {{ domain.created_at|date }}{% endif %}
+    {% if domain.display_created_at %}
```
DaisyGuti, line 9: UI date reads display_created_at.

Q: Do you approve showing the registry date, with the record date only as a fallback?

src/registrar/admin.py
```diff
@@ -5139,7 +5139,7 @@ def get_annotated_queryset(self, queryset):
         "converted_state_territory",
         "state",
         "expiration_date",
-        "created_at",
+        "created_at_display",
```
DaisyGuti, line 5164: Returns display_created_at so UNKNOWN rows show a date instead of blank.

src/registrar/utility/csv_export.py
```diff
@@ -850,7 +850,8 @@ def get_fields(cls, model):
             "SO": model.get("converted_so_name"),
             "SO email": model.get("converted_so_email"),
             "Security contact email": model.get("security_contact_email"),
-            "Created at": model.get("domain__created_at"),
+            # Registrar record date fills in for UNKNOWN domains, which have no registry date (matches old created_at)
+            "Created at": model.get("domain__x_registry_created_at") or model.get("domain__created_at_reference"),
```
DaisyGuti, line 854: Registry date, falls back to the record date.

The fallback is needed bc DomainDataType (the "my domains" export) filters only on which domains a user manages, with no state filter, so it can include UNKNOWN domains. The other exports exclude UNKNOWN.

Untouched and pre-existing: DomainDataFull and DomainDataFederal output a "Created at" value that is never a select in the query, so the column are empty. This predates the PR. Probably already has a bug ticket...

src/registrar/views/report_views.py
```diff
@@ -27,8 +27,10 @@ def get(self, request):
         last_30_days_approved_applications = models.DomainRequest.objects.filter(
             created_at__gt=thirty_days_ago, status=models.DomainRequest.DomainRequestStatus.APPROVED
         )
+        # created_at_reference is when the domain record was created, which is the approval moment.
```
DaisyGuti, line 30: This is the analytic number in the PR description I mentioned.

- created_at_reference is the approval, so every approved domain has one. 
- x_registry_created_at is only set once a domain syncs to the registry, which lags approval - so recently-approved domains that have not synced yet have no value and are not in the average.

SQl to get the #s:
  ```
SELECT COUNT(d.created_at)                                   AS rows_created_at,
         AVG(d.created_at - dr.last_submitted_date)            AS avg_created_at,
         COUNT(d.created_at_reference)                         AS rows_reference,
         AVG(d.created_at_reference - dr.last_submitted_date)  AS avg_reference,
         COUNT(d.x_registry_created_at)                        AS rows_x_registry,
         AVG(d.x_registry_created_at - dr.last_submitted_date) AS avg_x_registry
  FROM registrar_domainrequest dr
  JOIN registrar_domain d ON d.id = dr.approved_domain_id
  WHERE dr.status = 'approved'
    AND dr.created_at > NOW() - INTERVAL '30 days';
```

File: src/registrar/management/commands/send_domain_setup_reminder.py [hunk 6]
```diff
@@ -120,8 +120,8 @@ def handle(self, *args, **options):
         seven_days_ago_start, seven_days_ago_end = self._get_seven_days_ago_date_range(today)
 
         domains = Domain.objects.filter(
-            created_at__gte=seven_days_ago_start,
-            created_at__lte=seven_days_ago_end,
+            created_at_reference__gte=seven_days_ago_start,
```
DaisyGuti, line 123: Filter now uses created_at_reference.

Finds domains approved 7 days ago that still need DNS setup. Runs only on UNKNOWN/DNS_NEEDED
domains, where the record date is the old created_at value.

src/registrar/management/commands/send_expiring_soon_domains_notification.py
```diff
@@ -51,7 +51,7 @@ def handle(self, *args, **options):  # noqa: C901
             | Q(
                 expiration_date__isnull=True,
                 state__in=[Domain.State.UNKNOWN],
-                created_at__in=[today + timedelta(days=d) - timedelta(days=365) for d in days_to_check],
+                created_at_reference__in=[today + timedelta(days=d) - timedelta(days=365) for d in days_to_check],
```
DaisyGuti, line 54: created_at_reference, because these are all scoped to UNKNOWN.

UNKNOWN domains have a null x_registry_created_at.

src/registrar/management/commands/delete_expired_domains_not_setup.py
```diff
@@ -73,14 +73,23 @@ def get_domains(self):
 
         for domain in domains_in_expired_state:
             if domain.expiration_date is None:
+                if domain.created_at_reference is None:
```
DaisyGuti, line 76: updated to created_at_reference, plus added new null guard.

The "creation + 1yr" fallback only runs when expiration_date is null, which per this method's docstring is the UNKNOWN case.

The old created_at could never be null, the new column can be, and this code deletes domains. So it now logs and skips.

Q: Approve skipping deletion when the date is missing?

src/registrar/management/commands/update_first_ready.py
```diff
@@ -11,26 +11,35 @@ class Command(BaseCommand, PopulateScriptTemplate):
 
     def handle(self, **kwargs):
         """Loops through each valid Domain object and updates it's first_ready value if it is out of sync"""
-        filter_conditions = {"state__in": [Domain.State.READY, Domain.State.ON_HOLD, Domain.State.DELETED]}
+        filter_conditions = {
```
DaisyGuti, line 14: Added x_registry_created_at__isnull=False. The script calls .date() on the value, and the new column is null for anything never registered. So at the queryset level covers the null cases. The local None check in update_record has to be there for mypy and it's an extra safety net.

**Q: Approve skipping domains with no registry date? first_ready cannot be derived without one.**

src/registrar/management/commands/transfer_transition_domains_to_domains.py
```diff
@@ -189,6 +189,7 @@ def update_or_create_domain(self, transition_domain: TransitionDomain, debug_on:
                     # the created_at date was null (violated a key constraint).
                     # How do we want to handle this case?
                     target_domain.created_at = transition_domain_creation_date
+                    target_domain.x_registry_created_at = transition_domain_creation_date
```
DaisyGuti, line 192: Kept the existing created_at write and added x_registry_created_at beside it.

The TODO just above (what to do when the creation date is null) is still open. Though now: created_at is NOT NULL, which is what broke the constraint originally, but x_registry_created_at
is nullable - so "no registry date" is null. 

**Q: Approve the added write. Separately, do we want a follow-up ticket to close the TODO now that a
null registry date is ok?**

src/registrar/management/commands/delete_expired_domains_not_setup.py
lizpearl: This is fine.  After the script backfill, in practice the created_at_reference should never be null.  I know it's not a db requirement but didn't want to put that in just to remove the column at the end.  I do like this here in case we do hit this case and then we'll know something is up

src/registrar/management/commands/populate_first_ready.py
```diff
@@ -54,11 +54,11 @@ def handle(self, **kwargs):
...
     def update_first_ready_for_domain(self, domain: Domain, debug: bool):
-        """Grabs the created_at field and associates it with the first_ready column.
+        """Grabs the registry creation date and associates it with the first_ready column.
         Appends the result to the to_update list."""
-        created_at = domain.created_at
-        if created_at is not None:
-            domain.first_ready = domain.created_at
+        registry_created_at = domain.x_registry_created_at
+        if registry_created_at is not None:
+            domain.first_ready = domain.x_registry_created_at
```
lizpearl, line 61: Note to other reviewers, this was a management script we ran to fill in our best estimate of `first_ready`. We don't regularly run this and this is not our typical definition of `first_ready` which is when the domain first moved into the ready state

src/registrar/management/commands/transfer_transition_domains_to_domains.py
lizpearl: The addition is fine.  No need to open a new ticket since this script is old and would most like need to create a new one if doing another transition

src/registrar/management/commands/update_first_ready.py
lizpearl: Yes, this is fine!   I ended up add an AC on the last ticket of this epic to add db constraints that made sense so we don't have to constantly do checks.

src/registrar/models/domain.py
lizpearl: Yes, that looks fine

abe-alam-ecs (approved):
Reviewed the full diff and think the logic checks out. The null-handling, save() ordering, and custom_filter/filter_conditions sequencing are correctly guarded. 

A few items that I want to verify:
- Are we OK skipping domains without a registry date? Seems like the answer is yes based on what Liz commented in `populate_first_ready.py`
-  Daisy wrote this in `domain_dates.html`....

> UI date reads display_created_at.
> 
> Q: Do you approve showing the registry date, with the record date only as a fallback?

No bugs found on testing locally. Approving as I don't think any of these are blockers unless the answer to the above changes.

src/registrar/utility/csv_export.py
```diff
@@ -1103,7 +1105,7 @@ def get_fields(cls, model):
             "SO": model.get("so_name"),
             "SO email": model.get("senior_official__email"),
             "Security contact email": model.get("security_contact_email"),
-            "Created at": model.get("domain__created_at"),
+            "Created at": model.get("domain__x_registry_created_at"),
```
therealslimhsiehdy: **Question:** This drops the `UNKNOWN` fallback that above (line 853) has, should this also be `model.get("domain__x_registry_created_at") or model.get("domain__created_at_reference")`?

src/registrar/utility/csv_export.py
```diff
@@ -1230,7 +1232,7 @@ def get_fields(cls, model):
...
```
therealslimhsiehdy: Same as line 1108 --  should this be `model.get("domain__x_registry_created_at") or model.get("domain__created_at_reference")` as well?

src/registrar/utility/csv_export.py
therealslimhsiehdy: Also if yes, it would repeat 3 times in this file - worth putting into a helper maybe?

src/registrar/views/report_views.py
```diff
@@ -27,8 +27,10 @@ def get(self, request):
...
+        # FYSA: x_registry_created_at is null until a domain reaches the registry.
         avg_approval_time = last_30_days_approved_applications.annotate(
-            approval_time=F("approved_domain__created_at") - F("last_submitted_date")
+            approval_time=F("approved_domain__created_at_reference") - F("last_submitted_date")
```
therealslimhsiehdy, line 33: **Question:** Is the "average approve time" changed in the PR description what you were referring to for this calculation?

(Sorry still trying to wrap my head around this)

>One exception: analytics "average approval time" changed**. created_at_reference is the approval moment, so every approved domain has one. x_registry_created_at is only set once the domain syncs to the registry, which lags approval, so recently-approved domains that have not synced yet have no value at all and are left out of the average calculation.

therealslimhsiehdy (commented):
Left some comments as I'm still grasping the PR, will come back on Monday and do another round + test this out.

therealslimhsiehdy (approved):
LGTM (did leave some comments/question) - tested that:
1. `Domain.save()` auto populates `created_at_reference` while leaving `x_resgistry_created_at` untouched 
2. `display_created_at` falls back to `created_at_reference` when no registry date exists + tested `display_created_at` prefers `x_registry_created_at` over `created_at_reference when both exist (bc registry date should always win once it's available` 
3. resaving a domain doesn't overwrite an alr set `created_at_reference`

src/registrar/views/report_views.py
lizpearl: Yes, that's the one!  For newer domains `created_at_reference` <= `created_at` so that approval time will almost definitely decrease unless everyone who had an approved domain logged in and looked at their domain immediately after approval.

src/registrar/utility/csv_export.py
lizpearl: Updated those lines!  I decided not to make a helper function because it fits into a single line and it's one less thing to look up!

src/registrar/templates/includes/domain_dates.html
lizpearl: Yes this is fine!

Issue #5053, opened 2026-05-29
Delete CF DNS account when a domain is deleted


kimallen:
### Issue description

When we delete a domain, we need to remove all DNS -related records and resources associated with that domains. That means all db records (account, zone and record) data will be removed. That is covered in this [ticket](https://github.com/cisagov/manage.get.gov/issues/4560). 
We also need to delete associated resources in cloudflare. Deleting an associated account will remove all of it's zones and records as well.

We need to ensure we remove CF accounts when a domain is deleted.

To do this we need:
-  account deletion API in CloudflareService
- a method in DnsHostService that calls a account deletion in CloudflareService
- call the DnsHostService in the appropriate place triggered by domain deletion

Do this first in a transaction with db account/zone/record deletions such that if CF deletion fails, we don't delete from db yet.

Feel free to break this work down into multiple PRs.

### Acceptance criteria

- [x] When a domain is deleted in our application, the corresponding CF account (and related resources) are also deleted
- [x] If CF account deletion fails, the db account/zone/record deletions likewise do not
- [x] Incorporate a MockCloudflareService success response for account deletion 
- [x] Corresponding tests

### Additional context

In order to test if the CF account (and zone and records) have been deleted, you can check for it using the GET api or look in the CF UI for the account (you can add yourself as a member of the account in order to view it in the UI: 
```
curl --location 'https://api.cloudflare.com/client/v4/accounts/:account_id/members' \
--header 'X-Auth-Email: REDACTED \
--header 'X-Auth-Key;' \REDACTED
--header 'Content-Type: application/json' \
--data-raw '{
  "email": [YOUR_EMAIL],
  "auto_accept": true,
  "roles": ["05784afa30c1afe1440e79d9351c7430"],
  "status": "active"
}'
```
** the role here refers to Adminstrator role **

### Links to other issues

- Relates to https://github.com/cisagov/manage.get.gov/issues/4560

PR #5225, opened 2026-07-22
#4560 Cascade delete DNS objects from db when deleting domain in EPP

### LINKED ISSUE #4560: Set up cascading deletes of all dns hosting tables when a domain is deleted
### Issue description

Currently we have cascading delete of a record in the DnsZone when a domain is deleted. However, the records on the VendorDnsZone and DnsZone_VendorDnsZone tables remains.

Similarly, when the DnsZone record gets deleted, associated DnsRecords get deleted, but the other tables' corresponding records (VendorDnsRecord and DnsRecord__VendorDnsRecord) don't.

Also, an account does not get deleted when a domain is deleted because the Domain is associated with a Zone, not an account. So we also want to delete the associated dns account records on all 3 tables)

### Acceptance criteria

_No response_

### Additional context

_No response_

### Links to other issues

_No response_

erinysong:
## Ticket

Resolves #4560

## Changes

- When an registrar admin/analyst deletes a domain from EPP, also delete the domain's DNS data from the database.

## Context for reviewers

Ticket does not include handling an EPP-deleted domain's DNS data on Cloudflare. This will be addressed in #5053 

## Setup

This PR can be tested either by reviewing and running the added unit test `test_delete_domain_in_epp_deletes_db_dns_data` or checking that a domain's associated DNS data is successfully removed from the database. 

**Option 1: Review and run unit test `test_delete_domain_in_epp_deletes_db_dns_data`**
This unit test tests that after a domain is deleted in EPP, the domains' associated DNS account, zone, and record data is fully removed from the registrar database.

**Option 2: Test manually**
Alternatively, you can go to the Domains table on /admin and find a Domain enrolled in DNS hosting (Beekeeper can be used to identify domains enrolled in DNS hosting). Verify the domain is enrolled in DNS hosting and has associated DNS account, zone, and record (you may need to add records to the domain yourself) data in the database. If you test using this route, I recommend keeping track of the database id's of the domain's DNS objects (`DnsX`, `VendorDnsX`, DnsX_VendorDnsX`).

On the domain page, click the Delete from registry button on the top right corner
image

Verify that all the DNS objects you documented from earlier no longer exist in the database. If you're testing locally from fresh fixtures, you can also quick-check that the objects got deleted by checking there is one less row than the highest id number (for example, a DnsAccount table with id's that go all the way to 59 may have 58 rows since we deleted a DnsAccount).

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons
  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References


src/registrar/models/domain.py
```diff
@@ -1275,6 +1277,54 @@ def _delete_dnssecdata(self):
...
+        )
+
+        if self.is_enrolled_in_dns_hosting:
+            logger.debug("Deleting DNS data for %s.", self.name)
+            try:
+                with transaction.atomic():
+                    dns_zone = DnsZone.objects.get(domain_id=self.id)
+                    logger.info("Removing db DNS records associated with %s.", self.name)
+                    records = DnsRecord.objects.filter(dns_zone=dns_zone)
+                    # Deleting DnsRecord cascade deletes associated DnsRecord_VendorDnsRecord.
```
erinysong: [Slack thread](https://cisa-corp.slack.com/archives/C05BGB4L5NF/p1784738634391259) with further discussion
We manually delete the associated VendorDnsObject connected to the `DnsObject_VendorDnsObject` that is cascade deleted from deleting the `DnsObject`. Open to other suggestions to handle deleting the other parent fk.

src/registrar/models/domain.py
```diff
@@ -1275,6 +1277,55 @@ def _delete_dnssecdata(self):
...
+                DnsRecord_VendorDnsRecord,
+            )
+
+            logger.debug("Deleting DNS data for %s.", self.name)
+            try:
+                with transaction.atomic():
+                    dns_zone = DnsZone.objects.get(domain_id=self.id)
+                    logger.info("Removing db DNS records associated with %s.", self.name)
+                    records = DnsRecord.objects.filter(dns_zone=dns_zone)
+                    # Deleting DnsRecord cascade deletes associated DnsRecord_VendorDnsRecord.
```
erinysong, line 1310: [Slack thread](https://cisa-corp.slack.com/archives/C05BGB4L5NF/p1784738634391259) with further discussion
We manually delete the associated VendorDnsObject connected to the DnsObject_VendorDnsObject that is cascade deleted from deleting the DnsObject. Open to other suggestions to handle deleting the other parent fk.

src/registrar/tests/test_models_domain.py
```diff
@@ -3316,6 +3316,61 @@ def test_analyst_deletes_domain_with_ds_data(self):
...
+        self.assertFalse(DnsAccount_VendorDnsAccount.objects.filter(dns_account_id=account_id).exists())
+        self.assertFalse(VendorDnsAccount.objects.filter(id=vendor_account_id).exists())
+        self.assertFalse(DnsZone.objects.filter(domain=domain).exists())
+        self.assertFalse(DnsZone_VendorDnsZone.objects.filter(dns_zone_id=zone_id).exists())
+        self.assertFalse(VendorDnsZone.objects.filter(id=vendor_zone_id).exists())
+        self.assertFalse(DnsRecord.objects.filter(dns_zone_id=zone_id).exists())
+        self.assertFalse(DnsRecord_VendorDnsRecord.objects.filter(dns_record_id=record_id).exists())
+        self.assertFalse(VendorDnsRecord.objects.filter(id=vendor_record_id).exists())
+
+        # reset to avoid test pollution
```
erinysong, line 3388: I noticed this line was getting used to reset our hosts in other unit tests where we remove the hosts (necessary to set up a domain being eligible to delete in EPP). I'm not too familiar with domain hosts or test suites involving them - is this approach better to do individually by unit test when necessary, or would it be worth considering resetting our host data in setup/cleanup?
```
# Mock the InfoDomain command data to return a domain with no hosts
        # This is needed to simulate the domain being able to be deleted
        self.mockDataInfoDomain.hosts = []
```

src/registrar/models/domain.py
```diff
@@ -1192,6 +1192,8 @@ def _delete_domain(self):  # noqa
 
         self._delete_dnssecdata()
 
+        self._delete_db_dns_data()
+
```
abe-alam-ecs: Out of curiosity, why do we delete nameservers, hosts, dnssec data before we check if the domain itself can be deleted?

Shouldn't we check that first, and preserve the data if the domain isn't supposed to deleted?

src/registrar/models/domain.py
abe-alam-ecs: I don't think Django provides a built-in way to solve this problem, so I like this "manual" approach. I think it's clear, and I don't know that there's another way to do this other than implemented a helper that would just obfuscate what we're trying to do here.

kimallen: I did a little research and looks like there's a way to do this on the db level. I think that would be preferred over doing it manually. Basically it would mean setting up a `post_delete` signal. (There's one set up already for cleaning up domain invitations in `signals.py`). Basically it would be "listening" for delete of a record in the join table, and if there was one, then it would delete the 3rd table record (vendor_dns_*). I'll put more in the thread that you linked.

src/registrar/tests/test_models_domain.py
abe-alam-ecs: I've gone back and forth on this question too. I broke a lot of tests when I was going through and editing tests with hosts, which is why the structure remains as it is.

I think a cleanup effort would be nice, but I also think that individually declaring them by unit test is the way to go (at least, for now). The cleanup involved in a refactor is a 3 point effort at minimum, but it would be nice to reduce some of the complexity. If there's a strong opinion one way or the other, that's the way to go. Since Kim is doing a review, I'll defer to her judgment

abe-alam-ecs (commented):
I posted a couple of thoughts and questions that need to be addressed before approval of this PR.

Given that those thoughts may change the code, I thought it best to leave a comment rather than granting an explicit approval. The changes as they exist are appropriate, and I will approve as soon as I have confirmation about the ordering of deleting nameservers, dnssec data, etc before checking if `domain_can_be_deleted`

src/registrar/tests/test_models_domain.py
kimallen: Maybe @therealslimhsiehdy has some insight in regards to this being necessary related to domain deletion?

kimallen: And if it's needed and we are needing to do it in most of the tests in the test suite, I'd consider including it in the setup/cleanup. Though I agree with Abe that if it ends up being a large lift, you can do it individually for now and create a ticket for refactoring later.

kimallen (commented):
@erinysong I'm not opposed to doing the deletions manually, though would like if you could look into the way I mentioned in comment/thread, as I think it would be preferred if it did it automatically on the model level.

src/registrar/models/domain.py
allly-b: The _domain_can_be_deleted function really is checking if nameservers, etc have been deleted from epp first, and trying again to delete them if anything failed.  It returns true if the domain exists and has no hosts left. 

That said. THANK YOU for asking this question. The reason everything above this line is just doing epp calls but then deletes the database data below is as follows: **any** db transaction done above this line will take effect even if the EPP functions in _domain_can_be_deleted fail.

_domain_can_be_deleted _can_ throw a RegistryError, but we are _catching_ the error in [admin.py ](https://github.com/cisagov/manage.get.gov/blob/15f716dd5be312eabc004f4b8edb458fc38e6b0d/src/registrar/admin.py#L5429-L5430) which negates the atomic transaction that all requests go through. (see [this link ](https://docs.djangoproject.com/en/5.2/topics/db/transactions/#tying-transactions-to-http-requests) and scroll down to "avoid catching exceptions inside atomic" if you're curious).

TLDR: if this function succeeds and  _domain_can_be_deleted fails, the dns data _stays_ deleted while the domain is _not_ in a deleted state.

**Change request:** move _delete_db_dns_data() until after line 1214 or, even better, put it in _delete_related_objects_from_db (since delete_db_dns_data is also deleting related objects). 
If you don't mind, if you could surround line  [5427&5428 in admin.py](https://github.com/cisagov/manage.get.gov/blob/15f716dd5be312eabc004f4b8edb458fc38e6b0d/src/registrar/admin.py#L5427-L5429) in a `with atomic():` for sanities sake too that would be appreciate. (If not that can get thrown into another ticket).

src/registrar/models/domain.py
```diff
@@ -1275,6 +1277,55 @@ def _delete_dnssecdata(self):
...
+            logger.debug("Deleting DNS data for %s.", self.name)
+            try:
+                with transaction.atomic():
+                    dns_zone = DnsZone.objects.get(domain_id=self.id)
+                    logger.info("Removing db DNS records associated with %s.", self.name)
+                    records = DnsRecord.objects.filter(dns_zone=dns_zone)
+                    # Deleting DnsRecord cascade deletes associated DnsRecord_VendorDnsRecord.
+                    # Removes VendorDnsRecord associated with deleted DnsRecord_VendorDnsRecord
+                    for record in records:
+                        vendor_record = DnsRecord_VendorDnsRecord.objects.get(dns_record=record).vendor_dns_record
```
allly-b: **change request** If I'm reading this right, this model allows for duplicate dns_record values as long as only one of them is active. So 2+ values can be returned here which will result in an error. Also, this will error if 0 items are returned. While 0 is unlikely, it could happen if it's still in the middle of provisioning right? IE a race condition is occurring. At the very least account for the case of 2+ items, but ideally cover both scenarios.

src/registrar/models/domain.py
```diff
@@ -1275,6 +1277,55 @@ def _delete_dnssecdata(self):
...
+                        vendor_record.delete()
+                    records.delete()
+                    logger.info(
+                        "Removed db DNS records associated with zone for domain %s: %s.", self.name, str(records)
+                    )
+                    logger.info("Removing db DNS zone for domain %s.", self.name)
+                    vendor_zone = DnsZone_VendorDnsZone.objects.get(dns_zone=dns_zone).vendor_dns_zone
+                    dns_zone.delete()
+                    vendor_zone.delete()
+                    logger.info("Removed db DNS zone for domain %s.", self.name)
```
allly-b: **nitpick:** I think this was meant to be up 1 line.

src/registrar/tests/test_models_domain.py
```diff
@@ -3316,6 +3316,61 @@ def test_analyst_deletes_domain_with_ds_data(self):
...
+            DnsRecord_VendorDnsRecord,
+        )
+        from registrar.tests.helpers.dns_data_generator import create_initial_dns_setup, create_dns_record
+
+        _, dns_account, dns_zone = create_initial_dns_setup(domain=domain)
+        dns_record = create_dns_record(dns_zone)
+        account_id, zone_id, record_id = dns_account.id, dns_zone.id, dns_record.id
+
+        vendor_account_id = DnsAccount_VendorDnsAccount.objects.get(dns_account=dns_account).vendor_dns_account.id
+        vendor_zone_id = DnsZone_VendorDnsZone.objects.get(dns_zone=dns_zone).id
```
allly-b: **change before merging**
```suggestion
        vendor_zone_id = DnsZone_VendorDnsZone.objects.get(dns_zone=dns_zone).vendor_dns_zone.id
```

src/registrar/tests/test_models_domain.py
```diff
@@ -3316,6 +3316,61 @@ def test_analyst_deletes_domain_with_ds_data(self):
...
+        vendor_record_id = DnsRecord_VendorDnsRecord.objects.get(dns_record_id=record_id).id
```
allly-b: **change before merging:** I'm pretty sure this is just working because all their ids are 1 because they are the only dns objects in this test class at the moment
```suggestion
        vendor_record_id = DnsRecord_VendorDnsRecord.objects.get(dns_record_id=record_id).vendor_dns_record.id
```

src/registrar/models/domain.py
```diff
@@ -1275,6 +1277,55 @@ def _delete_dnssecdata(self):
...
+                    logger.info("Removing db DNS records associated with %s.", self.name)
+                    records = DnsRecord.objects.filter(dns_zone=dns_zone)
+                    # Deleting DnsRecord cascade deletes associated DnsRecord_VendorDnsRecord.
+                    # Removes VendorDnsRecord associated with deleted DnsRecord_VendorDnsRecord
+                    for record in records:
+                        vendor_record = DnsRecord_VendorDnsRecord.objects.get(dns_record=record).vendor_dns_record
+                        vendor_record.delete()
+                    records.delete()
+                    logger.info(
+                        "Removed db DNS records associated with zone for domain %s: %s.", self.name, str(records)
```
allly-b: **Nitpick:**
```suggestion
                        "Removed db DNS records associated with zone for domain %s", self.name
```

records is deleted, so it should print an empty list here. You could also move this above the delete call and say "about to remove"

src/registrar/models/domain.py
```diff
@@ -1275,6 +1277,55 @@ def _delete_dnssecdata(self):
...
+                    logger.info("Removed db DNS zone for domain %s.", self.name)
+                    logger.info("Removing db DNS accounts for %s.", self.name)
+                    dns_account = dns_zone.dns_account
+                    vendor_account = DnsAccount_VendorDnsAccount.objects.get(dns_account=dns_account).vendor_dns_account
+                    dns_account.delete()
+                    vendor_account.delete()
+                    logger.info("Removed db DNS account for domain %s.", self.name)
+
+            except Exception as e:
+                logger.error("Error deleting DNS data for %s: %s", self.name, e)
```
allly-b:
```suggestion
                logger.error("Error deleting DNS data for %s: %s", self.name, e, exc_info=True)
```

**Nitpick:** Debugging why a deletion failed sucks, I would make sure the traceback logs correctly

src/registrar/tests/test_models_domain.py
```diff
@@ -3316,6 +3316,61 @@ def test_analyst_deletes_domain_with_ds_data(self):
         # reset to avoid test pollution
         self.mockDataInfoDomain.hosts = ["fake.host.com", "fake2.host.com"]
 
+    def test_delete_domain_in_epp_deletes_db_dns_data(self):
```
allly-b, line 3336: **nitpick: ** Add @less_console_noise_decorator

src/registrar/tests/test_models_domain.py
```diff
@@ -3316,6 +3316,61 @@ def test_analyst_deletes_domain_with_ds_data(self):
...
+        """
+        Deleting a domain from EPP removes the domain's DNS data from the database.
+        """
+        # Create a domain with DS data
```
allly-b: **Nitpick:** copy and paste error I bet.
```suggestion
        # Create a domain with DNS data
```

src/registrar/models/domain.py
allly-b: I reviewed the manual vs signal/post_delete approach, and I vote in favor of manual as well. 

- this function (deleteInEpp) doesn't _delete_ the domain object, it just changes state to be `DELETED`. This means you couldn't trigger a cascading delete by this state change, and you couldn't have a post_delete be triggered by it. 
- You _could_ keep the  DnsRecord manual and then have signals for deleting the models with the `Vendor` prefix. For instance, this code would delete DnsAccount_VendorDnsAccount, but then a post_delete signal could then handle actually deleting `VendorDnsAccount`. You would repeat this for DnsZone_VendorDnsZone and DnsRecord_VendorDnsRecord. So **three** signals, and it only saves a handful of lines of code. 
- signals run on every test globally, can be difficult to test for corner cases, are even more annoying to debug.  and you have to avoid deleting using the name of the foreign key field and use id field for deletion instead to avoid errors. This can be easy to miss and hard to catch that something went wrong.
- dns_record.py looks like it's already deleting DNSRecord and VendorDNSRecord. So adding a `post_delete` on DnsRecord_VendorDnsRecord means VendorDnsRecord gets deleted twice. (Once by the receiver, once by the existing vendor_dns_record.delete() line in DnsRecord.delete_by_x_record_id()) Something there would need to change.

allly-b (approved):
Before merging please make the changes requested. Nitpicks, etc are optional.

src/registrar/tests/test_models_domain.py
erinysong: hm yeah agree with what people have said here that it'd be better to handle more efficient unit test host handling separately

src/registrar/models/domain.py
erinysong: good point on the order we do things here and deleting DNS data from the database only after we confirm we can delete the domain! Just moved our db DNS cleanup in `_delete_related_objects_from_db` and wrapped our domain deletion (those 2 lines) in an atomic transaction. Great suggestions, thank you both!

src/registrar/models/domain.py
erinysong: I'll make it clearer that it's deleting our data and not just the zone - thank you for the catch!

src/registrar/tests/test_models_domain.py
erinysong: thank you, good catch! Just updated this and the original code I copied from

src/registrar/models/domain.py
erinysong: just updated this so that we filter for any vendor_records that might still exist/not exist - does this match what you were talking about above? Technically the model DnsRecord_VendorDnsRecord's dns_record field should have only have unique values but it doesn't hurt to filter for anything that may remain that we don't want. Let me know if this aligns with what you were intending!

src/registrar/models/domain.py
```diff
@@ -1354,6 +1352,12 @@ def _delete_related_objects_from_db(self):
                 c.delete()
         except Exception as e:
             logger.error("Error deleting contacts for domain %s: %s", self.name, str(e))
+        
+        logger.info("Deleteing DNS Data")
```
abe-alam-ecs: Typo here!

src/registrar/models/domain.py
abe-alam-ecs: Yeah, I think Erin's right in that there should only ever be unique values. The updates can't hurt, though!

abe-alam-ecs (approved):
All requested changes were made as Alysia requested. I think this looks good. 

As a note, I was unable to successfully delete a domain locally. I think this is because of EPP registry syncs based on the failure logs, so I don't think this is anything to worry about. The test still executed correctly when I ran that.

Approved.

src/registrar/models/domain.py
erinysong: good catch! Just corrected, thanks Abe!

PR #5229, opened 2026-07-23
#5131 - Run script to create missing registrant contacts [sm]

### LINKED ISSUE #5131: Run script to create missing registrant contacts
### Issue description

Now that we have a script to create missing registrant contacts, we need to run it.

### Acceptance criteria

- [x] Meet with Product to go through what the new script does/doesn't do
- [x] Run the script from #4982 on staging
- [x] 🛑  Check in with Product to check the registrant contacts on staging (and optionally OT&E if relevant)
- [ ] Run the script on Stable
- [ ] Any additional tasks based on what the script does and doesn't do (for example, if it doesn't make updates to OT&E and the original script for updating registrant data needs to be run, we should do that!)

### Additional context

_No response_

### Links to other issues

- 🔄  Related to #4982 
- 🔄  Related to #3618

smishraECS:
## Ticket

Resolves #5131 

## Changes

Updated logic to call epplib to link new registrant to the domain

## Context for reviewers
This code is needed due to an edge case which causes the addRegistrant() method to fail to update the registry domain with the new Registrant contact due to the registry id values in the registrar not matching the registry. 

Currently, the registry Registrant value is the same as the Admin contact and the existing logic requires the ids to be the same. Should only be required for this one time run as after this the contact info between the registrar and registry will be synced.

## Setup

To test this with your local/sandbox you will need to seed some Public Contact data. Reach out and I can provide the fixtures script used to populate the data if needed.

Make sure you have some domains created without Registrant Contacts. This may require manipulating existing contact sets (adding/deleting)

Run the management command using: docker-compose exec app python ./manage.py update_missing_registrant_contacts

Run test cases using: docker-compose exec app python ./manage.py test registrar.tests.test_update_missing_registrant_contacts

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons
  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References


src/registrar/management/commands/update_missing_registrant_contacts.py
```diff
@@ -66,7 +66,16 @@ def handle(self, *args, **options):
...
-                        domain.addRegistrant()
+                        # Create the registrant in the registrar DB
+                        registry_id = domain.addRegistrant()
+
+                        # This is needed because currently, the Admin contact is listed as the registrant in CloudFlare
+                        # so addRegistrant() fails to update the value since the ids don't match.
+                        registrant = PublicContact.objects.filter(registry_id=registry_id)
+
+                        logger.info(f"Updating registry Registrant Public Contact {registry_id} for {domain.name}")
+                        domain.update_domain_registrant_in_registry(registrant)
```
allly-b: **Blocking:** registrant is a queryset  not a public contact, so this will fail for every contact it's called on. Easy fix though.

src/registrar/management/commands/update_missing_registrant_contacts.py
```diff
@@ -66,7 +66,16 @@ def handle(self, *args, **options):
...
+
                         add_count += 1
                     except Exception as e:
```
allly-b, line 84: **Change requested:** The error above is being swallowed by this exception, you need to re-raise the error here so that your summary output can show the error appropriately.

src/registrar/management/commands/update_missing_registrant_contacts.py
```diff
@@ -66,7 +66,16 @@ def handle(self, *args, **options):
                 else:
                     logger.info(f"Creating Registrant Public Contact for {domain.name}")
                     try:
-                        domain.addRegistrant()
+                        # Create the registrant in the registrar DB
+                        registry_id = domain.addRegistrant()
+
+                        # This is needed because currently, the Admin contact is listed as the registrant in CloudFlare
```
allly-b: **nitpick/non-blocking** ```suggestion
                        # This is needed because currently, the Admin contact is listed as the registrant in the Registry
```

If vendors change in the future, it's less of a headache to limit when we use the vendors name.

File: src/registrar/management/commands/update_missing_registrant_contacts.py [hunk 4]
```diff
@@ -66,7 +66,16 @@ def handle(self, *args, **options):
                 else:
                     logger.info(f"Creating Registrant Public Contact for {domain.name}")
                     try:
-                        domain.addRegistrant()
+                        # Create the registrant in the registrar DB
```
allly-b:
```suggestion
                        # Create the registrant in the registrar DB and the registry
```

src/registrar/models/domain.py
```diff
@@ -2680,3 +2680,11 @@ def delete_with_no_dns(self, *args, **kwargs):
         # Delete domain
         self.deleteInEpp()
         self.save()
+
+    def update_domain_registrant_in_registry(self, contact: PublicContact):
```
allly-b: **Blocking question:** Maybe I'm missing something here, so I want to clarify. AddRegistrant on domain, 1.) creates the public contact object. 2.) it calls save 3.) saving a public contact adds it to the domain. 

So given you already called addRegistrant  before you use this function, what is the purpose of this function? I'm reading it as it's just going to update the domain with the same contact that was just added the domain.

src/registrar/management/commands/update_missing_registrant_contacts.py
```diff
@@ -66,7 +66,16 @@ def handle(self, *args, **options):
...
+                        registry_id = domain.addRegistrant()
```
allly-b, line 69: **blocking question:** Did anyone check if the missing registrant ids are actually missing from the registry? It could be in the registry and not in the registrar, and in which case we could just pull what the registry has without having to create a new object. Then only create a new object if it's missing.

src/registrar/management/commands/update_missing_registrant_contacts.py
allly-b:
```suggestion
                        domain._add_registrant_to_existing_domain( registrant)
```
when registrant is a contact object you can just call the add registrant function. Was there some other reason for creating a new function in domain.py?

src/registrar/tests/test_update_missing_registrant_contacts.py
```diff
@@ -72,11 +72,13 @@ def test_command_update_missing_registrant_contacts_dry_run(self):
             self.assertEqual(Domain.objects.all().count(), 3)
             self.assertEqual(update_mock.call_count, 0)
 
-    def test_command_update_missing_registrant_contacts_no_dry_run(self):
-        with patch("registrar.models.domain.Domain.addRegistrant") as update_mock:
-            call_command("update_missing_registrant_contacts", dry_run=False)
-            self.assertEqual(Domain.objects.all().count(), 3)
-            self.assertEqual(update_mock.call_count, 2)
+    @patch("registrar.models.domain.Domain.addRegistrant")
+    @patch("registrar.models.domain.Domain.update_domain_registrant_in_registry")
```
allly-b: **Change requested:** Maybe a typo here? You're mocking the function you added which  helped to hide the errors.  I would heavily suggest just mocking the actually epp calls sent though I know that's a couple more steps. At the very lease something testing the meat of the logic here would be nice

Also, the patches are reversed: patches are passed to a function in bottom up order. So update_domain_registrant_in_registry is being mapped to mock_add_registrants and addRegistrant is being mapped to mock_update_epp

**suggestion:/ non mandatory **Also, I would suggest having one test for the passing case and one test for fail case. And in general on unit tests have a comment for any numbers used in tests as if they start failing it's a pain to try to understand why 2 or 3 worked before. A simple "2 calls to create public contact" in a comment helps.

allly-b (changes requested):
See comments. Haven't ran it live, found a couple small things. Ping when ready for re-review

src/registrar/models/domain.py
smishraECS: So the issue that required these changes is that when we ran the initial script (just to add the registrant) we noticed the script was creating the contact in the registry but not linking it to the domain. After some sleuthing, I realized that this was because the addRegistrant method only works if the registrant contact is not set in the registry (which in our case the registrant contact is set, its set to the same as the Admin contact). 

That's why we have to force update the contact info, since the registrant contacts in the registrar and registry have different IDs

src/registrar/management/commands/update_missing_registrant_contacts.py
smishraECS: These domains were missing registrant contacts in the registrar as I believe they were legacy domains, were creating new registrant contacts for them in the registrar.

src/registrar/management/commands/update_missing_registrant_contacts.py
smishraECS: Decided to create a new method for it since _add_registrant_to_existing_domain is an internal method

src/registrar/management/commands/update_missing_registrant_contacts.py
allly-b: But why not pull in the registry value and just use that on the public contact? Such as by fetching all the contacts from the registry, pulling out the registrant and creating a public contact on our side that matches whatever the registry returns? 

Also, removing the block on the question. For a one off script I can see how it may just be easier to create them new if that's the reasoning.

smishraECS: The registry ID is currently the same as the admin ID. Doing so would cause a duplicate ID to be created in the registrar.

allly-b (dismissed):
**Change request:** The update_missing_registrant_contacts script is not filtering the domains and is instead doing Domain.objects.all(). This means it's including domains in a deleted, onhold, and unknown state. Deleted and unknown domains _should_ be missing all their public contacts. Deleted because it's been deleted from the registry, and Unknown because it has yet to be added to the registry. At the very least these two need to be filtered out.

For on hold, the domain is on hold usually because it's about to be deleted but I suppose it's fine to add a contact to these. The one thing I would keep in mind is if a domain has the status "clientUpdateProhibited" while on hold then the code will error as the registry will block changes. I think we only set _clienthold_ when they are on hold, but it may also be worth a sanity check by finding a real onhold domain and verifying what statuses it has.

src/registrar/tests/test_update_missing_registrant_contacts.py
```diff
@@ -69,14 +69,18 @@ def setUp(self):
...
             self.assertEqual(Domain.objects.all().count(), 3)
+            # Dry run so none should be added
             self.assertEqual(update_mock.call_count, 0)
 
-    def test_command_update_missing_registrant_contacts_no_dry_run(self):
-        with patch("registrar.models.domain.Domain.addRegistrant") as update_mock:
-            call_command("update_missing_registrant_contacts", dry_run=False)
-            self.assertEqual(Domain.objects.all().count(), 3)
-            self.assertEqual(update_mock.call_count, 2)
+    @patch("registrar.models.domain.Domain.addRegistrant")
```
lizpearl, line 77: I would update this to patch the epplib call because that was the part that wasn't getting called in the previous version of the script.

lizpearl (changes requested):
Change request: Add an option to run this script for a single domain.  When deploying, we'll want to test a few single domains before we run the entire script.

src/registrar/management/commands/update_missing_registrant_contacts.py
```diff
@@ -30,14 +30,25 @@ def add_arguments(self, parser):
...
         dry_run = bool(options.get("dry_run", True))
+        target_domain = options.get("target_domain", None)
+
         # Get all contacts
         all_contacts = PublicContact.objects.all()
-        # Get all domains
-        all_domains = Domain.objects.all()
+        # Get domains
+        if target_domain:
+            domains_list = Domain.objects.filter(name=target_domain)
```
lizpearl: Nice to have for deployment: only allow target domains that are in ready or dns needed.

src/registrar/management/commands/update_missing_registrant_contacts.py
```diff
@@ -66,8 +73,23 @@ def handle(self, *args, **options):
...
                     logger.info(f"Creating Registrant Public Contact for {domain.name}")
                     try:
-                        domain.addRegistrant()
-                        add_count += 1
+                        # Create the registrant in the registrar DB and registry
+                        registry_id = domain.addRegistrant()
+
+                        # This is needed because currently, the Admin contact is listed as the registrant in CloudFlare
+                        # If vendors change in the future, it's less of a headache to limit when we use the vendor name.
+                        registrant = PublicContact.objects.filter(registry_id=registry_id).first()
```
lizpearl: Question: I'm not understanding this comment.  Can a registry have multiple public contacts with the same id?

src/registrar/tests/test_update_missing_registrant_contacts.py
```diff
@@ -69,54 +69,19 @@ def setUp(self):
...
-        self.contact_three.registry_id = "contact"
-        self.contact_three.domain = self.domain_three
-
-        self.contact_three.save(skip_epp_save=True)
-
-        update_count = call_command("update_missing_registrant_contacts", dry_run=True)
-
-        self.assertEqual(update_count, 0)
+    @patch("registrar.models.domain.Domain.addRegistrant")
+    @patch("registrar.models.domain.Domain._add_registrant_to_existing_domain")
```
lizpearl, line 78: non-blocking/comment: When we were referring to mocking epplib functions, we were referring to [this line](https://github.com/cisagov/manage.get.gov/blob/35b733105098ae1b956e72cb907be167d697a4a3/src/registrar/models/domain.py#L965).  Generally, they are functions imported from this: https://github.com/cisagov/manage.get.gov/tree/35b733105098ae1b956e72cb907be167d697a4a3/src/epplibwrapper

lizpearl (approved):
Just a few comments but nothing blocking.  I'll approve so it doesn't block.  I tested the following:
- Ran the script for a single domain in my sandbox and verified that the contact was added both in the registrar and the OT&E registry
- Ran the dry run version of the script and
   - did a spot check that the listed domains did not have a registrant in the registry
   - Did a query to count all ready and dns_needed domains and then another to count all ready and dns needed domains with a registrant. Compared to the number that the dry run output and made sure the math made sense.

src/registrar/management/commands/update_missing_registrant_contacts.py
smishraECS: Yes, it looks like a one-to-many relationship within CloudFlare. Currently, the same ID serves as the Admin Contact as well as the Registrant Contact

src/registrar/management/commands/update_missing_registrant_contacts.py
smishraECS: Oh good idea!

src/registrar/management/commands/update_missing_registrant_contacts.py
lizpearl: Looked it up!  It looks like Public contact ids [should be unique in the registry](https://datatracker.ietf.org/doc/html/rfc5731#section-2.2). The registry_ids don't have to be unique in the registrar  - only with domain_id, registrar_id and contact type.

I still don't understand the comment about the vendor?

src/registrar/management/commands/update_missing_registrant_contacts.py
```diff
@@ -66,8 +73,23 @@ def handle(self, *args, **options):
...
+
+                        logger.info(f"Updating registry Registrant Public Contact {registry_id} for {domain.name}")
+                        try:
+                            # This is a one off script, makes more sense to use the internal method than create
+                            # a new public access method which we need to maintain.
+                            domain._add_registrant_to_existing_domain(registrant)
```
chaswick: This method was patched out for the test.  It can quietly fail if there is an issue with the registry (domain.py:976).  But if it does, the registrant was already added in this script on line 77. I think the right approach in this case to avoid dealing with the issue in domain.py is

- Make sure we process each domain regardless of whether its in registrant_domain_set
- inside the try: before domain.addRegistrant() check that registrant is None (in case addRegistrant was already run on it)
- Whether it already existed or was created after that check, always call _add_registrant_to_existing_domain() to finish the link

chaswick (changes requested):

src/registrar/management/commands/update_missing_registrant_contacts.py
smishraECS: Reworded it

src/registrar/tests/test_update_missing_registrant_contacts.py
lizpearl: I need to make a correction, I linked to the functions that create the message but not the actual send.  If you were to mock, probably do it for the send.

src/registrar/management/commands/update_missing_registrant_contacts.py
```diff
@@ -30,47 +30,62 @@ def add_arguments(self, parser):
...
+                    if not PublicContact.objects.filter(domain=domain).first():
+                        registry_id = domain.addRegistrant()
+                    else:
+                        registry_id = domain.registrant_contact.registry_id
+
+                    # This is needed because currently, the Admin contact is listed as the registrant in CloudFlare
+                    # and the addRegistrant method requires the Registrant contact to be blank in Cloudflare to
+                    # update it. Due to this, we use _add_registrant_to_existing_domain to force update it.
+                    registrant = PublicContact.objects.filter(registry_id=registry_id).first()
+                    logger.info(f"Updating registry Registrant Public Contact {registry_id} for {domain.name}")
```
chaswick: I think we need to filter on the type=REGISTRANT for the first, and we need to use that registrant in case there was a previously failed link (meaning there could be an unlinked registrant to the domain).
```
                    registrant = PublicContact.objects.filter(
                        domain=domain,
                        contact_type=PublicContact.ContactTypeChoices.REGISTRANT,
                    ).first()
                    if registrant is None:
                        registry_id = domain.addRegistrant()
                        registrant = PublicContact.objects.get(
                            domain=domain,
                            contact_type=PublicContact.ContactTypeChoices.REGISTRANT,
                            registry_id=registry_id,
                        )

                    # This is needed because currently, the Admin contact is listed as the registrant in CloudFlare
                    # and the addRegistrant method requires the Registrant contact to be blank in Cloudflare to
                    # update it. Due to this, we use _add_registrant_to_existing_domain to force update it.
                    logger.info(
                        f"Updating registry Registrant Public Contact {registrant.registry_id} for {domain.name}"
                    )
```

chaswick (changes requested):
Suggested code

src/registrar/management/commands/update_missing_registrant_contacts.py
```diff
@@ -30,47 +30,63 @@ def add_arguments(self, parser):
...
+                        registry_id = domain.addRegistrant()
+                        registrant = PublicContact.objects.filter(
+                            domain=domain,
+                            registry_id=registry_id,
+                            contact_type=PublicContact.ContactTypeChoices.REGISTRANT,
+                        ).first()
+                    # This is needed because currently, the Admin contact is listed as the registrant in CloudFlare
+                    # and the addRegistrant method requires the Registrant contact to be blank in Cloudflare to
+                    # update it. Due to this, we use _add_registrant_to_existing_domain to force update it.
+                    logger.info(f"Updating registry Registrant Public Contact {registry_id} for {domain.name}")
```
chaswick:
```suggestion
logger.info(f"Updating registry Registrant Public Contact {registrant.registry_id} for {domain.name}")
```

chaswick (approved):

PR #5322, opened 2026-08-20
#5257: Rdap endpoint validation - [none]

### LINKED ISSUE #5257: Rdap endpoint should validate input
### Issue description

The RDAP endpoint (`/api/v1/rdap/`) should validate input.

See [an example discussed here](https://cisa-corp.slack.com/archives/C068GJKRGCV/p1784152883142849)

### Acceptance criteria

- [ ] Validate input parameters for the rdap endpoint
   - The [availability endpoint](https://github.com/cisagov/manage.get.gov/blob/bd48379eb5ea01062bcc2b0b121ebadfd7f6f669/src/api/views.py#L57) does this and is a good reference.

### Additional context

_No response_

### Links to other issues

_No response_

allly-b:
## Ticket

Resolves #5257 

## Changes

- The http request that actually performs the rdap call is moved out into it's own function with caching applied to it  (see context)
- rdap view function now validates the "domain" parameter passed to it, and returns an error message on invalid data.
- caching now works on the rdap function
- set a max cache limit

## Context for reviewers

The print statement is there just to help with review to show the caching works and so you can also see the value is getting cleaned, such as trimmed of whitespaces etc. It **will** be deleted before merging

**Why caching didn't work:**
The rdap function takes parameters "request" and "domain" and the `ttl_cache` decorator caches based off parameters passed to the function. This means it was caching based on **both** request and domain, but you won't get repeat requests. Even if you make two back to back curls from the same machine it will count as a different request, so caching basically was doing nothing.

**How it's fixed:**
Moving the `ttl_cache` decorator to the get_rdap function which takes only one parameter, `domain`, allows it to cache based on domain names passed.

**max cache size limit**
Added as this is a public endpoint. without a max size repeated rdap calls could grow the cache quite a bit. However, the cache is _per worker_, and each of our workers only accepts a max of 1k requests, before it restarts. This means the cache would already reset after 1k requests were received by the worker, so an argument could be made that a max size doesn't gain us much. But figured it didn't hurt to add in. Open to opinions
## Setup

1. deploy code to your sandbox
2. in terminal run `curl "<yoursandboxpath>/api/v1/rdap/?domain=example.gov"`
3. repeat the command above and see that it doesn't trigger the print in the cloud.gov logs (this shows caching is working)
4. now change example.gov to be an invalid domain name to double check the validation is working correctly. Example: empty string, enter a series of whitespace characters, enter characters that aren't . or -, enter a .com address or multiple periods, etc.

Remember when testing this, it's a parameter being passed to a url so spaces and characters need to be _percent encoded_. For example if you want to simulate a user adding a space such as "my example.gov", a space in percent encoding is %20 so you would use the url: <yoursandboxpath>/api/v1/rdap/?domain=my_%20_example.gov. Here's a [free online url encoder ](https://www.url-encode-decode.com/)if you don't want to google special character values indivdually

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons
  - [ ] Chrome
  - [ ] Microsoft Edge
  - [ ] FireFox
  - [ ] Safari

### References


src/api/views.py
```diff
@@ -70,21 +73,45 @@ def available(request, domain=""):
...
+@ttl_cache(maxsize=128, ttl=600)
+def get_rdap_data(domain):
+    """Fetch RDAP data for a domain from the Cloudflare API.
+    Used by the /api/v1/rdap endpoint; separated out
+    so that caching works properly.
+    Domain parameter is cleaned upstream in the rdap view function.
+    Returns a JSON dictionary of the RDAP data.
+    """
+    print(f"Will delete before merging, request for domain: {domain}")
+    return requests.get(RDAP_URL.format(domain=quote(domain, safe="")), timeout=5).json()
```
therealslimhsiehdy, line 87: **Question:** Thoughts on putting in into a try/except block to catch any errors that arise (this is me assuming that there could be a status_code error coming from the .get)?

src/api/views.py
```diff
@@ -70,21 +73,45 @@ def available(request, domain=""):
     return json_response
 
 
+# Since we cache domain RDAP data, cache time (ttl) may need to be re-evaluated this if we encounter any memory issues
```
therealslimhsiehdy: **NIT:** remove the `this` (so it'd be `re-evalluated if we encounter`)

src/api/views.py
```diff
@@ -70,21 +73,45 @@ def available(request, domain=""):
...
+# maxsize is the number of domains to cache, oldest accessed domain names are removed when maxsize is reached.
+@ttl_cache(maxsize=128, ttl=600)
```
therealslimhsiehdy, line 78: **Question:** Why did we choose 128 here?

src/api/tests/test_rdap.py
```diff
@@ -31,22 +33,34 @@ def test_rdap_get_no_tld(self):
...
         request = self.factory.get(API_BASE_PATH + "whitehouse.com")
         request.user = self.user
         response = rdap(request, domain="whitehouse.com")
 
-        self.assertContains(response, "errorCode")
+        self.assertContains(response, "errorCode", status_code=400)
         response_object = json.loads(response.content)
         self.assertIn("errorCode", response_object)
 
+    @patch("registrar.models.utility.domain_helper.check_domain_available")
```
therealslimhsiehdy, line 46: **Question:** Would it be possible to also write a unit test that mocks `requests.get` and calls `rdap` twice with the same domain + asserts that it was only hit once? (Not sure how painful it is to do it, sorry)

src/api/tests/test_rdap.py
```diff
@@ -18,6 +19,7 @@ def setUp(self):
         super().setUp()
         self.user = get_user_model().objects.create(username="username")
         self.factory = RequestFactory()
+        get_rdap_data.cache_clear()
```
therealslimhsiehdy, line 22: **Question:** Sorry dumb question, just to clarify - we add `cache_clear` in so we clear out the cache so there is no overwriting for the unit tests, correct?

therealslimhsiehdy (commented):
Left some comments!

Thank you to Alysia for clarifying + help - she had pushed just a print statement onto `potato` so I could see the before - will push to my own sandbox the changes to test it!

~~Also had a few extra questions when I tested it on `potato`:~~
~~1. I tested `curl "https://getgov-potato.app.cloud.gov/api/v1/rdap/?domain=meowardisthebest.gov"` twice, but saw `RDAP request for domain:  meowardisthebest.gov` twice - just to confirm, we should only see it once if we test in succession, right?~~

~~2. Also to test an invalid domain I did `curl "https://getgov-potato.app.cloud.gov/api/v1/rdap/?domain=351#%.gov"`, but still saw this `15:52:04.282: [APP/PROC/WEB.0] RDAP request for domain:  351` -- what error is usually expected for an invalid domain?~~

src/api/views.py
allly-b: So this is a cached function, which means it won't cache if the value if an exception is raised. If a try and except is placed here then that response will be cached. The ideal procedure with caches is to not cache exceptions on requests because a request.get can throw errors if the third party vendor goes offline, is too slow to response, if there is network interference etc. All of those could resolve within a few seconds, hence caching may return inaccurate data by showing the endpoint is still failing when really it may not be anymore.

src/api/views.py
allly-b: TBH, I was following[ this article ](https://www.dataleadsfuture.com/implement-a-cache-decorator-with-time-to-live-feature-in-python/) in regards to picking the size

src/api/tests/test_rdap.py
allly-b: Yeah I went with not making a caching unit test since we didn't have one when this was first implemented and the caching issue was only found as a side effect of working on this ticket, it wasn't a main AC. So I would vote not to given the size of the ticket.

src/api/tests/test_rdap.py
allly-b: Yes, without clearing the cache between tests some tests could pass just by accident due to an old value getting cached.

therealslimhsiehdy (approved):
LGTM, and when I did `curl "https://getgov-rh.app.cloud.gov/api/v1/rdap/?domain=3415234%20.gov"` I received 
`{"errorCode": 400, "title": "Invalid domain", "description": ["Enter a domain using only letters, numbers, or hyphens (though we don't recommend using hyphens)."]}%`

Image

chaswick (approved):

PR #5325, opened 2026-08-24
Kma/5045 mitigate console errors

### LINKED ISSUE #5045: Clean up console errors on the dns records page
### Current Behavior

Observed console errors on first visit to the page as well as on clicking record type and other basic actions. 

RecrodType change:
Image

Example of the CSP error on form load due to the django htmx inline styling:
`Applying inline style violates the following Content Security Policy directive 'style-src 'self' https://www.ssa.gov/accessibility/andi/andi.css'. Either the 'unsafe-inline' keyword, a hash ('sha256-faU7yAF8NxuMTNEwVmBz+VcYeIoBQ2EMHW3WaVxCvnk='), or a nonce ('nonce-...') is required to enable inline execution. The action has been blocked.`

### Expected Behavior

On first visit to the page as well as on clicking record type and all available actions to-date on the page there should be no console errors.

### Steps to Reproduce

1. Visit the DNS records page
2. Notice the errors in the console
3. Click Add record and select a record type
4. Notice new errors

Try several operations on the dns records page to ensure any errors are surfaced

### Environment

development

### Additional Context

We set a Content Security Policy that says, roughly, "only load styles from my own domain `('self')`" It blocks _inline_ `<style> `blocks by default — because injected inline styles/scripts are a classic way attackers sneak code into a page (XSS). So when htmx tries to drop that inline `<style>` in, the browser refuses and logs the error. 

Suggested options for **fixing the CSP**:

1. a config meta tag in base.html
`<meta name="htmx-config" content='{"includeIndicatorStyles": false}'>`
2. A CSP nonce

Suggested **fix for the record type drop down** in domain_dns_record_form_content.html line10:
` x-on:change="recordType = ($event.target && $event.target.id === 'id_type') ? $event.target.value : recordType"`

kimallen:
## Ticket

Resolves #5045 

## Changes

- Adds a `meta` tag to base.html to stop console error
- Moves `on:change` to the select so that it does not get flagged by Alpine CSP rule (it can't read block js, just simple code)

## Context for reviewers

I'm not sure why all the whitespace changes happened- I recommend reviewing with whitespace changes removed.
image

## Setup

Use a sandbox setup for dns hosting (can use kma)
Enroll a domain in /admin (might be easiest to approve a domain request and use that one, as there's some data misalignment with some existing domains)

Open the browser and dev tools console, then open the app --> should be no console errors
Go to a dns enrolled domain and DNS Records page --> check that there's no console errors
Keep an eye on the console while Adding a record, Editing a record. Try making each of the record types. There should be no console errors.

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons

  - [ ] Page header structure
  - [ ] Landmarks
  - [ ] Links and buttons


src/registrar/templates/base.html
```diff
@@ -64,6 +64,7 @@
 
   {% block feeds %}{% endblock %}
   {% block extra_head %}{% endblock %}
+  <meta name="htmx-config" content='{"includeIndicatorStyles": false}'>
```
therealslimhsiehdy: **Question/suggestion:** Would you mind adding a comment on why this is needed and/or what it does? I'm not as knowledgable in this, but I do have a concern of if this'll cause something else to silently fail? 😬

src/registrar/forms/domain.py
```diff
@@ -881,6 +881,7 @@ class Meta:
             attrs={
                 "class": "usa-select",
                 "required": "required",
+                "x-on:change": "recordType = $event.target.selectedIndex",
```
therealslimhsiehdy, line 884: **Question:** Why recordType specifically + how does this affect/hide the CSP errors?

therealslimhsiehdy (commented):
Tested on `aa` (errors) and then on `kma` (no errors), however I do have more questions on the code changes and how those helped resolve the CSP error 😅 

Image
Image

src/registrar/forms/domain.py
kimallen: The original offending code was set on `<section>`:
```
"if ($event.target && $event.target.id === 'id_type') { recordType = $event.target.selectedIndex }"
```
The issue is related to [how Alpine handles CSP](https://alpinejs.dev/advanced/csp). It uses utilities that violate our CSP, so we are using a CSP-friendly build of Alpine. That build can handle simple basic operations, but it can't handle the above block code. 

I tried modifying it like [this](https://github.com/cisagov/manage.get.gov/pull/5325/commits/dcb12ea9ff5980daa63468dfa48173f62abf055f), but that did not work. So, I was able to achieve the same functionality by moving the operation to the `<select>` element. Since it was already on the select element, I was able to drop the if statement.

Let me know if that answers your question!

src/registrar/templates/base.html
kimallen: That's a good question/concern! 

The error that was triggered is `Applying inline style violates the following Content Security Policy directive 'style-src 'self' ` The htmx function `includeIndicatorStyles` has inline styles in the source code, so unless `includeIndicatorStyles` is false, it will violate the CSP policy which doesn't allow inline styles. 
from htmx-2.js:
```
function insertIndicatorStyles() {
    if (htmx.config.includeIndicatorStyles !== false) {
      const nonceAttribute = htmx.config.inlineStyleNonce ? ` nonce="${htmx.config.inlineStyleNonce}"` : ''
      const indicator = htmx.config.indicatorClass
      const request = htmx.config.requestClass
      getDocument().head.insertAdjacentHTML('beforeend',
        `<style${nonceAttribute}>` +
        `.${indicator}{opacity:0;visibility: hidden} ` +
        `.${request} .${indicator}, .${request}.${indicator}{opacity:1;visibility: visible;transition: opacity 200ms ease-in}` +
        '</style>'
      )
    }
  }
```

We are not currently using the `htmx-indicator` in our htmx, so it won't break anything right now. But, there's a good chance with some of the upcoming design work that we will want to use it. So, I'm going to go ahead and move those styles to our own stylesheet so that it doesn't break if we do choose to use it in the future.

kimallen: I added a comment (and moved it next to other meta tags) and added the styles that are removed by disabling `includeIndicatorStyles` [6dbffba](https://github.com/cisagov/manage.get.gov/pull/5325/commits/6dbffba0940d526bfcc1a212521d5d6563e48100)

src/registrar/forms/domain.py
therealslimhsiehdy: This is SUPER helpful, thank you so much for the links + sharing what you tried!!

therealslimhsiehdy (approved):
LGTM, confirmed again with no console errors -- thank you so much for your thoughtful responses back to my questions!!

Image

chaswick (approved):

Issue #4938, opened 2026-04-22
DNS error handling: code-aware alert for the Enroll in DNS hosting admin action


DaisyGuti:
## Type
Admin UX / error surfacing.

## Key files
- `src/registrar/admin.py` — `do_enroll_dns_hosting` action (currently shows a generic message up top).
- `src/registrar/views/domain.py` — if we want matching behavior for any other admin DNS actions.

## Description
The "Enroll in DNS hosting" admin action ([`admin.py` `do_enroll_dns_hosting`](../../src/registrar/admin.py)) catches every failure and shows the same generic Django flash: *"Failed to enroll domain in DNS hosting."* Admins cannot tell whether the failure was auth, zone-not-found, rate-limited, or a transient 5xx without leaving the action and searching `DnsOperationLog` or OpenSearch manually.

A `DnsOperationLog` row is already written on both success and failure (good), and once #4921 lands every failure will raise a typed `DnsHostingError` with `.wire_code`, `.upstream_status`, and `.message`. This ticket uses that typed context to give the admin a specific, actionable flash.

Goals:

1. **Code-aware alert.** Replace the hard-coded generic string with one that includes the wire code, the user-facing message (sourced from the DB-backed message store in #4931 once it lands), and the `request_id` the user would see. Admin and end-user see the same copy.
2. **Direct link to the log row.** Include a link (or the PK) of the just-written `DnsOperationLog` entry in the flash so the admin can click through to the full technical context without re-searching.
3. **Preserve the success alert.** Only the failure path changes.

## Depends on
- #4921 (typed `DnsHostingError` from `CloudflareService`) — required so the action can read `exc.wire_code` / `exc.message` / `exc.upstream_status`.
- #4927 (`DnsOperationLog` + admin) — required so the flash can link to the row.
- #4924 (`request_id` ContextVar) — required so the flash can quote the `request_id`.
- #4931 (admin-editable message store) — nice-to-have; without it, fall back to `_error_mapping`.

## Acceptance criteria
- [ ] `do_enroll_dns_hosting` catches `DnsHostingError` specifically (not bare `Exception`) after #4921 ships; other exceptions fall through to the generic path.
- [ ] On failure, the admin alert includes: wire code (e.g., `DNS_ZONE_NOT_FOUND`), the user-facing message from the store / `_error_mapping` fallback, and the `request_id`.
- [ ] Flash contains a link to the newly-written `DnsOperationLog` row (`admin:registrar_dnsoperationlog_change`).
[...]

Issue #5197, opened 2026-07-14
Align and make consistent the success logs in CloudflareService and DnsHostService


kimallen:
### Issue description

Right now we have "success" logs (logger.info) for most if not all of the api calls _somewhere_ between the CloudflareService and DnsHostService. For some, we log in both places, and for others we log in once place or another. Let's create a strategy for how/where we use success logs and implement it

### Acceptance criteria

- [ ] We have a consistent strategy for logging success messages
- [ ] The code is updated to use that strategy for all api calls to Cloudflare tenant

Issue #4998, opened 2026-05-12
[P3] Define a pattern for DB write failures in DnsHostService


DaisyGuti:
### Issue Description
Some database-write blocks in DnsHostService catch a bare Exception and re-raise it, which hides what actually went wrong. Narrow those catches to the database errors we expect (IntegrityError, DatabaseError) and let anything else propagate, then log the failure with structured fields so it's searchable. A real database error should reach Django's 500 page.

### Acceptance Criteria
- [ ] A database-write failure in DnsHostService surfaces as IntegrityError or DatabaseError, and any other exception propagates unchanged.
- [ ] Database-write failures log structured fields: request_id, model name, and error class.
- [ ] The Exception Contract section of the developer guide describes the rule.
- [ ] Lint and tests pass.

### Additional Context
The affected blocks are the database writes in DnsHostService (account creation, record creation, record update).

When Cloudflare succeeds but the local write fails, that consistency case is handled by the upcoming sync work, not here.

### Links to other issues
- ⛔️ Is blocked by #4924

lizpearl:
@DaisyGuti Just wanted a clarification on overall approach in this ticket. Are you proposing catching, logging and then re-raising `IntegrityError` and `DatabaseError`s?

PR #5372, opened 2026-09-03
#4925 Standardize error responses [KMA]

### LINKED ISSUE #4925: [P2] Standardize DNS hosting error responses

kimallen:
## Ticket

Resolves #4925

## Changes

- Removed unnecessary mid-stream try/catches
- Removed a `JsonResponse` and replaced it with `messages.error` for updating the user's alert banner.
- Refactored `get_x_zone_id_if_zone_exists` which was still returning `nameservers` from early prototype. Created a new function to get nameservers and added/updated tests.
- Return an error message in /admin that includes a request ID and the wire code

## Context for reviewers

These changes are a bit scattered based on the AC. I tried to break down related changes by commit- look at individual commits to see breakdown.

The AC asks for a test to check that request-id is included in the 500 alert messages, but I found that the test in the view was tricky and that we already have a test that checks that the right message is passed in `TestDnsHostingError.test_message_resolves_from_error_mapping_with_request_id`, so I opted out of writing another test.

I simulated a failed enrollment by using a bad tenant id (in sandbox kma), which actually triggers a 400 Bad Request, whose message is not relevant to the /admin page. Messages surfacing in /admin are for a different set of users than in the app, so the message needed tweaking. I got a custom error message from Tara to use. Most importantly, I wanted to ensure the request_id was included. I also added the wire_code to give a little more context without trying to re-write all the error messages meant from FE users. 

Partway through, I realized there is [another ticket](https://github.com/cisagov/manage.get.gov/issues/4938) regarding /admin errors where we can examine/tweak further what's happening there, so it felt out of scope to dig further into how to handle the /admin errors here.

## Setup

dns_hosting waffle flag on, enrolled domain
Also will need a not-yet-enrolled domain

- In local env't: Ensure that the expected validation errors still propagate (use `error-*` options from `MockCloudflareService._mock_cf_error_response`) for record names to trigger different errors
- Walk through `dns_host_service.py` and `DomainDNSRecordsView` in `views/domain.py` and ensure we aren't raising any APIErrors or using any try/except blocks that interrupt `DnsHostingError`s already raised.
- In kma sandbox: To view the error in /admin, you can use a 'bad' tenant_key, which triggers a 400 error. (shows up as DNS_VALIDATION_FAILED)

## Screenshots

image


src/registrar/services/dns_host_service.py
```diff
@@ -142,16 +142,7 @@ def dns_zone_setup(self, domain_name, x_account_id):
         if zone_data:
             self.create_db_zone({"result": zone_data}, domain_name)
         else:
-            try:
```
kimallen, line 145: Logging and exceptions already happen in upstream calls

src/registrar/services/dns_host_service.py
```diff
@@ -228,17 +219,8 @@ def create_and_save_zone(self, domain_name, x_account_id):
         zone_data = self.dns_vendor_service.get_zone_by_id(x_zone_id)
 
         # Create and save zone in registrar db
-        try:
```
kimallen, line 231: Exceptions and error logging already happen in upstream calls.

src/registrar/views/domain.py
```diff
@@ -1018,14 +1016,10 @@ def _handle_edit(self, request, x_zone_id: str, form_record_data: dict, record_i
         """Update an existing DNS record and prepare the DB-backed row for rendering."""
         try:
             dns_record = self.dns_host_service.update_dns_record(x_zone_id, record_id, form_record_data)
-
         except ValueError as e:
             messages.error(request, str(e))
             raise GenericError(GenericErrorCodes.GENERIC_ERROR)
```
kimallen, line 1021: This gets caught in `post` as a GenericError. I'm not sure that's how we want to handle db or DnsHostService level errors, but there is a [ticket](https://github.com/cisagov/manage.get.gov/issues/4938) for developing the desired pattern.

src/registrar/services/dns_host_service.py
```diff
@@ -354,8 +336,7 @@ def _find_existing_zone_in_cf(self, zone_name, x_account_id) -> dict | None:
 
         return zone_data
 
-    def get_x_zone_id_if_zone_exists(self, domain_name) -> tuple[str | None, list[str] | None]:
-        # returns x_zone_id (and temporarily returns nameservers)
+    def get_x_zone_id_if_zone_exists(self, domain_name) -> str | None:
```
kimallen, line 339: We were returning `nameservers` here as a relic of the original prototype when we we needed nameservers in the view. This removes them from this function. Below I created a new function specific to retrieving nameservers and apply it where we need to use them.

src/registrar/admin.py
```diff
@@ -5726,24 +5726,37 @@ def get_form(self, request, obj=None, **kwargs):
...
-                failed_enrollment_message,
+                get_failed_enrollment_message(e.context.get("request_id"), e.wire_code),
                 messages.ERROR,
             )
         except Exception as e:
             logger.exception(e)
             self.message_user(
                 request,
-                failed_enrollment_message,
+                get_failed_enrollment_message(e.context.get("request_id"), e.wire_code),
```
allly-b, line 5759: **Change request:** Only the DnsHostingError have context added, but this is catching any generic exception. If this is triggered it will raise an additional error resulting in a 500.

src/registrar/views/domain.py
```diff
@@ -1115,12 +1108,10 @@ def post(self, request, *args, **kwargs):  # noqa: C901
...
-            x_zone_id, nameservers = self.dns_host_service.get_x_zone_id_if_zone_exists(self.object.name)
-            if not x_zone_id:
-                return JsonResponse(
-                    {"status": "error", "message": "DNS zone not found. Domain may not be enrolled."},
-                    status=400,
-                )
+            x_zone_id = self.dns_host_service.get_x_zone_id_if_zone_exists(self.object.name)
+            if x_zone_id is None:
+                messages.error(request, DnsHostingError.GENERIC_ERROR_MESSAGE)
+                return
```
allly-b, line 1114: **change requested:** I believe this line will raise an exception because you're returning None on a view's post which is not allowed. Still figuring out how to replicate though so don't have a screen shot to share as proof.

allly-b (commented):
Pausing my review, and will finish the review tomorrow. Figured I would at least share what I have so far though it's minimal at the moment.

src/registrar/views/domain.py
kimallen: Updated to return `_error_response`. It's currently set as a 400, though I'm not sure that's the correct one to use. Looking into it more.

src/registrar/admin.py
kimallen: Good catch- updated!

src/registrar/tests/services/test_dns_host_service.py
```diff
@@ -400,17 +400,18 @@ def test_enroll_domain_success(self):
         domain_name = "test.gov"
         domain = create_domain(**{"domain_name": domain_name})
 
-        mock_get_x_zone_id_if_zone_exists = Mock(return_value=(None, ["ns1.example.gov", "ns2.example.gov"]))
+        mock_get_x_zone_id_if_zone_exists = Mock(return_value=None)
         self.service.get_x_zone_id_if_zone_exists = mock_get_x_zone_id_if_zone_exists
         self.service.dns_account_setup = Mock(return_value="12345")
         self.service.dns_zone_setup = Mock()
+        self.service.get_nameservers_from_zone = Mock(return_value=["ns1.rainbow.gov, ns2.rainbow.gov"])
```
allly-b: **comment/non blocking:** why was this changed to a list with one string element? instead of a list with 2 nameserver strings

docs/developer/dns-error-handling.md
```diff
@@ -138,23 +138,12 @@ except DnsHostingError as exc:
     # The view catches on purpose to handle the user-facing business logic:
     # which message to show, which HTTP status, which template to render, etc.
     # No logger call here as the service already logged.
-    return dns_error_response(exc)
+    messages.error(request, exc.message). # displays the desired error message to the client
```
allly-b: **nitpick**
```suggestion
    messages.error(request, exc.message)  # displays the desired error message to the client
```

src/registrar/services/dns_host_service.py
allly-b: So if this is no longer returning the nameservers, I think the related tests need to be updated too. I was browsing for when this function is used and found
`test_added_record_edit_form_has_cancel_wiring` , `test_invalid_edit_post_triggers_messages_refresh, `test_post_valid_forms_create_dns_records_success` all still return nameservers. If you search in tests/views/test_domain_dns_record_and_form.py  for `svc.get_x_zone_id_if_zone_exists.return_value = ("zone-123", ["ex1.dns.gov"])` there are 14 cases where the zone and nameserver is still be returned. I'm okay if you want to punt updated those tests to a different ticket. Can you reply to this with what ticket those tests will be completed in (you can even make a 1 ptr ticket that does the cleanup there), or update the tests in this PR? Definitely, correct my understanding if I'm off base here and they don't need updating.

src/registrar/admin.py
```diff
@@ -5726,24 +5726,39 @@ def get_form(self, request, obj=None, **kwargs):
...
+                get_failed_enrollment_message(request_id, e.wire_code),
                 messages.ERROR,
             )
         except Exception as e:
             logger.exception(e)
+            request_id = None
             self.message_user(
                 request,
-                failed_enrollment_message,
+                get_failed_enrollment_message(request_id, e.wire_code),
```
allly-b: **fix before merge:** I just realized, would the exception, e, actually have wire_code as a value here? Since it's not a dnshosting error it won't have a wire code, right? I _think_ this function may just need to be called with (none, none). Correct me if I'm wrong though, or just fix before merging. Not blocking on it.

allly-b (approved):
Tested on local and in kma sandbox: 
Image

Image

Image

Suggestion:
I was tracing the code and saw that when _type_dns_error get’s called if it isn’t an HTTP Response (see [this line](https://github.com/cisagov/manage.get.gov/blob/a2f4d2aa7284b9b880da1f75aaabae27135202e9/src/registrar/services/cloudflare_service.py#L67-L68)) then there is no request id added. I think you should still be able to get the context and add the request id there. However not blocking on that.



A couple comments within the code, only one that probably should be fixed before merging. Thank you for updating the documentation and explaining why you did certain things. Much appreciated.

src/registrar/tests/services/test_dns_host_service.py
kimallen: oops, not intentional. Probably an initial blunder and then it got copy/pasted. I reverted it to a list of two nameserver strings. (the reason I changed the naming is b/c I'm trying to move away from using example.gov since it's actually now a real domain), though it's more important in the fixtures to move away from it.

src/registrar/services/dns_host_service.py
kimallen: Good catch- I'm surprised it doesn't error, but I guess the mock return value isn't checked against the signature.

I updated the tests.

kimallen:
> Suggestion: I was tracing the code and saw that when _type_dns_error get’s called if it isn’t an HTTP Response (see [this line](https://github.com/cisagov/manage.get.gov/blob/a2f4d2aa7284b9b880da1f75aaabae27135202e9/src/registrar/services/cloudflare_service.py#L67-L68)) then there is no request id added. I think you should still be able to get the context and add the request id there. However not blocking on that.

@allly-b I went ahead and updated this.

src/registrar/services/dns_host_service.py
asaki222: `create_db_account` has an exception with a similar log.

asaki222, line 167: `create_db_account` has an exception with a similar log.  This might be out of scope of this ticket. 🤔

asaki222 (approved):
The code looks good to me! I confirmed that the error message shows correctly on admin view with the id.

Image

src/registrar/services/dns_host_service.py
kimallen: Good eye! I agree that it's out of scope. [This one](https://github.com/cisagov/manage.get.gov/issues/5197) handles success messages and then the Exceptions (originating in DnsHostService) can be handled in [this ticket](https://github.com/cisagov/manage.get.gov/issues/4998).

Issue #5367, opened 2026-09-01
Integrity improvements in DNS backend flow: minor changes


kimallen:
### Issue description

Based on some evaluation of overall DNS flow, a few minor improvements are clustered here:

**DnsRecord.ttl**: The field defaults to 1 (cf's auto value), but the validation_ttl constrains to 60-86400. And then the helper function in `enums.py` `format_dns_ttl` returns only one of eight values, and otherwise will throw a key error. Let's make it more robust by:
  1. Adding an item (1, "5 minutes") to `DNS_TTL_CHOICES`
  2. Either update the return value of `format_dns_ttl` to have a default so that it doesn't throw a KeyError `dict(DNS_TTL_CHOICES).get(ttl, f"{ttl} seconds")`
OR
Update the model to only allow the values in DNS_TTL_CHOICES

Though those are the values we are choosing, when deciding the strategy, consider that importing records could include other values.

 **DnsZone.clean():** raise the ValidationError instead of returning the ValidationError (enforces "at least 2 nameservers"

 **in dns_record.py, `update_From_vendor_data`**: it lists excluded fields. Instead list included fields ("The Cloudflare response is a trust boundary; an unexpected or attacker-influenced field name lands directly on the model.") 

**Update comment field max limit in DnsRecord model**: The max char limit in CF is 100 for the comment field. However, in our db it's 500. Update the db to 100 max

### Acceptance criteria

- [ ] DnsRecord ttl is refactored and can handle an entry of 1 and can't break if a value not in DNS_TTL_CHOICES is submitted
- [ ] DnsZone.clean raises rather than returns an error
- [ ] DnsRecord.update_from_vendor_data lists included fields rather than excluded fields
- [ ] Max length for DnsRecord.comment is 100 chars
- [ ] The Max length and character count feature works as expected on the comment field in a DNS form (includes migration/db mitigation)

### Additional context

These could be split into different PRs for easier review if desired

PR #5388, opened 2026-09-10
[EXTERNAL CONTRIBUTOR] #5367: Fix KeyError crash in format_dns_ttl for TTL values outside DNS_TTL_CHOICES

soroush5:
Fixes the TTL bullet of #5367.

Problem: DnsRecord.ttl_display calls format_dns_ttl, which did dict(DNS_TTL_CHOICES)[ttl] and raised KeyError for anything outside the 8 listed values. That includes ttl=1 (the model default and Cloudflare's automatic TTL from vendor sync) and any valid 60-86400 value like 120 that passes model validation but is not in the form choices. Opening the DNS records page for such a record returns a 500.

Approach: fall back to a plain seconds label for unknown values instead of raising. I did not add 1 to DNS_TTL_CHOICES since model validation still rejects ttl=1, which would make the form offer a value that cannot validate.

Tests: added SimpleTestCase coverage in registrar/tests/test_enums.py (known values pass through, 1/120/600 render a fallback). Verified the new test fails with KeyError on the pre-fix code and passes after. Neighbor suites test_cleaners and test_validations still pass (21 tests, no DB).

Out of scope for this PR: the other #5367 bullets (DnsZone.clean, update_from_vendor_data allowlist, comment max length).


katypies:
Hi, @soroush5  We're closing this because the PR hasn't fully addressed the requirements of the issue, and this issue is part of a broader effort; we already have team members who have been working on this area of code and the upcoming changes. We appreciate your interest and contributions. Going forward, please comment on open issues before starting implementations, to help ensure that it's not duplicative work, or already planned for internal implementation.

PR #5389, opened 2026-09-10
[EXTERNAL CONTRIBUTOR] Isolate Cloudflare mocks to DNS clients

### LINKED ISSUE #5364: Improve MockCloudflareService setup for development
### Issue description

Currently we initialize `MockCloudflareService` in `app.py` so that it's running whenever we want to hit the cf api. However, there are a few problems with that:

- ready() runs in every process context unnecessarily (Migrations, collectstatic, manage.py load, cron/scheduled jobs, the autoreloader's parent process, the test runner)
- respx patches httpx globally, not per-client (this means other httpx calls (not just CF) is routed through the router, so future httpx calls that are not CF will blow up.

Let's do some cleanup and refactoring to mitigate some of these problems. Here's some suggestions:

1. Remove the MockCloudflare initializating from app.py and add it in `build_dns_client` (this will have it run for just the client, not every context)

in  `registrar/services/dns_http_client.py`
```
def build_dns_client():
    transport = RetryTransport(retries=CONNECT_RETRIES)
    client = httpx.Client(timeout=DNS_TIMEOUT, transport=transport)
    if settings.DNS_MOCK_EXTERNAL_APIS and not settings.IS_PRODUCTION:
        from registrar.services.mock_cloudflare_service import MockCloudflareService
        mock_service = MockCloudflareService()
        if not mock_service.is_active:
            mock_service.start()
    return client
```

2. Calls that don't come from CF get auto-mocked with a 200 response. Use a `pass_through` so that those calls are not mocked.

Anything that isn't Cloudflare goes to the real network. Without this,
 every httpx call in the process is auto-mocked with an empty 200.
```
self._mock_context.route(~M(host="api.cloudflare.com")).pass_through()
```

3. Now that we are passing through any of the non-CF calls, we want to flip `assert_all_mocked` to True - any CF paths that don't have mocking set up will fail, alerting us that we need a mock for it.

Do evaluate if there are other improvements to make regarding the setup of MockCloudflareService

4. Ensure the mocks can't get set in PROD

### Acceptance criteria

- [ ] We are no longer initializing MockCloudflareService for every process (it is no longer in apps.py) 
- [ ] The mocks still work as expected in development env't, and do not run in prod

### Additional context

[Slack thread](https://cisa-corp.slack.com/archives/C074E2QJ7J9/p1788299427017339)

### Links to other issues

_No response_

sylvesterkaczmarek:
## Summary

Move `MockCloudflareService` startup out of Django app initialization so it is only activated when a DNS client is actually built in non-production environments.

- start the mock lazily from `build_dns_client`
- pass non-Cloudflare HTTPX traffic through untouched
- fail on unregistered Cloudflare routes instead of returning an empty mocked response
- avoid resetting an already-active singleton when additional DNS clients are created
- update lifecycle and routing tests for the new behavior

## Testing

- targeted DNS HTTP client and MockCloudflareService tests: 25/25 passing
- `./manage.py lint`: all linting, formatting, type checking, and security scanning passed

Fixes #5364


katypies:
Hi, @sylvesterkaczmarek  We're closing this because the DNS hosting work is part of a broader effort, and we already have team members who have been working on this area of code and the upcoming changes. We appreciate your interest and contributions. Going forward, please comment on open issues before starting implementations, to help ensure that it's not duplicative work, or already planned for internal implementation.

sylvesterkaczmarek:
Thanks for clarifying. Understood — I’ll coordinate on the issue before picking up planned registrar work going forward.

